PR #12, opened 2020-09-26
Fix a typo in README.md

YingboMa:
BTW, why is
```Julia
@test autodiff(f1, Active(1.0)) ≈ 1.0
```
? Shouldn't it be `2`?


wsmoses:
Yes absolutely. Clearly we were too fast in setting up the tests and didn't check the readme.

PR #22, opened 2020-11-03
Bump Enzyme convention

wsmoses:
This needs to correspond with using a new build of Enzyme proper.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -12,7 +12,7 @@ Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 
 [compat]
 Cassette = "0.3"
-Enzyme_jll = "0.0.2"
+Enzyme_jll = "0.0.3"
 GPUCompiler = "0.7"
 LLVM = "3.2"
 julia = "1.5"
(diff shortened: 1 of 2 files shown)


vchuravy:
I think we need to update the Yggdrasil builds as well?

wsmoses:
On it, https://github.com/JuliaPackaging/Yggdrasil/pull/2042

vchuravy:
Maybe the binary is already following those conventions? Otherwise CI shouldn't pass.

wsmoses:
That's what I'm guessing given https://github.com/wsmoses/Enzyme.jl/issues/21 but I don't see how that happened

vchuravy:
Yeah I messed up and pushed an Enzyme_jll without incrementing the patch version https://travis-ci.com/github/wsmoses/Enzyme.jl/jobs/427340400

vchuravy:
needs https://github.com/JuliaRegistries/General/pull/24115

Project.toml
```diff
@@ -12,7 +12,7 @@ Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 
 [compat]
 Cassette = "0.3"
-Enzyme_jll = "0.0.1"
+Enzyme_jll = "0.0.2"
```
vchuravy, line 15:
```suggestion
Enzyme_jll = "0.0.3"
```

PR #35, opened 2021-01-14
Update README.md

ViralBShah:


vchuravy:
Thanks Viral,

know I just need to know how CI broke :)

ViralBShah:
You may already know - but the example crashed for me on a recent 1.6 build, but worked fine on 1.5.3 on mac.

vchuravy:
Yeah 1.6 is not supported yet, we are not building Enzyme itself for LLVM 11.

wsmoses:
I'm amused that of all the platforms and headaches it gave us, MacOS decides to be fine right now.

@vchuravy do we want to add an LLVM 11 build on Yggdrassil (or does that require some more julia's version of LLVM setup that isn't worth doing at the moment). We do test LLVM 11 in https://github.com/wsmoses/Enzyme proper

ViralBShah:
Maybe note the supported Julia versions in the README? Perhaps the Julia dependency in Project.toml should be stricter to restrict to 1.5 only?

vchuravy:
I was hoping that we would get https://github.com/JuliaPackaging/JLLWrappers.jl/pull/27 before we have to compile for multiple Julia versions.

PR #38, opened 2021-01-23
Dub tanh

### LINKED ISSUE #37: Enzyme autodiff breaking on tanh function
The following example crashes Julia on my system

```julia
using Enzyme
autodiff(tanh, Active(1.0))
```

version info of my system:
Julia Version 1.5.2
Commit 539f3ce943 (2020-09-23 23:17 UTC)
Platform Info:
OS: macOS (x86_64-apple-darwin18.7.0)
CPU: Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
WORD_SIZE: 64
LIBM: libopenlibm
LLVM: libLLVM-9.0.1 (ORCJIT, skylake)

wsmoses:
Resolves #37

### DIFF
--- a/src/Enzyme.jl
+++ b/src/Enzyme.jl
@@ -71,11 +71,11 @@ for op in (copysign,)
     end
 end
 
-for op in (asin,)
+for op in (asin,tanh)
     for (T, llvm_t) in ((Float32, "float"), (Float64, "double"))
         decl = "declare double @$(nameof(op))($llvm_t)"
         func = """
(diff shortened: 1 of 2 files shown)


vchuravy (approved):
Needs a test.

Everytime we add another overdub I die a little inside. The sad thing is that we can't fix this with custom adjoint's (the inlining of the function problem), so we need to figure out how we can make the frontend cooperate with the enzyme backend, while maintaining the capability of Julia to inline functions, and preferably doing so without reimplementing activity analysis...

PR #42, opened 2021-02-02
Bump versions

vchuravy:


src/compiler.jl
```diff
@@ -86,11 +86,23 @@ Create the `FunctionSpec` pair, and lookup the primal return type.
 
     # can't return array since that's complicated.
     rt = Core.Compiler.return_type(Cassette.overdub, overdub_tt)
-    @assert rt<:Union{AbstractFloat, Nothing}
+    if !(rt<:Union{AbstractFloat, Nothing})
+        @error "Return type should be <:Union{Nothing, AbstractFloat}" rt adjoint primal
```
wsmoses, line 90: Is there a nice way to test this per chance via Julia exception handling?

wsmoses (approved):

src/compiler.jl
vchuravy: The right thing to do is to define a custom exception. And then test for that.

PR #60, opened 2021-05-14
Initial garbage collection handling

wsmoses:


vchuravy (approved):

src/compiler/cassette.jl
```diff
@@ -59,6 +59,46 @@ function ir_element(x, code::Vector)
     return x
 end
 
+@inline function Base.copy(a::Array{T,1}) where {T}
```
vchuravy, line 62: This is type-piracy... We should do this in LLVM IR and link it in as a module.

PR #67, opened 2021-06-02
Eliminate need for cassette

wsmoses:
Requires https://github.com/JuliaGPU/GPUCompiler.jl/pull/196 and https://github.com/wsmoses/Enzyme/pull/209


vchuravy (commented):
Can we land the diverse bugfixes?

PR #68, opened 2021-06-23
Remove erf special case

wsmoses:


src/typetree.jl
```diff
@@ -110,6 +110,9 @@ function typetree(@nospecialize(T), ctx, dl)
     end
 
     if fieldcount(T) == 0
+        if T <: Function
```
vchuravy, line 113: If the Field count is 0 we might want to just return `TypeTree()` since it is a ghost or singleton?

test/abi.jl
```diff
@@ -82,6 +82,9 @@ using Test
     @test dx[] ≈ 3.0
     @test dy[] ≈ 2.0
 
+    mid, = Enzyme.autodiff((f, x) -> f[1](x), [x->x*x], Active(2.0))
```
vchuravy, line 85:
```suggestion
    mid, = Enzyme.autodiff((fs, x) -> fs[1](x), (x->x*x,), Active(2.0))
```

Since the moment you have more than one entry in `fs` the array version will explode

PR #79, opened 2021-07-04
Initial handling of jl_apply_generic

wsmoses:
Requires https://github.com/wsmoses/Enzyme/pull/222

### DIFF
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -36,7 +36,7 @@ function array_shadow_handler(B::LLVM.API.LLVMBuilderRef, OrigCI::LLVM.API.LLVMV
 
     b = LLVM.Builder(B)
 
-    vals = Vector{LLVM.Value}()
+    vals = LLVM.Value[]
     for i = 1:numArgs
         push!(vals, LLVM.Value(unsafe_load(Args, i)))
     end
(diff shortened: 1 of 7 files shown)


src/compiler.jl
```diff
@@ -87,6 +87,343 @@ function null_free_handler(B::LLVM.API.LLVMBuilderRef, ToFree::LLVM.API.LLVMValu
...
+
+@generated function alloc(tt::Type{T}) where T
+    quote
+        ref::Any = Base.llvmcall(("""
+        declare {}*** @julia.ptls_states()
+        declare noalias nonnull {} addrspace(10)* @julia.gc_alloc_obj(i8*, i64, {} addrspace(10)*)
+        define {} addrspace(10)* @allocate(i64 %size, {} addrspace(10)* %ty) #1 {
+            %1 = call {}*** @julia.ptls_states()
+            %2 = bitcast {}*** %1 to i8*
+            %res = call noalias nonnull {} addrspace(10)* @julia.gc_alloc_obj(i8* %2, i64 %size, {} addrspace(10)* %ty)
```
vchuravy, line 101: You are not setting the typetag of the allocation. So I would expect `typeof(alloc(Int))` to fail.

vchuravy:
```
  %9 = call noalias nonnull {}* @jl_gc_pool_alloc(i8* %ptls_i8, i32 1400, i32 16) #1
  %10 = bitcast {}* %9 to i64*
  %11 = getelementptr inbounds i64, i64* %10, i64 -1
  store atomic i64 140278619936176, i64* %11 unordered, align 8

```

where `140278619936176` is `Base.pointer_from_objref(T)`

vchuravy: Oh you are calling gc_alloc_obj, right...

vchuravy:
```
@generated function alloc(tt::Type{T}) where T
           sz = sizeof(T)
           type = reinterpret(Int64, Base.pointer_from_objref(T))
           mod ="""
               declare {}*** @julia.ptls_states()
               declare noalias nonnull {} addrspace(10)* @julia.gc_alloc_obj(i8*, i64, {}*)
               define {} addrspace(10)* @allocate() #1 {
                   %1 = call {}*** @julia.ptls_states()
                   %2 = bitcast {}*** %1 to i8*
                   %res = call noalias nonnull {} addrspace(10)* @julia.gc_alloc_obj(i8* %2, i64 $sz, {}* inttoptr (i64 $type to {}*))
                   ret {} addrspace(10)* %res
               }
               attributes #1 = { alwaysinline }
               """
           quote
               ref = Base.llvmcall(($mod, "allocate"), Any, Tuple{})
               return ref
           end
       end
```

src/compiler.jl
```diff
@@ -609,25 +608,50 @@ function apply_latest_rev(B::LLVM.API.LLVMBuilderRef, OrigCI::LLVM.API.LLVMValue
...
+        jlerror_func = LLVM.Function(mod, "jl_error", jlerror_T)
+        push!(function_attributes(jlerror_func), EnumAttribute("noreturn"; ctx))
+    end
+
+    # 2. Call error function and insert unreachable
+    call!(B, jlerror_func, LLVM.Value[globalstring_ptr!(B, string)])
+    unreachable!(B)
+
+    # 3. Change insertion point so that we don't stumble later
+    # FIXME(@wsmoses): It seems like Enzyme does not use the `B` passed here
```
vchuravy, line 631: @wsmoses One can't create basic blocks since the builder afterwards doesn't know where to continue.

wsmoses:
There's almost certainly a simpler reproducer, but this was the example causing internal GC issues:

```
using Enzyme

using OrdinaryDiffEq
function lorenz(du,u,p,t)
 du[1] = 10.0(u[2]-u[1])
 du[2] = u[1]*(28.0-u[3]) - u[2]
 du[3] = u[1]*u[2] - (8/3)*u[3]
end

function f(u0)
  tspan = (0.0,1.0)
  prob = ODEProblem(lorenz,u0,tspan)
  sol = solve(prob,Tsit5())[end]
  sol[1]::Float64
end

u0 = [1.0;0.0;0.0]
du0 = [0.0;0.0;0.0]
@show f(u0)
Enzyme.API.printall!(true)
Enzyme.API.inlineall!(true)
@show Enzyme.autodiff(f, Duplicated(u0,du0))
```

PR #90, opened 2021-07-18
WIP: Add demo

vchuravy:


notebooks/demo.jl
```diff
@@ -0,0 +1,636 @@
...
+	Duplicated(R, ∂z_∂R),
+	Duplicated(A, ∂z_∂A),
+	Duplicated(B, ∂z_∂B))
+
+# ╔═╡ 62182140-d067-4814-a83d-f6c9d6b3a15f
+R ≈ A * B
+
+# ╔═╡ f6856069-8bfa-49ed-8f5e-f1bd6009c4f4
+∂z_∂A ≈ ∂z_∂R * B' # equivalent to Zygote.pullback(*, A, B)[2](∂z_∂R)[1]
+
```
wsmoses, line 162: This explanation is a bit odd because it makes it seem like this manual math is equivalent to zygote, rathre than explaining the result of Enzyme sets the value the same as the pullback.

PR #94, opened 2021-07-20
Improve activity deduction

wsmoses:


src/Enzyme.jl
```diff
@@ -73,6 +73,8 @@ Base.eltype(::Type{<:Annotation{T}}) where T = T
 function guess_activity(T)
     if T <: AbstractFloat || T <: Complex{<:AbstractFloat}
         return Active{T}
+    elseif T <: AbstractArray
+        return Duplicated{T}
```
vchuravy, line 77: Can we do that for return?

vchuravy (commented):
Needs tests

PR #104, opened 2021-07-23
Mark aliasing property of error

### LINKED ISSUE #102: Julia crash with Enzyme on product of sums
Enzyme (current master) crashes Julia when auto-diffing a product of sums.

A single sum works fine:

```julia
using Enzyme

x = rand(10)
dx = zero(x)

foo(x::AbstractVector{<:Real}) = sum(x)
foo(x)
autodiff(foo, Duplicated(x, dx))
```

But

```julia
bar(x::AbstractVector{<:Real}) = sum(x) * sum(x)
bar(x)
autodiff(bar, Duplicated(x, dx))
```

goes boom:

```
julia: /buildworker/worker/package_linux64/build/src/llvm-late-gc-lowering.cpp:857: std::vector<int, std::allocator<int> > LateLowerGCFrame::NumberAllBase(State&, llvm::Value*): Assertion `Tracked.size() == Numbers.size()' failed.

signal (6): Aborted
in expression starting at REPL[9]:1
gsignal at /lib64/libc.so.6 (unknown line)
abort at /lib64/libc.so.6 (unknown line)
__assert_fail_base at /lib64/libc.so.6 (unknown line)
__assert_fail at /lib64/libc.so.6 (unknown line)
...
Aborted (core dumped)
```

wsmoses:
Resolves #102

### DIFF
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -4044,7 +4044,7 @@ function GPUCompiler.codegen(output::Symbol, job::CompilerJob{<:EnzymeTarget};
                                   ])
             continue
         end
-        # Since this is noreturn, it can't write to any operations in the function
+        # Since this is noreturn and it can't write to any operations in the function
         # in a way accessible by the function. Ideally the attributor should actually
         # handle this and similar not impacting the read/write behavior of the calling
         # fn, but it doesn't presently so for now we will ensure this by hand
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -4044,6 +4044,15 @@ function GPUCompiler.codegen(output::Symbol, job::CompilerJob{<:EnzymeTarget};
                                   ])
             continue
         end
+        # Since this is noreturn, it can't write to any operations in the function
+        # in a way accessible by the function. Ideally the attributor should actually
+        # handle this and similar not impacting the read/write behavior of the calling
+        # fn, but it doesn't presently so for now we will ensure this by hand
+        if func == Base.Checked.throw_overflowerr_binaryop
+            llvmfn = functions(mod)[k.specfunc]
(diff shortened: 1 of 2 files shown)


test/runtests.jl
```diff
@@ -126,6 +126,17 @@ end
     @test dinp ≈ Float64[1.0, 1.0]
 end
 
+@testset "Advanced array tests sq" begin
+    function arsumsq(f::Array{T}) where T
+        return sum(f) * sum(f)
+    end
+    inp = Float64[1.0, 2.0]
+    dinp = Float64[0.0, 0.0]
+    autodiff(arsum2, Active, Duplicated(inp, dinp))
```
vchuravy, line 135:
```suggestion
    autodiff(arsumsq, Active, Duplicated(inp, dinp))
```

vchuravy:
Rebase?

src/compiler.jl
```diff
@@ -4044,6 +4044,15 @@ function GPUCompiler.codegen(output::Symbol, job::CompilerJob{<:EnzymeTarget};
                                   ])
             continue
         end
+        # Since this is noreturn, it can't write to any operations in the function
+        # in a way accessible by the function. Ideally the attributor should actually
```
vchuravy, line 4048: This is not generally true for no-return functions in the presence of longjump/setjmp. It is true for this function

src/compiler.jl
```diff
@@ -4044,6 +4044,15 @@ function GPUCompiler.codegen(output::Symbol, job::CompilerJob{<:EnzymeTarget};
                                   ])
             continue
         end
+        # Since this is noreturn, it can't write to any operations in the function
```
wsmoses, line 4047:
```suggestion
        # Since this is noreturn and it can't write to any operations in the function
```

PR #111, opened 2021-07-30
Fix the readme's example

mcabbott:
As written it tries to compare a tuple to a number:
```
julia> @test autodiff(f1, Active(1.0)) ≈ 2.0
Error During Test at REPL[236]:1
  Test threw exception
  Expression: autodiff(f1, Active(1.0)) ≈ 2.0
  MethodError: no method matching isapprox(::Tuple{Float64}, ::Float64)
```


wsmoses:
Thanks! Yeah we changed the API for that a while back and clearly didn't update the readme.

PR #115, opened 2021-07-31
Refactor ABI calling convention

wsmoses:


vchuravy:
Now if you can explain me why this breaks the windows ABI xD

vchuravy:
My debuging results so far:

```
; Function Attrs: alwaysinline
define fastcc void @diffejulia_f_2337wrap({ [2 x double] }* %0, [2 x double] %1, [2 x double] %2) #2 {
entry:
  %.repack = getelementptr inbounds { [2 x double] }, { [2 x double] }* %0, i64 0, i32 0, i64 0
  %.elt = extractvalue [2 x double] %2, 0
  store double %.elt, double* %.repack, align 8
  %.repack1 = getelementptr inbounds { [2 x double] }, { [2 x double] }* %0, i64 0, i32 0, i64 1
  %.elt2 = extractvalue [2 x double] %2, 1
  store double %.elt2, double* %.repack1, align 8
  ret void
}
```

Whereas on the caller side:

```
     %7 = ptrtoint <2 x i64>* %3 to i64
     %8 = inttoptr i64 %.fca.0.extract to void (i64, [2 x double]*, [2 x double]*)*
     call void %8(i64 %7, [2 x double]* byval %4, [2 x double]* byval @_j_const1)
```

So Julia thinks that for the ccall it needs to pass a pointer instead of an struct. Now why that is only a problem on windows...

PR #122, opened 2021-08-12
Support GC.at_preserve

vchuravy:


Project.toml
```diff
@@ -10,6 +10,7 @@ Enzyme_jll = "7cc45869-7501-5eee-bdea-0790c847d4ef"
 GPUCompiler = "61eb1bfa-7361-4325-ad38-22787b887f55"
 LLVM = "929cbde3-209d-540e-8aea-75f648917ca0"
 Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
+MPI = "da04e1cc-30fd-572f-bb4f-1f8673147195"
```
vchuravy, line 13:
```suggestion
```

wsmoses:
@vchuravy rebased, this now passes

src/compiler.jl
```diff
@@ -1720,6 +1939,8 @@ function GPUCompiler.codegen(output::Symbol, job::CompilerJob{<:EnzymeTarget};
         augmented_primalf = nothing
     end
 
+    # fix_inlining!(mod)
```
vchuravy, line 1942: So we can remove `fix_inling` from this PR?

src/compiler/validation.jl
```diff
@@ -320,6 +320,33 @@ function check_ir!(job, errors, imported, inst::LLVM.CallInst, known_fns)
...
+                if ptr == cglobal(:jl_array_del_beg)
+                    fn = "jl_array_del_beg"
+                end
+                if ptr == cglobal(:jl_array_del_end)
+                    fn = "jl_array_del_end"
+                end
+                if ptr == cglobal(:jl_array_del_at)
+                    fn = "jl_array_del_at"
+                end
+		if ptr == cglobal(:jl_value_ptr)
```
vchuravy, line 341: Formatting

src/compiler.jl
```diff
@@ -975,28 +1101,121 @@ function annotate!(mod)
...
+
     for rfn in ("jl_object_id_", "jl_object_id")
         if haskey(fns, rfn)
             fn = fns[rfn]
             push!(function_attributes(fn), LLVM.EnumAttribute("readonly", 0; ctx))
         end
     end
+
+    # MPI_Irecv
+    for rfn in ("MPI_Irecv", "PMPI_Irecv")
```
vchuravy, line 1142: Can we split those out?

vchuravy (commented):
I am going to split the MPI changes into their own PR

src/compiler.jl
wsmoses: it was to attempt to fix mpi misbehaving, but I don't think it ended up being that issue. Regardless fine to split with MPI

PR #131, opened 2021-09-15
add fix_inlining

vchuravy:


vchuravy:
@wsmoses no idea what this was needed for. So I rescued the code to a branch xD

wsmoses:
Yeah basically when enzyme-inline is run Enzyme itself inlines a bunch of stuff for the reverse pass it generates. This results in many different calls to ptls states, sometimes in loops, etc. In case it was an issue this canonicalizes them all together.

PR #139, opened 2021-09-30
Handle undefined symbols at runtime

wsmoses:


src/compiler.jl
```diff
@@ -908,6 +915,22 @@ function setfield_rev(B::LLVM.API.LLVMBuilderRef, OrigCI::LLVM.API.LLVMValueRef,
     return nothing
 end
 
+function get_binding_or_error_fwd(B::LLVM.API.LLVMBuilderRef, OrigCI::LLVM.API.LLVMValueRef, gutils::API.EnzymeGradientUtilsRef, normalR::Ptr{LLVM.API.LLVMValueRef}, shadowR::Ptr{LLVM.API.LLVMValueRef}, tapeR::Ptr{LLVM.API.LLVMValueRef})::Cvoid
+    CI = API.EnzymeGradientUtilsNewFromOriginal(gutils, OrigCI)
+    err = emit_error(LLVM.Builder(B), "Enzyme: unhandled forward for jl_get_binding_or_error")
```
vchuravy, line 920: this seems wrong? We just want to call the original function.

vchuravy (changes requested):

src/compiler.jl
wsmoses: We already call the original function [and don't remove that call], we just err after the fact, if for some reason resolution comes back ok

vchuravy, line 920:
```suggestion
    err = emit_error(LLVM.Builder(B), "Enzyme: unreachable reached")
```

wsmoses: So I wouldn't say this is necessarily an unreachable, so much as an unimplemented global, no?

vchuravy (approved):

PR #142, opened 2021-10-01
Parallel task + closure support

wsmoses:
Requires JLL bump

### DIFF
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -20,19 +20,22 @@ end
 # User facing interface
 abstract type AbstractThunk{F, RT, TT} end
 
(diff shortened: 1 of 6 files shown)


src/compiler.jl
```diff
@@ -204,6 +207,38 @@ end
 dedupargs() = ()
 dedupargs(a, da, args...) = (a, dedupargs(args...)...)
 
+
+function runtime_newtask_fwd(ret_ptr::Ptr{Any}, fn::Any, dfn::Any, post::Any, ssize::Int)
+    @warn "active variables passeed by value to jl_new_task are not yet supported"
```
vchuravy, line 212: Instead of warning unconditionally can we instead check if they are active?

PR #152, opened 2021-11-01
[WIP] CUDA.jl internal support

wsmoses:
Now that the other pieces have landed, rebased this if you can take a look at the apparent internal errors @vchuravy

### DIFF
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -45,7 +45,7 @@ function get_function!(builderF, mod, name)
     if haskey(functions(mod), name)
         return functions(mod)[name]
     else
-        return builderF(mod, context(mod), name)
+        return builderF(mod, LLVM.context(mod), name)
     end
 end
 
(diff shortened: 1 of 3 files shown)


src/compiler.jl
```diff
@@ -2018,26 +1984,67 @@ function GPUCompiler.codegen(output::Symbol, job::CompilerJob{<:EnzymeTarget};
...
+            actualRetType = k.ci.rettype
+        end
 
+        Base.isbindingresolved(jlmod, name) && isdefined(jlmod, name) || continue
+        func = getfield(jlmod, name)
         sparam_vals = mi.specTypes.parameters[2:end] # mi.sparam_vals
-
-        if func == Base.println
+        
+        if func == Base.println || func == CUDA.context || func == CUDA.context! || func == CUDA.capability
```
vchuravy, line 2043: We need this to work without adding a dependency on `CUDA`

vchuravy (changes requested):

src/compiler.jl
```diff
@@ -2160,6 +2235,7 @@ function GPUCompiler.codegen(output::Symbol, job::CompilerJob{<:EnzymeTarget};
         adjointf = primalf
         augmented_primalf = nothing
     end
+    NameToMethod.clear()
```
vchuravy, line 2238: `empty!(NameToMerhod)`

PR #156, opened 2021-11-24
Fix GC return error of generic calls

### LINKED ISSUE #155: GC error
```julia
using Enzyme

Enzyme.API.printall!(true)

t = 0.0

function tobedifferentiated(cond, a)::Float64
    if cond
        t + t
    else
        0.0
    end
end

autodiff(tobedifferentiated, true, Active(2.1))
```

wsmoses:
Resolves #155

Requires https://github.com/wsmoses/Enzyme/pull/383 (and jll bump)

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -447,6 +447,30 @@ end
     @test 1.0 ≈ autodiff(foo, false, Active(2.14))[1]
 end
 
+@testset "Return GC error" begin
+	t = 0.0
+
+	function tobedifferentiated(cond, a)::Float64
+		if cond
+			t + t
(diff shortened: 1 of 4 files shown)


test/runtests.jl
```diff
@@ -448,6 +448,30 @@ end
...
+			a + t
+		else
+			0.0
+		end
+	end
+
+	@test 1.0 ≈ autodiff(tobedifferentiated2, true, Active(2.1))[1]
+end
+
+@testset "Return GC error" begin
```
vchuravy, line 474: Is this the right name for the test below?

PR #173, opened 2022-01-16
Support LLVM 13

vchuravy:
- sret is now a TypedAttribute
- can't create a PhiNode from token


vchuravy:
Still not fully fixed, but improves the situation.

PR #182, opened 2022-01-26
Test against  Enzyme main

vchuravy:
- add Preference propagation to test
- Test against locally built Enzyme


wsmoses (approved):
Reflect

PR #187, opened 2022-01-27
Set default activity of any

wsmoses:
Doesn't fix default for https://github.com/wsmoses/Enzyme.jl/issues/185, but will result in an error (since duplicated)


vchuravy:
Any idea, why this fails CI?

vchuravy (approved):

PR #209, opened 2022-02-02
[0.8] Warn on non definite number of fields instead of failing

vchuravy:


test/typetree.jl
```diff
@@ -0,0 +1,19 @@
...
+
+import Enzyme: typetree, TypeTree, API
+
+const ctx = LLVM.Context()
+const dl = string(LLVM.DataLayout(LLVM.JITTargetMachine()))
+
+tt(T) = string(typetree(T, ctx, dl))
+
+@testset "TypeTree" begin
+    @test tt(Float16) == "{[0]:Float@half}" # TODO(VC) inconsistent
```
vchuravy, line 13: @wsmoses this should probably be `[-1]` as well?

wsmoses: It should be, yes

PR #224, opened 2022-02-06
Handle sincos

wsmoses:
Should fix both https://github.com/wsmoses/Enzyme.jl/issues/219 and https://github.com/wsmoses/Enzyme.jl/issues/99

### DIFF
--- a/src/compiler/interpreter.jl
+++ b/src/compiler/interpreter.jl
@@ -70,11 +70,13 @@ end
 function is_primitive_func(@nospecialize(TT))
     isa(TT, DataType) || return false
     ft = TT.parameters[1]
+    if ft === typeof(Base.string) 
+       return true
+    end
     if ft === typeof(Base.cbrt) || ft === typeof(Base.sin) || ft === typeof(Base.cos) ||
        ft === typeof(Base.tan) || ft === typeof(Base.exp) || ft === typeof(Base.log) ||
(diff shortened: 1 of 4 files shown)


src/compiler/interpreter.jl
```diff
@@ -70,11 +70,13 @@ end
 function is_primitive_func(@nospecialize(TT))
     isa(TT, DataType) || return false
     ft = TT.parameters[1]
+    if ft === typeof(Base.string)
```
vchuravy, line 73: This was intentionally only for `string` with FloatingPoint args

vchuravy (approved):

src/compiler/interpreter.jl
wsmoses: I mean regardless this is inactive and we want to mark as such?

wsmoses:
@vchuravy the remaining failure appears to be unrelated due to a change in SpecialFunctions.jl

Good to merge?

src/compiler/interpreter.jl
vchuravy: Yes but we don't want to change inlining behavior more than necessary

vchuravy:
Reproducer:

```
using Enzyme
using SpecialFunctions

Enzyme.autodiff((y) -> SpecialFunctions.bessely(2, y), Active(1))
```

PR #248, opened 2022-03-05
Cleanup FFI support

vchuravy:


src/compiler.jl
```diff
@@ -1953,18 +1953,18 @@ end
 
 function __init__()
     API.EnzymeSetHandler(@cfunction(julia_error, Cvoid, (Cstring,)))
-    API.EnzymeRegisterAllocationHandler(
-        "jl_alloc_array_1d",
+    register_handler!(
```
wsmoses, line 1956: RegisterAllocationHandler is distinct from registerhandler

Issue #278, opened 2022-04-13
Out of place broadcasting creates unexpected results


ToucheSir:
MWE from Slack:

```julia
using Enzyme, LinearAlgebra, Zygote

x = Float32[3, 2, 1]
w = Float32[1, 2, 3]
dw = zero(w)

loss(w, x) = sum(w .* x)

Enzyme.autodiff(loss, Active, Duplicated(w, dw), Const(x))
zdw = gradient(w -> loss(w, x), w) |> only
@show w, dw, zdw
@assert isapprox(dw, zdw; atol=1e-3)
```

Output:
```
(w, dw, zdw) = (Float32[1.0, 2.0, 3.0], Float32[3.0, 2.0, 1.0], Float32[4.0, 4.0, 4.0])
ERROR: AssertionError: isapprox(dw, zdw; atol = 0.001)
...
```

wsmoses:
Just for fun can you also print w

ToucheSir:
I didn't save the original seed, but this repros with fixed inputs as well so I've updated the MWE to match.

wsmoses:
This feels like what was fixed in https://github.com/EnzymeAD/Enzyme/pull/604

I’ll try it locally to see if it resolves. If so I’ll do a jll bump with it and the fix for https://github.com/EnzymeAD/Enzyme.jl/pull/277 once ready

wsmoses:
Reducing a bit:

```
using Enzyme
Enzyme.API.printall!(true)
x = Float32[3]
w = Float32[1]
dw = zero(w)

# loss(w, x) = @inbounds (w .* x)[1]

loss(w, x) = @inbounds Base.materialize(Base.broadcasted(*,w,x))[1]

@show w, dw, x
Enzyme.autodiff(loss, Active, Duplicated(w, dw), Const(x))

@show w, dw, x
# (w, dw, x) = (Float32[1.0], Float32[3.0], Float32[4.0])
```

Incidentally the bug here is not that it computes the wrong derivative (it gets the right one) but somewhere activity analysis incorrectly believes x to be active and +='s its derivative into x.

Incidentally making x duplicated should get you the desired outcome. This also requires the broadcast and works fine without it. Looking into it here (https://fwd.gymni.ch/H2S9ke).

@vchuravy this relates to our earlier convo.

wsmoses:
Reducing further

```
using Enzyme
Enzyme.API.printall!(true)
x = Float32[3]
w = Float32[1]
[...]

PR #279, opened 2022-04-16
Inactive arg copy

wsmoses:
Requires https://github.com/EnzymeAD/Enzyme/pull/614

Fixes https://github.com/EnzymeAD/Enzyme.jl/issues/278

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1133,7 +1133,7 @@ end
 	  @inbounds w[1] * x[1]
 	end
 
-	Enzyme.autodiff(loss, Active, Duplicated(w, dw), Const(x), Const(false))
+	Enzyme.autodiff(inactiveArg, Active, Duplicated(w, dw), Const(x), Const(false))
 
     @test x ≈ [3.0]
     @test w ≈ [1.0]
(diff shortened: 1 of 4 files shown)


test/runtests.jl
```diff
@@ -835,3 +835,41 @@ end
...
+	end
+
+	Enzyme.autodiff(loss, Active, Duplicated(w, dw), Const(x), Const(false))
+
+    @test x ≈ [3.0]
+    @test w ≈ [1.0]
+    @test dw ≈ [3.0]
+
+    # It would be nice to get this right without enabling this
+    Enzyme.API.runtimeActivity!(true)
```
vchuravy, line 858: Should we always enable this for Julia?

vchuravy (approved):

test/runtests.jl
wsmoses: Not yet, it currently has a very large perf penalty. We should eventually have a three state, guaranteed active, guaranteed inactive, unknown and default to active (and check at runtime). Currently 1 and 3 are the same

vchuravy:
bump?

PR #290, opened 2022-04-27
Better error on unknown concrete type

vchuravy:


src/compiler.jl
```diff
@@ -1590,6 +1590,12 @@ function arraycopy_common(fwd, B, orig, origArg, gutils)
 	data0!(tt)
     ct = API.EnzymeTypeTreeInner0(tt)
 
+    if ct == API.DT_Unknown
+        @warn "Unknown concrete type" tt=string(tt)
```
wsmoses, line 1594: Use safe_warn ?

vchuravy, line 1594:
```suggestion
        GPUCompiler.@safe_warn "Unknown concrete type" tt=string(tt)
```

PR #299, opened 2022-05-10
Move compilation out of threadsfor

wsmoses:
Has a remaining bug in reverse mode, and prints need removal.


src/compiler.jl
```diff
@@ -1223,168 +1215,211 @@ function nested_codegen!(mod::LLVM.Module, f, tt)
...
+    function rev()
+        ntape = unsafe_load(tapes, Base.Threads.threadid())
+        @show "rev", ntape, Base.Threads.threadid(), tapes
+        flush(stdout)
+        @show InteractiveUtils.@code_llvm thunk(ntape)
+        flush(stdout)
+        thunk(ntape)
+    end
+    Base.Threads.threading_run(rev)
+    @show "post run", thunk
```
wsmoses, line 1256: Weirdly enough introducing this show/flush removes one of the gc segfaults, though the result remains incorrect -- as if the reverse thunk code was never run

wsmoses:
@vchuravy 

```julia
wmoses@beast:~/git/Enzyme.jl (rtest) $ cat thread.jl
using Enzyme
Enzyme.API.printperf!(true)
function foo(y, s, e)
        Base.Threads.threading_run() do
            @inbounds y[s] *= 2
        end
        nothing
end

x = [1.0, 2.0, 3.0]

dx = [1., 1.0, 1.0]
Enzyme.API.printall!(true)
@show Enzyme.autodiff(foo, Duplicated(x, dx), 1, 1)
@show x, dx
```

vchuravy:
Failed for me on 1.7.1 with:

```
ERROR: LoadError: MethodError: Cannot `convert` an object of type Core.LLVMPtr{UInt8, 0} to an object of type Ptr{Nothing}
Closest candidates are:
  convert(::Type{Ptr{T}}, ::Union{Int64, UInt64}) where T at ~/builds/julia-1.7.1/share/julia/base/pointer.jl:26
  convert(::Type{Ptr{T}}, ::Ptr{T}) where T at ~/builds/julia-1.7.1/share/julia/base/pointer.jl:29
  convert(::Type{Ptr{T}}, ::Ptr) where T at ~/builds/julia-1.7.1/share/julia/base/pointer.jl:30
  ...
Stacktrace:
 [1] Enzyme.Compiler.Tape(thunk::Enzyme.Compiler.AdjointThunk{typeof(Base.Threads.threading_run), Const{Nothing}, Tuple{Duplicated{var"#1#2"{Vector{Float64}, Int64}}}, Val{0x0000000000000001}, Nothing}, internal_tape::Core.LLVMPtr{UInt8, 0}, shadow_return::Nothing, resT::Type)
   @ Enzyme.Compiler ~/src/Enzyme/src/compiler.jl:304
in expression starting at /home/vchuravy/src/Enzyme/thread.jl:18
```

vchuravy:
Replaced by #342

PR #307, opened 2022-05-12
fix doctests

vchuravy:


docs/src/index.md
```diff
@@ -4,7 +4,7 @@ CurrentModule = Enzyme
 
 # Enzyme
 
-Documentation for [Enzyme.jl](https://github.com/wsmoses/Enzyme.jl), the Julia bindings for [Enzyme](https://github.com/wsmoses/enzyme).
+Documentation for [Enzyme.jl](https://github.com/EnzymeAD/Enzyme.jl), the Julia bindings for [Enzyme](https://github.com/EnzymeADenzyme).
```
vchuravy, line 7:
```suggestion
Documentation for [Enzyme.jl](https://github.com/EnzymeAD/Enzyme.jl), the Julia bindings for [Enzyme](https://github.com/EnzymeAD/enzyme).
```

vchuravy, line 7:
```suggestion
Documentation for [Enzyme.jl](https://github.com/EnzymeAD/Enzyme.jl), the Julia bindings for [Enzyme](https://github.com/EnzymeAD/enzyme).
```

src/Enzyme.jl
```diff
@@ -798,18 +779,16 @@ of this is the transpose of the Forward [`jacobian`](@ref)
...
-using Enzyme
-
 f(x) = [x[1]*x[2], x[2]]
 
 grad = jacobian(Reverse, f, [2.0, 3.0])
 
 # output
 
-(([3.0, 2.0], [0.0, 1.0]),)
+([3.0, 2.0],)
```
vchuravy, line 788: @wsmoses that this didn't work worries me. Should this be `jacobian(Reverse, f, [2.0, 3.0]; n_outs=Val(2))`

wsmoses: It should absolutely be n_outs=2 and that should also absolutely NOT have a default value. It needs to be precisely the length of the returned array (here 2).

vchuravy: We will need to present a better error message to users then. Keyword values are supposed to have a default value

wsmoses: Make it a required arg? I just was worried that having two args would be confusing/lead to mis ordering

PR #313, opened 2022-05-17
0.9: Mark default_rng inactive

vchuravy:


wsmoses:
Perhaps it would be useful to merge and backport the inactive list PR, and add this to that?

vchuravy:
```
f(x)=x*rand()
```

is fixed by this,

but

```
f(x)=x*sum(randn(64))
```

Is not.

The actual issue us that

````
No augmented forward pass found for malloc
declare i64 @malloc(i64) local_unnamed_addr
```

Created by `Libc.malloc`

vchuravy:
> Perhaps it would be useful to merge and backport the inactive list PR, and add this to that?

That's a rather large refactor

PR #321, opened 2022-05-25
Add tests for random numbers

vchuravy:
Add tests for rand/randn


vchuravy:
So `randn` fails on 1.8 with a type-inference issue @wsmoses  and on 1.6 it fails since we want do differentiate through seeding

vchuravy:
https://gist.github.com/vchuravy/ba8a055321fc1f4a8fcfd265650bd465 for the randn issue

vchuravy:
replaced by #333

PR #296, opened 2022-05-04
Generalize activity sets

wsmoses:

PR #327, opened 2022-05-27
Fix in to haskey

wsmoses:


vchuravy (approved):

vchuravy:
Depends on #296 and thus does not apply to 0.9

PR #331, opened 2022-05-28
disable broken test on 1.8

vchuravy:
@wsmoses can you take a look at `threadsfor_fwd`? `vals` is not defined and I am unsure what you wanted to write there

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -569,6 +569,8 @@ end
     @test 5.0 ≈ dx[1]
 end
 
+# FIXME: Enzyme: Unknown concrete type in arraycopy_common
+if VERSION < v"1.8.0-"
 @testset "broadcast" begin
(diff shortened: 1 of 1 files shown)


wsmoses:
This should be fixed in https://github.com/EnzymeAD/Enzyme.jl/pull/299 once that lands (eg once the GC bug is tracked down and fixed)

vchuravy:
> (eg once the GC bug is tracked down and fixed)

Are the fixes separable? Also can you post a MWE for the GC bug?

wsmoses:
Oh sorry, threadsfor_fwd should be fixed by https://github.com/EnzymeAD/Enzyme.jl/pull/299, this is distinct. Going to look at this right now to see if there's an easy fix.

wsmoses:
Subsumed by https://github.com/EnzymeAD/Enzyme.jl/pull/335

PR #333, opened 2022-05-28
Rand

wsmoses:


src/compiler.jl
```diff
@@ -69,7 +70,10 @@ const InactiveFunctions = Set([Base.CoreLogging.logmsg_code,
                                Base.eps,
                                Base.nextfloat,
                                Base.prevfloat,
-                               Core.kwfunc
+                               Core.kwfunc,
+                               Random.rand,
```
vchuravy, line 74: I am generally worried about these very generic opt-outs. I suspect if I let @femtomc think hard enough he could come up with a scenario where one would need to take the derivative through the state of a sampler, or something like that.

But let's cross that bridge later.

vchuravy (approved):

Project.toml
```diff
@@ -13,6 +13,8 @@ Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 ObjectFile = "d8793406-e978-5875-9003-1fc021f44a92"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+ValueShapes = "136a8f8c-c49b-4edb-8b98-f3d64d48be8f"
```
vchuravy, line 17: Why was this added?

wsmoses: ValueShapes I don't know and am confused by. Random is so we mark the functions inactive?

PR #245, opened 2022-03-04
Add split mode

vchuravy:
```
using Enzyme

julia> f(x) = x^2
f (generic function with 1 method)

julia> ((tape, res), backward) = Enzyme.forward_back(f, Active, Active(2.0))
((Ptr{Nothing} @0x0000000003da4c80, 4.0), Enzyme.Compiler.AdjointThunk{typeof(f), Active{Float64}, Tuple{Active{Float64}}, Nothing}(f, Ptr{Nothing} @0x00007fc41b835060, nothing))

julia> backward(Active(2.0), 1.0, tape)
(4.0,)
```

~@wsmoses I don't understand why I have to pass `Active(1.0)` to the backwards pass. I expected it to be `Active(2.0)`~

PR #334, opened 2022-05-29
Add EnzymeCore changes for split mode

vchuravy:
@wsmoses what do you think about this as an API?

### DIFF
--- a/src/Enzyme.jl
+++ b/src/Enzyme.jl
@@ -30,7 +30,7 @@ convert(::Type{API.CDerivativeMode}, ::ForwardMode) = API.DEM_ForwardMode
 
 guess_activity(::Type{T}, mode::Mode) where T = guess_activity(T, convert(API.CDerivativeMode, mode))
 
-@inline function guess_activity(::Type{T}, Mode::API.CDerivativeMode=API.DEM_ReverseModeCombined) where {T}
+@inline function guess_activity(::Type{T}, Mode::API.CDerivativeMode) where {T}
     if T isa Union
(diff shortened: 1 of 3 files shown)


wsmoses:
@vchuravy how would you get the augmented fwd as distinct from gradient in this case?

vchuravy:
So in #245 I would evaluate the augmented primal immediatly and then return the gradient thunk to the user.

It might generally be more flexible to provide as a base API just one that returns the thunks.

wsmoses:
Okay, I'm okay with this. The other two things we may want to include in this:
* whether or not to return the original return.
* whether or not to free in the reverse (or forward split).

wsmoses (approved):

src/Enzyme.jl
```diff
@@ -21,12 +21,15 @@ export EnzymeRules
...
 import LLVM
 include("api.jl")
 
+convert(::Type{API.CDerivativeMode}, ::ReverseMode{false}) = API.DEM_ReverseModeCombined
+convert(::Type{API.CDerivativeMode}, ::ReverseMode{true}) = API.DEM_ReverseModeGradient
+convert(::Type{API.CDerivativeMode}, ::ForwardMode) = API.DEM_ForwardMode
+
+guess_activity(::Type{T}, mode::Mode) where T = guess_activity(T, convert(API.CDerivativeMode, mode))
+
 @inline function guess_activity(::Type{T}, Mode::API.CDerivativeMode=API.DEM_ReverseModeCombined) where {T}
```
vchuravy, line 33:
```suggestion
@inline function guess_activity(::Type{T}, Mode::API.CDerivativeMode) where {T}
```

PR #343, opened 2022-05-30
Use LLVM.@dispose instead of do-block resource handling

vchuravy:
Replaces the remaining `ModulePassManager() do` and `ThreadSafeContext() do` blocks with `LLVM.@dispose`, which does the same acquire/release via `try`/`finally` without allocating a closure per call.

The `GPUCompiler.JuliaContext() do` sites are deliberately left as-is: the do-form hands the body the inner `LLVM.Context` and activates it, while the no-arg constructor returns an unactivated `ThreadSafeContext`, so the two are not interchangeable.

(Rebase of a stale 2022 PR; most of the original diff targeted the legacy pass manager and no longer applies.)

### DIFF
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -5429,7 +5429,7 @@ function lower_convention(
         LLVM.run!(pb, mod)
     end
 
-    ModulePassManager() do pm
+    @dispose pm = ModulePassManager() begin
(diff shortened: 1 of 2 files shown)


wsmoses:
@vchuravy rebase?

PR #351, opened 2022-06-01
Jacobian returns one array

vchuravy:


vchuravy:
Ok now we are in agreement with ForwardDiff.jl

src/Enzyme.jl
```diff
@@ -780,6 +787,7 @@ end
         adjoint(Duplicated(x, dx), tape)
         return dx
     end
+    mapreduce(LinearAlgebra.adjoint, vcat, rows)
```
wsmoses, line 790: Kind of hacky, but would it be possible to create the zero init'd matrix, then unsafe_array it at the corresponding offset for each index to avoid making a bunch of temporaries

test/runtests.jl
```diff
@@ -984,12 +987,38 @@ end
...
+
+    @test J_r_1(A, x) == [
+        1.0  1.0  0.0  0.0  0.0  0.0;
+        1.0  0.0  1.0  0.0  0.0  0.0;
+        1.0  0.0  0.0  1.0  0.0  0.0;
+        1.0  0.0  0.0  0.0  1.0  0.0;
+        1.0  0.0  0.0  0.0  0.0  1.0;
+    ]
+
+    @test J_r_2(A, x) == [
```
vchuravy, line 999:
```suggestion
    @test_broken J_r_2(A, x) == [
```

PR #361, opened 2022-06-16
Make same or 1 recursive

wsmoses:


src/Enzyme.jl
```diff
@@ -187,31 +187,30 @@ import .Compiler: CompilationException
...
-    end
-
+@inline function same_or_one_helper(current, next)
     if current == -1
-        current = 1
+        return next
+    elseif current == next
+        return next
+    else
+        @assert False, "Multiple distinct batch sizes"
```
vchuravy, line 196: Just use `error` also `False` is a python-ism

vchuravy (approved):

PR #384, opened 2022-07-21
Add dual annotation

tgymnich:


vchuravy (approved):

vchuravy:
Maybe we should update the examples and tests?

wsmoses (changes requested):
Having another name without a semantic difference feels more likely to cause confusion.

If we're going to introduce dual we should forbit the use of duplicated in forward mode and the use of dual in reverse.

Issue #375, opened 2022-07-05
Differentiating trigonometric functions with KernelGradients and CUDAKernels


michel2323:
The following code errors with
```
cannot handle (augmented) unknown intrinsic
  %1 = call i32 @llvm.nvvm.d2i.hi(double %a) #14
```

when dispatched on a `CUDADevice`. So far confirmed for `sin` and `cos`.

```julia
using CUDA
using KernelAbstractions
using CUDAKernels
using KernelGradients
using Enzyme

@kernel function cos_kernel!(y, x)
    I = @index(Global, Linear)

    y[I] = cos(x[I])
end

x = ones(10)
dx = zeros(10)
y = zeros(10)
dy = zeros(10)
cux = CuArray(x)
cuy = CuArray(y)
cudx = CuArray(dx)
cudy = CuArray(dy)

ev = autodiff(cos_kernel!(CPU()))(
    Duplicated(y, dy), Duplicated(x, dx), ndrange=length(x)
)
wait(ev)

# Breaks
ev = autodiff(cos_kernel!(CUDADevice()))(
    Duplicated(cuy, cudy), Duplicated(cux, cudx), ndrange=length(x)
)
wait(ev)
```

classner:
Similarly, I'm encountering problems for:
* max (Float32):
   ```
   cannot handle (reverse) unknown intrinsic
   llvm.nvvm.fmax.ftz.f
      %47 = call float @llvm.nvvm.fmax.ftz.f(float %44, float noundef 0.000000e+00) #6, !dbg !158
  ```
* abs (Float32):
  ```
  cannot handle (reverse) unknown intrinsic
  llvm.nvvm.fabs.ftz.f
     %47 = call float @llvm.nvvm.fabs.ftz.f(float %44) #6, !dbg !158
  ```
* exp (Float32):
   ```
   cannot handle (reverse) unknown intrinsic
   llvm.nvvm.fma.rn.ftz.f
     %49 = call float @llvm.nvvm.fma.rn.ftz.f(float %46, float noundef 0x3F777313A0000000, float noundef 5.000000e-01) #7, !dbg !209
   ```

wsmoses:
@vchuravy fixing this basically just requires running the preservenvvm pass prior to enzyme for the libbc like we do in clang for other languages.

I can take a stab at this next week.

vchuravy:
Should be fixed by #379

wsmoses:
[...]

PR #395, opened 2022-07-31
fix cuda tests

wsmoses:
Fixes remaining bugs in https://github.com/EnzymeAD/Enzyme.jl/issues/375

### DIFF
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -1282,7 +1282,7 @@ function nested_codegen!(mod::LLVM.Module, f, tt)
     params = Compiler.PrimalCompilerParams()
     job    = CompilerJob(target, funcspec, params)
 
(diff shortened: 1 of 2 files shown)


src/compiler.jl
```diff
@@ -1282,7 +1282,7 @@ function nested_codegen!(mod::LLVM.Module, f, tt)
     params = Compiler.PrimalCompilerParams()
     job    = CompilerJob(target, funcspec, params)
 
-    otherMod, meta = GPUCompiler.codegen(:llvm, job; optimize=false, cleanup=false, validate=false, ctx)
+    otherMod, meta = GPUCompiler.codegen(:llvm, job; optimize=false, validate=false, ctx)
```
vchuravy, line 1285: Don't we need cleanup=false here as well?

PR #396, opened 2022-08-01
Custom allocator support

vchuravy:
- Add API for custom allocator
- add dummy allocator/deallocator

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -6,6 +6,7 @@ version = "0.10.4"
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
 CEnum = "fa961155-64e5-5f13-b03f-caf6b980ea82"
+CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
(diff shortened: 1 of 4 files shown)


Project.toml
```diff
@@ -6,6 +6,7 @@ version = "0.10.4"
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
 CEnum = "fa961155-64e5-5f13-b03f-caf6b980ea82"
+CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
```
vchuravy, line 9:
```suggestion
```

src/compiler/utils.jl
```diff
@@ -1,22 +1,30 @@
 
-function get_function!(builderF, mod, name)
```
vchuravy, line 2: The goal of the design here was that we didn't need to create the FunctionType everytime.

wsmoses: We do, however, regardless, need the FT to ensure the right type is returned if the current function has the wrong functiontype (eg returns i64 rather than i8*)

PR #410, opened 2022-08-11
Copy over correct IR dumping code

pchintalapudi:
This should fix the IR dumping code by cribbing the relevant lines from https://github.com/JuliaGPU/GPUCompiler.jl/pull/324, which fixed dumping in GPUCompiler.

### DIFF
--- a/src/compiler/reflection.jl
+++ b/src/compiler/reflection.jl
@@ -28,15 +28,42 @@ function reflect(@nospecialize(func), @nospecialize(A), @nospecialize(types);
     return llvmf, mod
 end
 
+# For VERSION >= v"1.9.0-DEV.516"
+struct jl_llvmf_dump
+    TSM::LLVM.API.LLVMOrcThreadSafeModuleRef
+    F::LLVM.API.LLVMValueRef
+end
+
 function enzyme_code_llvm(io::IO, @nospecialize(func), @nospecialize(A), @nospecialize(types);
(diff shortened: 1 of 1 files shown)


src/compiler/reflection.jl
```diff
@@ -28,15 +28,42 @@ function reflect(@nospecialize(func), @nospecialize(A), @nospecialize(types);
...
+                    value = Ref(jl_llvmf_dump(ts_mod.ref, entry_fn.ref))
+                    str = ccall(:jl_dump_function_ir, Ref{String},
+                          (Ptr{jl_llvmf_dump}, Bool, Bool, Ptr{UInt8}),
+                          value, !raw, dump_module, debuginfo)
+                end
+            else
+                GC.@preserve ts_mod entry_fn begin
+                    # N.B. jl_dump_function_ir will `Libc.free` the passed-in pointer
+                    value_ptr = reinterpret(Ptr{jl_llvmf_dump},
+                                            Libc.malloc(sizeof(jl_llvmf_dump)))
```
wsmoses, line 55: Do we not need to free this later? Alternatively, could a stack allocation not suffice?

pchintalapudi: Unfortunately not, it's a consequence of the underlying function deleting the pointer in DEV.516. This was rectified in DEV.672, but compatibility in between is handled by this block.

wsmoses (approved):

Issue #347, opened 2022-05-31
Pseudo-zero initialize non-differential values for sret


SBuercklin:
AD with respect to a struct that has a non-differentiable field fails

```julia
struct Foo2{X,Y}
    x::X
    y::Y
end

test_f(f::Foo2) = f.x^2

julia> autodiff(test_f, Active(Foo2(3.0, 2.0)))
(Foo2{Float64, Float64}(6.0, 0.0),)

julia> autodiff(test_f, Active(Foo2(3.0, :two)))
ERROR: UndefRefError: access to undefined reference
Stacktrace:
 [1] getproperty
   @ ./Base.jl:42 [inlined]
 [2] getindex
   @ ./refvalue.jl:56 [inlined]
 [3] macro expansion
   @ ~/.julia/packages/Enzyme/fgrLi/src/compiler.jl:3641 [inlined]
 [4] enzyme_call
   @ ~/.julia/packages/Enzyme/fgrLi/src/compiler.jl:3434 [inlined]
 [5] CombinedAdjointThunk
   @ ~/.julia/packages/Enzyme/fgrLi/src/compiler.jl:3416 [inlined]
 [6] autodiff
   @ ~/.julia/packages/Enzyme/fgrLi/src/Enzyme.jl:202 [inlined]
 [7] autodiff(f::typeof(test_f), args::Active{Foo2{Float64, Symbol}})
   @ Enzyme ~/.julia/packages/Enzyme/fgrLi/src/Enzyme.jl:227
 [8] top-level scope
   @ REPL[12]:1

```
This is on `1.7.3` with both `Enzyme@0.9.6` and the current `Enzyme#main`

vchuravy:
So this works with forward-mode, but not reverse-mode :/ The issue is that we want to zero initialize this struct which is not valid due to the symbol (at least that's my guess).

```
julia> autodiff(Forward, test_f, Duplicated(Foo2(3.0, :two), Foo2(1.0, :two)))
(6.0,)
```

wsmoses:
@vchuravy that is indeed the cause. Specifically, what happens from looking at the IR (below) is as follows:

```
res = { 0.0, null }
res[1] += 2 * f.x
```

And since null as a symbol creates the segfault. Since the symbol isn't initialized in the function, there's no shadow store that Enzyme would also use here. There are several potential remedies, but I think the immediate one is to say that this is improper as a usage of Active in contrast to Duplicated.

```
[...]

PR #422, opened 2022-08-20
Attempt to provide a better error for undef sret

vchuravy:
I was thinking if we can find which values weren't initalized by Enzyme for the shadow
sret we could fix them up with fictious values instead of Julia throwing an error, but
I haven't found a solution yet since for some values we can't take a pointer `Symbol`...

x-ref: #347

### DIFF
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -4992,6 +4992,33 @@ end
 
 Base.unsafe_convert(::Type{Ptr{T}}, b::Box{T}) where T = Base.unsafe_convert(Ptr{T}, Base.unsafe_convert(Ptr{Cvoid}, b))
 
+function detect_undef_values(box::Box{T}) where T
+    if isassigned(box)
+        return
+    end
+    ptr = Base.unsafe_convert(Ptr{Cvoid}, box)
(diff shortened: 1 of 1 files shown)


vchuravy:
I think the better soution would be to zero-initialize the sret (instead of undef as it is currently) and then define something like.

`zeroof(::Symbol) = :Zero`

wsmoses:
So I thought about this a bit more, and I think there's a reasonably nice solution here. Namely, this error can only occur if an active input value contains an allocatedinline value for whom bits of 0's are not legal.

Since we know the proper julia type, and this only is an issue on the outermost level where there's an ABI interface from Enzyme to Julia (already wrapped), prior to returning from the llvmcall we create, we can initialize any allocatedinline types with placeholders of choice.

wsmoses:
My comment/proprosal above wouldn't just be a better error, but in fact a fix.

wsmoses:
Zero init'ing the sret wouldn't be sufficient, you'd need to post copying the result from the Enzyme struct, or alternatively only copy the relevant parts of the LLVM struct.

vchuravy:
>  we can initialize any allocatedinline types with placeholders of choice.

Can you implement that?

PR #433, opened 2022-08-29
Fix nightly

wsmoses:


src/compiler.jl
```diff
@@ -1274,9 +1395,20 @@ function duplicate_rev(B::LLVM.API.LLVMBuilderRef, OrigCI::LLVM.API.LLVMValueRef
     return nothing
 end
 
+@static if VERSION < v"1.9-"
+else
+const ctxToThreadSafe = Dict{LLVM.Context, LLVM.ThreadSafeContext}()
```
vchuravy, line 1400: I am not a fan of this change. @pchintalapudi it looks like we might need to pass TSCtx through in more places in GPUCompiler.jl?

wsmoses: The issue here is that given an LLVM module, it is no longer sufficient to be able to call `GPUCompiler.codegen` from within a custom rule method. This, as far as I can tell, cannot be derived from the existing module/etc and thus, without modifying each enzyme call to pass in extra data (and say hold the thread safe context), we need an outside-of-api way of ensuring this information is available -- hence the dict.

pchintalapudi: If the issue is that the module can't give a TSCtx, then the module should probably become a TSModule, which does carry a TSCtx shared ptr and will manage its memory correctly.

wsmoses: The Module isn't provided explicitly, it is derived from an LLVM instruction. Thus it cannot become the TSModule.

Issue #309, opened 2022-05-12
MethodError on dynamic `Val` construction


ToucheSir:
MWE:
```julia
using Enzyme

f(::Val{D}) where D = prod(D)
mwe(x, t) = x / f(Val(t))

Enzyme.autodiff(mwe, Active, Active(1.0), Const((1, 2)))
```
Results in:
```
ERROR: LoadError: MethodError: no method matching DataType()
Stacktrace:
 [1] macro expansion
   @ ~/.julia/packages/GPUCompiler/EV8pB/src/cache.jl:0 [inlined]
 [2] specialization_id(job::GPUCompiler.CompilerJob{Enzyme.Compiler.EnzymeTarget, Enzyme.Compiler.EnzymeCompilerParams, GPUCompiler.FunctionSpec{DataType, Tuple{}}})
   @ GPUCompiler ~/.julia/packages/GPUCompiler/EV8pB/src/cache.jl:12
 [3] thunk
   @ ~/.julia/dev/Enzyme/src/compiler.jl:4462 [inlined]
 [4] thunk(f::Type{Val{(1, 2)}}, df::Nothing, ::Type{Const{Val{(1, 2)}}}, tt::Type{Tuple{}}, ::Val{Enzyme.API.DEM_ReverseModePrimal}, ::Val{0x0000000000000001})
   @ Enzyme.Compiler ~/.julia/dev/Enzyme/src/compiler.jl:4456
in expression starting at /home/brianc/projects/juliamwes/enzyme/mwe.jl:6
```
The above was reduced from https://github.com/FluxML/NNlib.jl/blob/master/src/impl/conv_direct.jl#L47-L56, where we use this pattern + a function barrier for type stability.

vchuravy:
On main I get a slightly better answer:

```
julia> Enzyme.autodiff(mwe, Active, Active(1.0), Const((1, 2)))
ERROR: MethodError: no method matching typeof(mwe)(::Float64, ::Tuple{Int64, Int64})
Stacktrace:
 [1] macro expansion
   @ ~/.julia/packages/GPUCompiler/1FdJy/src/cache.jl:0 [inlined]
 [2] specialization_id(job::GPUCompiler.CompilerJob{Enzyme.Compiler.EnzymeTarget, Enzyme.Compiler.EnzymeCompilerParams, GPUCompiler.FunctionSpec{DataType, Tuple{Float64, Tuple{Int64, Int64}}}})
[...]

PR #438, opened 2022-09-04
Dynamic val fix

wsmoses:
Fixes https://github.com/EnzymeAD/Enzyme.jl/issues/309


vchuravy:
Can we please keep PR at one fix at a time? It is not clear that the batch generic issue is related to the original issue aimed to be fixed, making reviewing much harder.

PR #439, opened 2022-09-05
Handle not concrete type trees

wsmoses:


vchuravy:
Should also have added a test where we return `Tuole{Any, Any}` to check that indeed this is the right representation?

Issue #337, opened 2022-05-29
Alloc -> Malloc downgrade on sret


wsmoses:
Suppose you have something like

```
%z = alloca { {} addr(10)*, {} addr(10)* }
call f(%z sret(...) )
load %z[0]
load %z[1]
```

This ends up deciding to upgrade %z to a malloc to preserve all at once for the reverse, rather than separately cache %z[0]/%z[1]. However, this means that the gc frame lowering fails since it assumes an alloca is what is used.

```
julia: /buildworker/worker/package_linux64/build/src/llvm-late-gc-lowering.cpp:1463: State LateLowerGCFrame::LocalScan(llvm::Function&): Assertion `SRet' failed.

signal (6): Aborted
in expression starting at /mnt/Data/git/Enzyme.jl/lux.jl:18
gsignal at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
abort at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
unknown function (ip: 0x7f2c0c2713f9)
__assert_fail at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
LocalScan at /buildworker/worker/package_linux64/build/src/llvm-late-gc-lowering.cpp:1463
runOnFunction at /buildworker/worker/package_linux64/build/src/llvm-late-gc-lowering.cpp:2675
_ZN4llvm13FPPassManager13runOnFunctionERNS_8FunctionE at /mnt/Data/git/Enzyme.jl/julia-1.7.2/bin/../lib/julia/libLLVM-12jl.so (unknown line)
_ZN4llvm13FPPassManager11runOnModuleERNS_6ModuleE at /mnt/Data/git/Enzyme.jl/julia-1.7.2/bin/../lib/julia/libLLVM-12jl.so (unknown line)
_ZN4llvm6legacy15PassManagerImpl3runERNS_6ModuleE at /mnt/Data/git/Enzyme.jl/julia-1.7.2/bin/../lib/julia/libLLVM-12jl.so (unknown line)
LLVMRunPassManager at /mnt/Data/git/Enzyme.jl/julia-1.7.2/bin/../lib/julia/libLLVM-12jl.so (unknown line)
LLVMRunPassManager at /home/wmoses/.julia/packages/LLVM/gE6U9/lib/12/libLLVM_h.jl:4741 [inlined]
[...]

PR #440, opened 2022-09-05
Remove preprocessed funcs & fix typeof

wsmoses:


vchuravy (approved):
The change to Core.Typeof is sensible. Why do we need to remove the preprocessed function?

wsmoses:
The preprocessed function itself can have a call which triggers https://github.com/EnzymeAD/Enzyme.jl/issues/337
By removing it, any time that the error occurs, we can ensure it comes from the actual derivative code (as opposed to a auxillary data structure). Removing it actually caused the parameter test to compile!

Also, reducing the amount of code LLVM needs to post optimize in half is certainly a good thing for latency.

PR #446, opened 2022-09-08
Fix type `delf` -> `delF`

avik-pal:


wsmoses:
Can you add a test using array delete to trigger this and prevent a future regression?

vchuravy (approved):

PR #462, opened 2022-09-28
Fix recursive GC

wsmoses:
Note that this _will_ cause a performance regression for codes with recursion. But it will no longer compile assertion.

We should fix this later to restore the better performance without the assertion occurring in GC cases.


vchuravy:
1.6 and 1.7 were failing?

wsmoses:
They were failing in the jll released, this and the other PR at least make it debuggable since it fixes ABI expectations.

Nevertheless all Julia CI is now failing.

PR #478, opened 2022-10-03
Use CMake_jll so that users don't need to install it manually

vchuravy:


wsmoses:
@vchuravy relatedly do you know why julia clang on macos wasn't able to fnd the paths

giordano:
> relatedly do you know why julia clang on macos wasn't able to fnd the paths

Clang_jll knows nothing about system paths, you need to instruct it a little bit.  And macOS in particular has well hidden system paths.  You could try something `xcrun --sdk macosx --show-sdk-path` (from https://github.com/JuliaLang/julia/pull/47184#discussion_r998423119)

wsmoses:
fair, also relatedly @giordano we could also do something like this for reactant with bazel_jll on systems it's supported which might be nice for people given bazel nonsense to install

PR #489, opened 2022-10-05
Downstream Core tests on SciMLSensitivity

ChrisRackauckas:
These almost all hit Enzyme and stress test it. Core2 fails on latest release, this should test master.

### DIFF
--- a/.github/workflows/Downstream.yml
+++ b/.github/workflows/Downstream.yml
@@ -1,0 +1,57 @@
+name: IntegrationTest
+on:
+  push:
+    branches: [master]
+    tags: [v*]
+  pull_request:
+
+jobs:
+  test:
+    name: ${{ matrix.package.repo }}/${{ matrix.package.group }}/${{ matrix.julia-version }}
+    runs-on: ${{ matrix.os }}
+    env:
+      GROUP: ${{ matrix.package.group }}
+    strategy:
+      fail-fast: false
+      matrix:
+        julia-version: [1,1.6]
+        os: [ubuntu-latest]
+        package:
+          - {user: SciML, repo: SciMLSensitivity.jl, group: Core1}
+          - {user: SciML, repo: SciMLSensitivity.jl, group: Core2}
+          - {user: SciML, repo: SciMLSensitivity.jl, group: Core3}
+          - {user: SciML, repo: SciMLSensitivity.jl, group: Core4}
+          - {user: SciML, repo: SciMLSensitivity.jl, group: Core5}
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Bumping this. https://github.com/SciML/SciMLSensitivity.jl/pull/763 could've been prevented with the bigger test suite!

.github/workflows/Downstream.yml
```diff
@@ -0,0 +1,57 @@
...
+jobs:
+  test:
+    name: ${{ matrix.package.repo }}/${{ matrix.package.group }}/${{ matrix.julia-version }}
+    runs-on: ${{ matrix.os }}
+    env:
+      GROUP: ${{ matrix.package.group }}
+    strategy:
+      fail-fast: false
+      matrix:
+        julia-version: [1,1.6]
```
vchuravy, line 17: Only test `1.6`?

ChrisRackauckas: only v1 is probably more important than only v1.6, if the problem is number of tests.

wsmoses:
@ChrisRackauckas FYI on main we now have a slightly different calling convention as we prepare for 0.11.

ChrisRackauckas:
I couldn't tell from the docs. Are the dev version of the docs just not updated yet?

wsmoses:
They should be now, reload

ChrisRackauckas:
I still don't see what's different. Are there any notes that describes what changed?

wsmoses:
Mode is now a required argument, and non-active arguments in reverse mode will return a nothing in the tuple (such that the n-th element of the tuple corresponds to the n-th argument). Finally the derivative tuple in reverse mode is now wrapped in a second tuple (e.g. (derivative vals, ) ). This enables the new API for also having reverse mode return the original return of the function as well as the second tuple element.

https://enzymead.github.io/Enzyme.jl/dev/api/#Enzyme.autodiff-Union{Tuple{ReturnPrimal},%20Tuple{A},%20Tuple{F},%20Tuple{EnzymeCore.ReverseMode{ReturnPrimal},%20F,%20Type{A},%20Vararg{Any}}}%20where%20{F,%20A%3C:EnzymeCore.Annotation,%20ReturnPrimal}

vchuravy:
In general it would be great if we could identify a few relevant test to include as part of our CI suite, I am hesitant to take on the additional CI time.

ChrisRackauckas:
Paying for CI time is a lot cheaper than human time.

PR #498, opened 2022-10-08
Add EnzymeCore

vchuravy:
The idea is to have a place where we can put the type declaration such that people
can define their own rules without depending fully on Enzyme.

### DIFF
--- a/src/Enzyme.jl
+++ b/src/Enzyme.jl
@@ -1,136 +1,24 @@
 module Enzyme
 
+import EnzymeCore: Forward, Reverse
 export Forward, Reverse
+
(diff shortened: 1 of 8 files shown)


src/Enzyme.jl
```diff
@@ -1,136 +1,24 @@
...
 export Const, Active, Duplicated, DuplicatedNoNeed, BatchDuplicated, BatchDuplicatedNoNeed
+
+import EnzymeCore: batch_size
+export batch_size
+
 export autodiff, jacobian, gradient, gradient!
 export markType, batch_size, onehot, chunkedonehot
 
 using LinearAlgebra
+import EnzymeCore: ReverseMode, ForwardMode
```
vchuravy, line 16:
```suggestion
import EnzymeCore: ReverseMode, ForwardMode, Annotation 
```

Issue #488, opened 2022-10-05
Segmentation fault


Finn-Sueberkrueb:
Thank you very much for this wonderful package!

The following minimal example.
```julia
using Enzyme;

function objective!(x, loss, R)
    for i in 1:1000
        #y = R * x
        y = zeros(3)
        y[1] = R[1,1] * x[1] + R[1,2] * x[2] + R[1,3] * x[3]

        loss[] = y[1]
    end
    return nothing
end;

x = zeros(3)
R = [1.0 0.0 0.0
    0.0 1.0 0.0
    0.0 0.0 1.0]
loss = Ref(0.0)
dloss = Ref(1.0)

for i in 1:1000
    loss[] = 0.0
    dloss[] = 1.0
    autodiff(Reverse, objective!, Duplicated(x, zero(x)), Duplicated(loss, dloss), R)
end
```

Raises the following segmentation fault. The error occurs after varying number of AutoDiff calls.
```
signal (11): Segmentation fault: 11
in expression starting at /Users/src/minimal_example.jl:26
gc_mark_loop at /Applications/Julia-1.8.app/Contents/Resources/julia/lib/julia/libjulia-internal.1.8.dylib (unknown line)
_jl_gc_collect at /Applications/Julia-1.8.app/Contents/Resources/julia/lib/julia/libjulia-internal.1.8.dylib (unknown line)
ijl_gc_collect at /Applications/Julia-1.8.app/Contents/Resources/julia/lib/julia/libjulia-internal.1.8.dylib (unknown line)
jl_gc_pool_alloc_noinline at /Applications/Julia-1.8.app/Contents/Resources/julia/lib/julia/libjulia-internal.1.8.dylib (unknown line)
jl_gc_alloc at /Applications/Julia-1.8.app/Contents/Resources/julia/lib/julia/libjulia-internal.1.8.dylib (unknown line)
_new_array_ at /Applications/Julia-1.8.app/Contents/Resources/julia/lib/julia/libjulia-internal.1.8.dylib (unknown line)
_new_array at /Applications/Julia-1.8.app/Contents/Resources/julia/lib/julia/libjulia-internal.1.8.dylib (unknown line)
ijl_alloc_array_1d at /Applications/Julia-1.8.app/Contents/Resources/julia/lib/julia/libjulia-internal.1.8.dylib (unknown line)
Allocations: 56871453 (Pool: 56818646; Big: 52807); GC: 61
```

Version Info
```
[...]

PR #501, opened 2022-10-10
WIP segfault err

wsmoses:


wsmoses:
See https://github.com/EnzymeAD/Enzyme.jl/issues/488

PR #507, opened 2022-10-11
Refactored AD in box model

michel2323:
Refactored AD in box model so it doesn't require any special AD implementation and Enzyme takes care of all the checkpoints. This example also works with [Checkpointing.jl](https://github.com/Argonne-National-Laboratory/Checkpointing.jl), however, it currently still requires Zygote.jl due to ChainRules.jl. We can either add Checkpointing.jl now with Zygote.jl in this example or wait for EnzymeRules.jl. It's your call.

@swilliamson7 Can you check whether the example still makes sense to you? I checked whether it gives the same numerical results, but I don't know whether I didn't screw up the description.

### DIFF
--- a/examples/box.jl
+++ b/examples/box.jl
@@ -1,17 +1,17 @@
 # # Enzyme for adjoint tutorial: Stommel three-box ocean model
 
-# The goal of this tutorial is to teach about a specific usage of Enzyme's automatic 
-# differentiation capabilities, and will be centered around the Stommel ocean model. This 
-# is a nice example to see how powerful Enzyme is, and the ability of it to take a 
+# The goal of this tutorial is to teach about a specific usage of Enzyme's automatic
+# differentiation capabilities, and will be centered around the Stommel ocean model. This
+# is a nice example to see how powerful Enzyme is, and the ability of it to take a
 # derivative of a complicated function (namely one that has many parts and parameters).
-# This tutorial will focus first on the computations and getting Enzyme running, 
-# for those interested a mathematical explanation of the model and what an adjoint 
-# variable is will be provided at the end. 
+# This tutorial will focus first on the computations and getting Enzyme running,
+# for those interested a mathematical explanation of the model and what an adjoint
+# variable is will be provided at the end.
 
-# # Brief model overview 
+# # Brief model overview
 
 # The Stommel box model can be viewed as a watered down full ocean model. In our example, we have three
-# boxes (Box One, Box Two, and Box Three) and we model the transport of fluid between 
+# boxes (Box One, Box Two, and Box Three) and we model the transport of fluid between
 # them. The full equations of our system are given by:
 #
 # ```math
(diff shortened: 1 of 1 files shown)


swilliamson7:
It seems fine from what I can tell, I like that you simplified the functions. Why are you wanting to remove the whole example on using Enzyme to just calculate a derivative though? I understand that the full sensitivity calculation is more interesting but I think it's useful to show some simpler calculations with Enzyme, especially considering this is intended to be a tutorial. If it's just a matter of rewriting this part with the new functions I'm happy to do this

sriharikrishna:
From a pedagogical standpoint, I am uneasy with this example. because
`function forward_func(state, fld_old, fld_now, dt, M)` 
is differentiated using:
`autodiff(
    forward_func,
    Duplicated(state_out, dout_old),
    Duplicated([Tbar; Sbar], din_now),
    Duplicated([Tbar; Sbar], din_old),
    10*day,
    M,
)`
I got completely lost because of the order of the formal arguments `fld_old, fld_now,` and the actual arguments `din_now, din_old` is counterintuitive. I had a lot of trouble getting the call right within Checkpointing.jl.

swilliamson7:
You're right, it is counterintuitive. I started using the "old, now, new" notation because this was how the Fortran code I was modelling mine on did it, but really it doesn't make a whole lot of sense here. I'm guessing you already figured this out but the outputs of the forward function shouldn't really be considered as at different time steps, rather one is the value before the smoother, and one is after the smoother has been applied. The fact that the adjoint variables further confused this is my fault. Unless I'm misunderstanding what you're referring to?

Maybe it would be worth it for me to go through and adjust this notation to make it less confusing? I still think that having a plain derivative example is useful, especially to see what happens with the shadow outputs (if I'm using that term correctly, I think this is what you call the placeholders in the autodiff call) when you use Enzyme

michel2323:
I thought users who want to get a gradient using Enzyme.jl might be overwhelmed by the notion of shadow copies etc. In one example, there maybe should only be one forward function, then Enzyme is applied, and it all works magically. However, I could readd the code for a single step. My main goal was to get rid of AD-specific functions to get rid of the impression that the code requires some special massaging.

In any case, feel free to push a more consistent variable naming. I see that I also added some more confusion (e.g., `Duplicated(state_out, dout_old)`).

swilliamson7:
Gotcha, correct me if I’m wrong but won’t people need to work with Duplicated if they want the gradient of a function that has vector output? Meaning learning the whole shadow stuff despite having a simple function. 

And yeah I’ll try and fix notation!

swilliamson7:
Sorry for all the comments, but I just sat down and really looked at the updated functions. I think that we're on two different pages about what the tutorial was supposed to do: it seems like it was edited to be efficient for this one specific example, but it's not really using the adjoint method anymore. I don't really see a way to store all the adjoint variables and use them for other purposes, and the goal was mainly to teach about using Enzyme for the adjoint method. The function that I wrote and called ad_calc was not intended to be something special for Enzyme, rather it was calculating the adjoint variables one by one and giving the potential to store any/all of them.

vchuravy:
There also speaks nothing against having two examples each focusing on a different pedagogical example.

sriharikrishna:
Changing the names would be helpful for sure. Thanks @swilliamson7. I agree that two separate implementations for two different purposes based on the same underlying physical model makes sense.

swilliamson7:
Cool, sounds good!

michel2323:
No worries. This is how we butchered it to use it in `Checkpointing.jl`. @vchuravy I think I'll bring the `heat.jl` example over from `Checkpointing.jl` for some variety. This can be closed then I guess.

vchuravy:
@michel2323 what do you want to do with this branch?

michel2323:
@swilliamson7 @vchuravy I would add `Checkpointing.jl` and make it an example of using Enzyme on a time-dependent model. @swilliamson7 I see your comment here. Would that change be okay with you? Since this is on the Enzyme page, it would be helpful to focus on Enzyme's API and less on the model itself.

swilliamson7:
@michel2323 I don't have any real opinion, it's been a super long time since I looked at this code...feel free to use it however you guys think is best!

PR #508, opened 2022-10-11
Throw a runtime error if tape size mismatch is detected

vchuravy:


vchuravy:
```
julia> sizeof(A{N, NT})
65568

julia> N = 684
684

julia> sizeof(A{N, NT})
8
```

So when we go over ~2^16 bytes we hit an internal threshold and we no longer allocate the NTuple as a chunk.

Issue #500, opened 2022-10-10
Segmentation fault on a linear regression model (arm & x86 chips)


freddycct:
There are mainly two issues here.
1. Segmentation fault 
2. autodiff messes up the mapreduce functions. This is less of an issue because we can avoid using mapreduce.

This is the MWE
```
using Lux
using Enzyme
using Random
using Optimisers

function lossMapReduce(model, X, Y, ps, st)
	# this doesn't work
	mapreduce(+, zip(X, Y)) do (x,y)
		yhat = Lux.apply(model, x, ps, st)[1]
		(yhat[1] - y)^2
	end
end

function loss(model, X, Y, ps, st)
	# this works but causes segmentation fault after running for a few times
	ll = 0.0f0
	for (x, y) in zip(X, Y)
		yhat = Lux.apply(model, x, ps, st)[1]
		ll += (yhat[1] - y)^2
	end
	return ll
end

function generateToyData(rng, K, N)
	W = randn(rng, Float32, K) # this is the "real" parameters
	b = randn(rng, Float32)    # bias
	
	X = map(x -> rand(rng, Float32, K), 1:N) # features
	Y = map(x -> W' * x + b, X) # ground truth
	return X, Y, W, b
end

function main()
	rng = Random.default_rng()
	Random.seed!(rng, 0)
	
	# lux specific codes
	model = Dense(16, 1)
	ps, st = Lux.setup(rng, model) # ps is the parameters, st is the state

	# generate some toy data
	X, Y, W, b = generateToyData(rng, 16, 1000)

	# setup optimisers
	optRule = Optimisers.Adam()
	optState = Optimisers.setup(optRule, ps)  # optimiser state based on model parameters

	# println("ps = ", ps)
	totalLoss = loss(model, X, Y, ps, st)
	println("0/100: loss = $(totalLoss)")

	totalEpochs1 = 200
	totalEpochs2 = totalEpochs1 + 10

	# this causes a segmentation fault after some 50+ epochs
	for epoch=1:totalEpochs1
		# zero the cache
		grads = Lux.fmap(zero, ps)

		# calculate gradients
		autodiff(Reverse, loss, Active, Const(model), Const(X), Const(Y), Duplicated(ps, grads), Const(st))

		# gradient update using adam optimizer
		optState, ps = Optimisers.update!(optState, ps, grads)

		totalLoss = loss(model, X, Y, ps, st)
[...]

PR #509, opened 2022-10-12
Fix large dynamic tape issue

wsmoses:
Fixes https://github.com/EnzymeAD/Enzyme.jl/issues/500

### DIFF
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -3414,10 +3414,6 @@ nfields(Type::LLVM.VectorType) = size(Type)
 nfields(Type::LLVM.ArrayType) = length(Type)
 nfields(Type::LLVM.PointerType) = 1
 
-mutable struct EnzymeTape{N, T}
-    data::NTuple{N, T}
-end
-
 mutable struct EnzymeTapeToLoad{T}
     data::T
 end
@@ -3617,7 +3613,7 @@ function julia_allocator(B, LLVMType, Count, AlignedSize, IsDefault, ZI)
         if Count isa LLVM.ConstantInt
             N = convert(Int, Count)
 
-            ETT = N == 1 ? EnzymeTapeToLoad{TT} : EnzymeTape{N, TT}
+            ETT = N == 1 ? EnzymeTapeToLoad{TT} : NTuple{N, TT}
             if sizeof(ETT) !=  N*convert(Int, AlignedSize)
                 @safe_error "Size of Enzyme tape is incorrect. Please report this issue" ETT sizeof(ETT) TargetSize = N*convert(Int, AlignedSize)
                 emit_error(B, nothing, "Enzyme: Tape allocation failed.") # TODO: Pick appropriate orig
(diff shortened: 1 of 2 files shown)


vchuravy:
That is illegal. `NTuple` is an immutable so despite it being allocated on the heap we shouldn't mutate it after it has been observed

wsmoses:
We're not mutating it after being observed, we're initializing it?

vchuravy:
We are guaranteed to never update values in place? That was my assumption for why it had to be a mutable

wsmoses:
thinking this through:

This technically overrides the Enzyme internal allocator in its entirety. 

We would only have a julia object internally if we needed to cache it for a tape. The tape should be immutable once initialized (technically we may memset it to 0 -- in fact in this very julia_allocator function), but otherwise once its value is set to actually preserve something, it is the same. We even attach invariant.group MD on it

src/compiler.jl
```diff
@@ -3617,7 +3613,7 @@ function julia_allocator(B, LLVMType, Count, AlignedSize, IsDefault, ZI)
         if Count isa LLVM.ConstantInt
             N = convert(Int, Count)
 
-            ETT = N == 1 ? EnzymeTapeToLoad{TT} : EnzymeTape{N, TT}
+            ETT = N == 1 ? EnzymeTapeToLoad{TT} : NTuple{N, TT}
```
vchuravy, line 3616:
```suggestion
            ETT = N == 1 ? TT : NTuple{N, TT}
```

src/compiler.jl
```diff
@@ -3414,10 +3414,6 @@ nfields(Type::LLVM.VectorType) = size(Type)
 nfields(Type::LLVM.ArrayType) = length(Type)
 nfields(Type::LLVM.PointerType) = 1
 
-mutable struct EnzymeTape{N, T}
-    data::NTuple{N, T}
-end
-
 mutable struct EnzymeTapeToLoad{T}
```
vchuravy, line 3417: Also delete this then?

vchuravy (approved):

src/compiler.jl
wsmoses: No EnzymeTapeToLoad is distinct and used in split mode contexts by Enzyme internally to say if it is recursive and we need to load the tape for the reverse pass before the call

test/runtests.jl
```diff
@@ -1559,3 +1559,28 @@ end
...
+	ps = randn(Float32, (1, 5))
+	bs = randn(Float32)
+
+	X = map(x->rand(Float32, 5), 1:1000)
+	Y = map(x->rand(Float32), 1:1000)
+
+	grads = zero(ps)
+	for epoch=1:1000
+		fill!(grads, 0)
+		autodiff(Reverse, ldtnloss, Const(X), Const(Y), Duplicated(ps, grads), Active(bs))
```
vchuravy, line 1583:
```suggestion
		autodiff(Reverse, ldynloss, Const(X), Const(Y), Duplicated(ps, grads), Active(bs))
```

PR #516, opened 2022-10-12
Add box float

wsmoses:


src/compiler.jl
```diff
@@ -246,12 +243,16 @@ function emit_allocobj!(B, T, size)
...
+	ty = LLVM.const_addrspacecast(ty, T_prjlvalue)
+
+    pgcstack = reinsert_gcmarker!(fn, B)
+    ct = inbounds_gep!(B, bitcast!(B, pgcstack, T_ppjlvalue), [LLVM.ConstantInt(current_task_offset(); ctx)])
+
 	size = LLVM.ConstantInt(T_size, size)
-    args = [reinsert_gcmarker!(fn), size, ty]
-	args[1] = bitcast!(B, args[1], parameters(eltype(llvmtype(func)))[1])
-    return call!(B, func, args)
+    return call!(B, func, [ct, size, ty])
```
vchuravy, line 255: You probably need to branch on versions here since some take the ct and some the ptls. See vc/rules2 for a full implementation. Also see the corrected constants there for `current_task_offset`

wsmoses: This is taken from vc/rules2 (or was attempted to).

vchuravy: Sorry all the PRs blend together.

What you want is https://github.com/EnzymeAD/Enzyme.jl/blob/410a7cc75a0d5fa04e6f2e75a931020e0434f1f9/src/compiler.jl#L3702-L3710
and https://github.com/EnzymeAD/Enzyme.jl/blob/410a7cc75a0d5fa04e6f2e75a931020e0434f1f9/src/compiler.jl#L3732-L3736 + https://github.com/EnzymeAD/Enzyme.jl/blob/410a7cc75a0d5fa04e6f2e75a931020e0434f1f9/src/compiler.jl#L3718-L3729

and if you really want to make me happy you merge that into `emit_allocobj` so that there is one correct implementation.

src/compiler.jl
```diff
@@ -3781,48 +3825,8 @@ function julia_allocator(B, LLVMType, Count, AlignedSize, IsDefault, ZI)
...
-            if VERSION < v"1.8.0"
-                obj = call!(B, alloc_obj, [ptls, Size, tag])
-            else
-                obj = call!(B, alloc_obj, [ct, Size, tag])
-            end
-        else
-            obj = call!(B, alloc_obj, [ptls, Size, tag])
-        end
+        
+        obj = emit_allocobj!(B, tag, Size, needs_dynamic_size_workaround)
```
vchuravy, line 3829: Can we sink the needs_dynamic_workaround check?

wsmoses: The first workaround (eg if constant or not), yes. The second workaround (whether there are gep users of a non-constant offset), no.

vchuravy (approved):

PR #532, opened 2022-10-24
Fastgen2

wsmoses:
Reduces from 4ms down to 650us. Still far from the 5.5us possible.

Currently only working forward diff, can also get augmented fwd. Reverse requires some thought.


vchuravy:
Is this still needed with fastgen3?

PR #535, opened 2022-10-25
Add immediate ABI

wsmoses:


vchuravy (commented):
I don't think we should do this. I rather add support in Cthulhu to understand  the deferred reflection

PR #537, opened 2022-10-27
Fastgen3

wsmoses:


Project.toml
```diff
@@ -5,6 +5,7 @@ version = "0.10.12"
 
 [deps]
 CEnum = "fa961155-64e5-5f13-b03f-caf6b980ea82"
+Cthulhu = "f68482b8-f384-11e8-15f7-abe071a5a75f"
```
vchuravy, line 8:
```suggestion
```

src/compiler.jl
```diff
@@ -714,66 +1054,287 @@ end
...
-                a[] += dd
+        end
+    end
+    fn = eval(efn)
+    FwdCache[(N,Width)] = fn
+    return fn
+
+end
+
+const AugCache = Dict{Tuple{Int64,Int64},Function}()
```
vchuravy, line 1181: Let's add tests that show that this is correct under invalidation of functions. Not that we need to bifurcate this cache on a world-age range.

src/Enzyme.jl
```diff
@@ -171,11 +171,11 @@ b = [2.2, 3.3]; ∂f_∂b = zero(b)
 c = 55; d = 9
 
 f(a, b, c, d) = a * √(b[1]^2 + b[2]^2) + c^2 * d^2
-∂f_∂a, ∂f_∂d = autodiff(Reverse, f, Active, Active(a), Duplicated(b, ∂f_∂b), c, Active(d))
+∂f_∂a, _, _, ∂f_∂d = autodiff(Reverse, f, Active, Active(a), Duplicated(b, ∂f_∂b), c, Active(d))[1]
```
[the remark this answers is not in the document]
wsmoses: This is to extend the capability that the tuple can contain not just the differentiated args, but also the primal return, roughly analagous to what you describe.

src/compiler.jl
wsmoses: So this cache is not type dependent, but arg/width, dependent. It contains the generic templated function for a given arg/width combo.

The bigger concern is that if you look at the impl it requires an eval (which ideally would be generated on the fly, but per an eval in a generated function being illegal, I just run a for loop over "common" sizes). We should also add a fallback to the current fastgen2 style vararg code if its outside that, unless we have a way to get around the eval in generated.

vchuravy: Can we use an OpaqueClosure instead? Or just an anonymous function?

wsmoses: sure, re either

src/compiler.jl
```diff
@@ -809,93 +1117,283 @@ function runtime_invoke_fwd(arg_ptr, shadow_ptr, activity_ptr::Val{ActivityTup},
...
+    if haskey(AugCache, (N,Width))
+        return AugCache[(N,Width)]
+    end
+    if !create
+        return runtime_generic_augfwd_fallback
     end
+    primargs, shadowargs, primtypes, allargs, typeargs, wrapped = setup_macro_wraps(false, N, Width)
+    funcname = Symbol("inner_runtime_generic_augfwd_"*string(N)*"_"*string(Width))
+    fn = eval(quote
+        function $funcname(activity::Val{ActivityTup}, width::Val{$Width}, RT::Val{ReturnType}, f::F, df::DF,$(allargs...)) where {ActivityTup,ReturnType, F, DF, $(typeargs...)}
```
vchuravy, line 1291: Why do we need to specialize on Width and allargs?

Can we not specialize on first call instead of going through this whole custom cache nonsense?

wsmoses: Having varargs prevents the correct type from being deduced, hence the splatted type. The caching is necessary because otherwise eval'ing a function isn't legal from a global scope. I would vastly prefer getting rid of the cache and specializing on the first call if you have ideas.

src/compiler.jl
```diff
@@ -325,6 +325,292 @@ declare_writebarrier!(mod) = get_function!(mod, "julia.write_barrier") do ctx
     T_prjlvalue = LLVM.PointerType(T_jlvalue, 10) 
     LLVM.FunctionType(LLVM.VoidType(ctx), [T_prjlvalue]; vararg=true)
 end
+@static if VERSION < v"1.8.0"
+declare_apply_generic!(mod) = get_function!(mod, "jl_apply_generic") do ctx
```
vchuravy, line 329: These functions are not used anywhere

src/compiler.jl
```diff
@@ -604,45 +912,50 @@ end
     end
 end
 
-@inline function common_interface_augfwd(annotation, forward, adjoint, args, width::Val{Width}, RT::Val{ReturnType}) where {Width,ReturnType} 
+struct Tape{TapeTy,ShadowTy,ResT}
+    internal_tape::TapeTy
+    shadow_return::ShadowTy
+end
+
+@inline function common_interface_augfwd(annotation, forward::ForwardTy, args::ArgsTy, width::Val{Width}, RT::Val{ReturnType}) where {ForwardTy,ArgsTy,Width,ReturnType}
```
vchuravy, line 920: `RT::Type{ReturnType}` no need to put it in a Val.

vchuravy (approved):

PR #546, opened 2022-11-14
AutoDiff API doc

michel2323:
I've been playing with the Enzyme API for vector forward over reverse and thought I'd write some documentation.

### DIFF
--- a/examples/autodiff.jl
+++ b/examples/autodiff.jl
@@ -1,0 +1,146 @@
+# # AutoDiff API
+
+# The goal of this tutorial is to give users already familiar with automatic
+# differentiation (AD) an overview
+# of the Enzyme differentiation API for the following differentiation modes
+# * Reverse mode
+# * Forward mode
+# * Forward over reverse mode
+# * Vector Forward over reverse mode
+# # Defining a function
+# Enzyme differentiates arbitrary multivariate vector functions as the most
+# general case in automatic differentiation
+# ```math
+# f: \mathbb{R}^n \rightarrow \mathbb{R}^m, y = f(x)
+# ```
+# For simplicity we define a vector function with ``m=1``. However, this
+# tutorial can easily be applied to arbitrary ``m \in \mathbb{N}``.
+using Enzyme
+
+function f(x::Array{Float64}, y::Array{Float64})
+    y[1] = x[1] * x[1] + x[2] * x[1]
+    return nothing
+end;
+
+# # Reverse mode
+# The reverse model in AD is defined as
+# ```math
+# \begin{aligned}
+# y &= f(x) \\
+# \bar{x} &= \bar{y} \cdot \nabla f(x)
+# \end{aligned}
+# ```
+# bar denotes an adjoint variable. Note that executing an AD in reverse mode
+# computes both ``y`` and the adjoint ``\bar{x}``.
+x  = [2.0, 2.0]
+bx = [0.0, 0.0]
+y  = [0.0]
+by = [1.0];
+
+# Enzyme stores the value and adjoint of a variable in an object of type
+# `Duplicated` where the first element represent the value and the second the
+# adjoint. Evaluating the reverse model using Enzyme is done via the following
+# call.
+Enzyme.autodiff(f, Duplicated(x, bx), Duplicated(y, by));
+# This yields the gradient of `f` in `bx` at point `x = [2.0, 2.0]`. `by` is called the seed and has
+# to be set to ``1.0`` in order to compute the gradient. Let's save the gradient for later.
+g = copy(bx)
+
+# # Forward mode
+# The forward model in AD is defined as
+# ```math
+# \begin{aligned}
+# y &= f(x) \\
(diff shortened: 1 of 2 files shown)


examples/autodiff.jl
```diff
@@ -0,0 +1,146 @@
...
+    Forward,
+    (x,y) -> Enzyme.autodiff_deferred(f, x, y),
+    BatchDuplicated(Duplicated(x, bx), Duplicated.(vdx, vdbx)),
+    BatchDuplicated(Duplicated(y, by), Duplicated.(vdy, vdby)),
+);
+
+# Again we obtain the first-order gradient.
+g[1] == dy[1][1]
+# We have now the first row/column of the Hessian
+vdbx[1][1] == 2.0
```
wsmoses, line 140: Make these approx rather than equals?

michel2323: Done.

examples/autodiff.jl
```diff
@@ -0,0 +1,146 @@
...
+# ```math
+# \begin{aligned}
+# y &= f(x) \\
+# \dot{y} &= f(x) \cdot \dot{x} \\
+# \bar{x} &= \bar{y} \cdot \nabla f(x) \\
+# \dot{\bar{x}} &= \bar{y} \cdot \nabla^2 f(x) \cdot \dot{x} + \dot{\bar{y}} \cdot \nabla f(x)
+# \end{aligned}
+# ```
+# To obtain the first column/row of the Hessian ``\nabla^2 f(x)`` we have to
+# seed ``dot{\bar{y}}`` with ``[0.0]``, ``\bar{y}`` with ``[1.0]`` and ``\dot{x}`` with ``[1.0, 0.0]``.
```
frapac, line 86: A `\` is missing: 
```
``\dot{\bar{y}}``
```

examples/autodiff.jl
```diff
@@ -0,0 +1,146 @@
...
+# In the forward mode the second element of `Duplicated` stores the tangent.
+Enzyme.autodiff(Forward, f, Duplicated(x, dx), Duplicated(y, dy));
+
+# We can now verify that indeed the reverse mode and forward mode yield the same
+# result for the first component of the gradient. Note that to acquire the full
+# gradient one needs to execute the forward model a second time with the seed
+# `dx` set to `[0.0,1.0]`.
+
+# Let's verify whether the reverse and forward model agree.
+g[1] ≈ dy[1]
```
frapac, line 72: Can we introduce a short paragraph to explain how to use vectorized forward mode here? This could be useful to explain how to evaluate sparse Jacobians, and it would be a nice increment to explain the vectorized forward over reverse later.

examples/autodiff.jl
```diff
@@ -0,0 +1,146 @@
...
+
+vdy = ([0.0],[0.0])
+vdx = ([1.0, 0.0], [0.0, 1.0])
+
+bx = [0.0, 0.0]
+by = [1.0]
+vdbx = ([0.0, 0.0], [0.0, 0.0])
+vdby = ([0.0], [0.0]);
+
+# The `BatchedDuplicated`` objects are constructed using the broadcast operator
```
frapac, line 128: The rendering is broken because of one additional tick:
```
`BatchedDuplicated`
```

examples/autodiff.jl
```diff
@@ -0,0 +1,146 @@
...
+# on our tuples of `Duplicated` for the tangents.
+Enzyme.autodiff(
+    Forward,
+    (x,y) -> Enzyme.autodiff_deferred(f, x, y),
+    BatchDuplicated(Duplicated(x, bx), Duplicated.(vdx, vdbx)),
+    BatchDuplicated(Duplicated(y, by), Duplicated.(vdy, vdby)),
+);
+
+# Again we obtain the first-order gradient.
+g[1] ≈ dy[1][1]
```
frapac, line 138: Shouldn't we use `vdy` instead?

frapac (commented):
This tutorial is very nice, especially for people like me unfamiliar with Enzyme's API.

examples/autodiff.jl
michel2323: 👍I will add a sparse Jacobian computation using vector forward and using the Symbolics.jl sparsity detection and SparseDiffTools.jl coloring.

examples/autodiff.jl
michel2323: Done

examples/autodiff.jl
michel2323: Thanks for catching!

examples/autodiff.jl
michel2323: Of course. That was a test of whether someone reads it 🙂. Thanks!

examples/autodiff.jl
michel2323: I'm working on a sparse example. Let's merge this first and do the batched forward in a separate one. I think it may make this one too heavy.

vchuravy (approved):

Issue #540, opened 2022-10-29
Enzyme.jl hits an unknown intrisic error when diffing through kepler_solver


sefffal:
Thanks for the very promising package!

Enzyme fails with an unknown intrinsic when differentiating through a small numerical function, `AstroLib.kepler_solver`.

```
cannot handle (augmented) unknown intrinsic
  %55 = call i128 @llvm.abs.i128(i128 %51, i1 noundef false) #18, !dbg !141
```

Reproducer:
```julia
using AstroLib
kep(a)= AstroLib.kepler_solver(a[1], a[2])
i = [1.0, 0.1]
o = similar(i)
Enzyme.autodiff(kep, Duplicated(i,o))
```

Link to the function source code: https://github.com/JuliaAstro/AstroLib.jl/blob/826e5d81062e007bd88a14f901836de82ea576d7/src/kepler_solver.jl#L4

sefffal:
Looks like error is caused by `rem2pi`. The following gives the same error:
```
Enzyme.autodiff(x->rem2pi(x,RoundDown), Active(0.5))
```

Curiously, `rem` causes a different error:
```julia
Enzyme.autodiff(x->rem(x,1), Active(0.5))
```
`cannot handle unknown binary operator:   %2 = frem double %0, 1.000000e+00, !dbg !7`

tgymnich:
Interestingly this, seems to work for me if I build Enzyme (0.0.45) myself.

wsmoses:
This should now be fixed on main. Retry and confirm?

PR #569, opened 2022-12-24
Add test for frem

wsmoses:
Add test for https://github.com/EnzymeAD/Enzyme.jl/issues/540


wsmoses:
This is landable upon a jll bump.

PR #598, opened 2023-02-03
Add primal return option

wsmoses:


vchuravy:
Great! Let's add some tests?

vchuravy (approved):

lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -113,9 +113,10 @@ abstract type Mode end
 
 Reverse mode differentiation
 """
-struct ReverseMode <: Mode
+struct ReverseMode{ReturnPrimal} <: Mode
 end
-const Reverse = ReverseMode()
+const Reverse = ReverseMode{false}()
```
vchuravy, line 118: Actually what's the reason for defaulting for false?

In combined forward and reverse mode I think the user will almost always want the forward value.

wsmoses: For semi ABI consistency, and I also think the "WithPrimal" seems more intuitive, since it implies the extra work that must be done.

PR #614, opened 2023-02-07
Move and simplify activity rules

wsmoses:


vchuravy (approved):

Project.toml
```diff
@@ -17,7 +17,7 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 
 [compat]
 CEnum = "0.4"
-EnzymeCore = "0.2.1"
+EnzymeCore = "0.2.2"
```
vchuravy, line 20: We haven't registered `0.2.1` so we don't need a bump, but I can register for you.

PR #637, opened 2023-02-23
Implement congruency check in EnzymeCore

### LINKED ISSUE #636: Implement Congruent
We should check for structural identity/congruence on construction of a DuplicatedArray

vchuravy:
Fixes #636

The goal is to check on construction of Duplicated that the shadows are structurally equivalent
and thus the reverse pass does not "index past the array".


lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -101,6 +105,31 @@ batch_size(::BatchDuplicated{T,N}) where {T,N} = N
...
 
+"""
+    congruent(a::T, b::T)::T
+
+Defines values to be congruent, e.g. structurally equivalent.
+"""
+function congruent end
+
+congruent(a::T, b::T) where T<:Number = true
+congruent(a::T, b::T) where T<:AbstractArray = length(a) == length(b)
```
wsmoses, line 116: This is not sufficient, say for an offset array they should have same offset likely

vchuravy: I think `axes(a) == axes(b)` should suffice.

lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -1,11 +1,33 @@
...
+"""
+function structure_check!(flag)
+    @set_preferences!("structure_check" => flag)
+    @info("structure_check toggled, restart your Julia session for this change to take effect!")
+
+    if VERSION <= v"1.6.5" || VERSION == v"1.7.0"
+        @warn """
+        Due to a bug in Julia (until 1.6.5 and 1.7.1), setting preferences in transitive dependencies
+        is broken (https://github.com/JuliaPackaging/Preferences.jl/issues/24). To fix this either update
+        your version of Julia, or add EnzyemCore as a direct dependency to your project.
```
wsmoses, line 25: Rename EnzymeCore ? (typo)

vchuravy, line 25:
```suggestion
        your version of Julia, or add EnzymeCore as a direct dependency to your project.
```

lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -61,6 +83,10 @@ accumulator for gradients (so ``\\partial f / \\partial x`` will be *added to*)
 struct Duplicated{T} <: Annotation{T}
     val::T
     dval::T
+    function Duplicated(val::T, dval::T; checked=structure_check) where T
```
wsmoses, line 86: if the preference isn't there, will this be a static no check, or will it load a global then check?

If the latter, I want to have a way to avoid the performance penalty.

vchuravy: It's a global const so it's propagated (and it's a compile time preference)

PR #641, opened 2023-03-03
Add rosenbrock example

vchuravy:


sriharikrishna:
The function should be called `rosenbrock` not `rosenbrook`.

sriharikrishna (changes requested):
Change the function name.

vchuravy:
https://enzymead.github.io/Enzyme.jl/previews/PR641/ as a preview

docs/src/index.md
```diff
@@ -18,6 +18,84 @@ The Enzyme binary dependencies will be installed automatically via Julia's binar
...
+((nothing,),)
+
+julia> dx
+2-element Vector{Float64}:
+ -400.0
+  200.0
+```
+
+Both the inplace and "normal" variant return the gradient. The difference is that with
+[`Active`](@ref) the gradient is returned and with [`Duplicated`](@ref) the gradient is returned in the shadow-heap.
```
wsmoses, line 59: rather than saying is returned in the shadow heap, perhaps "accumulated" in place. I think it's important for the user to know it's +='d rather than set.

docs/src/index.md
```diff
@@ -18,6 +18,84 @@ The Enzyme binary dependencies will be installed automatically via Julia's binar
...
+
+### Forward mode
+With forward mode we can seed the arguments. In forward mode `Duplicated(x, 0.0)` is equivalent to `Const(x)`,
+except that we can perform more optimizations for `Const`. We obtain as a result the primal return of `rosenbrock`
+and the tangent return.
+
+```julia-repl
```
wsmoses, line 66: Since ambiguous here without further context, can we say which is the primal and which is the derivative?

docs/src/index.md
```diff
@@ -18,6 +18,84 @@ The Enzyme binary dependencies will be installed automatically via Julia's binar
 
 The Enzyme.jl API revolves around the function [`autodiff`](@ref), see it's documentation for details and a usage example. Also see [Implementing pullbacks](@ref) on how to use Enzyme.jl to implement back-propagation for functions with non-scalar results.
 
+## Getting started
+
+```julia
+rosenbrock(x, y) = (1.0 - x)^2 + 100.0 * (y - x^2)^2
+rosenbrock_inp(x) = 1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2)^2
```
sriharikrishna, line 25: missing `(`. Should be `rosenbrock_inp(x) = (1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2)^2`

sriharikrishna (changes requested):

docs/src/index.md
vchuravy: Good catch

docs/src/index.md
```diff
@@ -18,6 +21,111 @@ The Enzyme binary dependencies will be installed automatically via Julia's binar
 
 The Enzyme.jl API revolves around the function [`autodiff`](@ref), see it's documentation for details and a usage example. Also see [Implementing pullbacks](@ref) on how to use Enzyme.jl to implement back-propagation for functions with non-scalar results.
 
+## Getting started
+
+```jldoctest rosenbrock ; output = false
```
vchuravy, line 26:
```suggestion
```jldoctest rosenbrock
```

sriharikrishna (approved):

PR #651, opened 2023-03-05
Add deferred gradient/jacobian helpers

wsmoses:


vchuravy:
I would rather see if we could turn `autodiff_deferred` on by default?

vchuravy (changes requested):
I don't think we should expose these.

Issue #695, opened 2023-04-05
[EnzymeRules]: Original function implementation determines types of `augemented_primal()` and `reverse()`


michel2323:
**Use Case**
I want to differentiate a function `g(x::Vector{Float64})::nothing` and provide custom rules using EnzymeRules.
**Expectation**
The signatures of the `augmented_primal()` and `reverse()` methods will only depend on `g`'s signature and not on the implementation. 
**Example**
In the code below, the signature of augmented_primal and `reverse()` are different for `g1` and `g2`, although they have the same signature.
```julia
using Enzyme
using Enzyme: EnzymeRules
import .EnzymeRules: inactive, augmented_primal, reverse, Annotation, has_rrule_from_sig
using .EnzymeRules

function g1(x)
    # x .= sum(x .^ 2)
    return nothing
end

function g2(x)
    x .= sum(x .^ 2)
    return nothing
end

function augmented_primal(
    config,
    func::FT,
    ret,
    x
) where {FT}
    println("augemented_primal(g, x::$(typeof(x)))")
    return AugmentedReturn(nothing, nothing, nothing)
end

function reverse(
    config,
    func::Const{FT},
    dret,
    tape,
    x
) where {FT}
    println("reverse(g, x::$(typeof(x)))")
    return ()
end

x = [1.0, 2.0, 3.0]
dx = [0.0, 0.0, 0.0]
autodiff(Reverse, g1, Const, Duplicated(x, dx))
autodiff(Reverse, g2, Const, Duplicated(x, dx))
```

michel2323:
AFAIK, as a side effect of this, the custom rule implementation will fail if Enzyme fails with the activity analysis on the implementation of `g`'s implementation, although `g` is never differentiated by Enzyme.

wsmoses:
@michel2323 your overload of FT is very dangerous -- you've overridden all single arg functions, including sum, identity, etc...

wsmoses:
```
julia> autodiff(Reverse, g1, Const, Duplicated(x, dx))

augemented_primal(g, x::Const{Vector{Float64}})
reverse(g, x::Const{Vector{Float64}})
((nothing,),)

julia> autodiff(Reverse, g2, Const, Duplicated(x, dx))
augemented_primal(g, x::Duplicated{Vector{Float64}})
reverse(g, x::Duplicated{Vector{Float64}})
((nothing,),)
```

wsmoses:
[...]

PR #697, opened 2023-04-05
Custom rule doc example

### LINKED ISSUE #704: Missing docstrings for types used in custom reverse rules
`AugmentedReturn` and `Config` are missing docstrings. It would be nice to have them since the user must use these ypes when writing a custom reverse rule.

gaurav-arya:
Adds custom rule tutorial, and also resolves #704 

Would welcome any comments on my progress so far 🙂


examples/custom_rule.jl
```diff
@@ -0,0 +1,90 @@
...
+# we would like to use a custom rule for `f`. For example, `f` may make a foreign call that Enzyme cannot 
+# differentiate, or we may have special higher-level knowledge about `f` that enables us to write a more 
+# efficient rule. So let's see how to write a custom rule for `f`.
+
+# ## Defining our first rule 
+
+# First, we import the functions [`EnzymeRules.forward`](@ref) and [`EnzymeRules.reverse`](@ref).
+# We will need to overload these functions in order to define our custom rules.
+
+import Enzyme.EnzymeRules: forward, reverse
```
wsmoses, line 39: Also augmented

examples/custom_rule.jl
```diff
@@ -0,0 +1,90 @@
...
+# If the output has a [`DuplicatedNoNeed`](@ref) annotation, it means we are only interested in its derivative.
+# To squeeze the last drop of performance, the below rule avoids computing the output of the original function and 
+# just computes its derivative.
+
+function forward(func::Const{typeof(f)}, ::Type{<:DuplicatedNoNeed}, y::Duplicated, x::Duplicated)
+    y.val .= x.val.^2 
+    y.dval .+= 2 .* x.val .* x.dval
+    return sum(y.dval)
+end
+
```
gaurav-arya, line 89: if I keep going like this, I'm going to need a lot of rules 🙃 I don't know if there's an easy way of specifying these more compactly, say to at least get things correct for the case where `f` really needs a custom rule. (And then if the performance is desired maybe relying on some dead code elimination / thunking).

If that's not a thing yet, maybe for this tutorial I should just give a couple more rules for `f` to cover some representative cases and leave things there.

wsmoses: So what you can do is not force the annotation on the type or return. And then can do if duplicated (compute return), else return just the derivative.

gaurav-arya: (oops, was meant to include the TODO at the end in the code highlight here)

gaurav-arya: is `Active` a third (non-batched) possibility for `RT`?

wsmoses: Only if the return is a float or complex of float/etc.

examples/custom_rule.jl
```diff
@@ -0,0 +1,189 @@
...
+y  = [0.0, 0.0]
+dy = [0.0, 0.0]
+
+@show autodiff(Forward, f, Duplicated(y, dy), Duplicated(x, dx)) # derivative of f w.r.t. x[1]
+@show dy; # derivative of y w.r.t. x[1] when f is run
+
+# !!! note "Custom rule dispatch"
+#     When multiple custom rules for a function are defined, the correct rule is chosen chosen using 
+#     [Julia's multiple dispatch](https://docs.julialang.org/en/v1/manual/methods/#Methods).
+#     In particular, it is important to understand that the custom rule does not *determine* the
```
gaurav-arya, line 111: Is this correct? Was inspired by #695.

(And now that I think of it, if it's true, maybe there are some cases where we'd want custom activity annotations... e.g. if one writes `x, y = f(a, b)` where `f` is some function not understandable by Enzyme, and say the Jacobian of `f` is actually diagonal, it could be good to know that for activity analysis?)

lib/EnzymeCore/src/rules.jl
```diff
@@ -18,6 +18,17 @@ the annotated function arguments.
...
+    ConfigWidth{Width} = Config{<:Any,<:Any, Width}
+
+Configuration type to dispatch on in custom reverse rules (see [`augmented_primal`](@ref) and [`reverse`](@ref)).
+* `NeedsPrimal` and `NeedsShadow` are boolean values specifying whether the primal and shadow (resp.) should be returned. 
+* `Width` is an integer that specifies the number of adjoints/shadows simultaneously being propagated.
+* `Overwritten` is ??? (TODO) 
+
+Getters for the four type parameters are provided by `needs_primal`, `needs_shadow`, `width`, and `overwritten`.
+"""
 struct Config{NeedsPrimal, NeedsShadow, Width, Overwritten} end
```
gaurav-arya, line 32: Wonder if this is better named `ReverseConfig`

lib/EnzymeCore/src/rules.jl
```diff
@@ -26,6 +37,12 @@ const ConfigWidth{Width} = Config{<:Any,<:Any, Width}
 @inline width(::Config{<:Any, <:Any, Width}) where Width = Width
 @inline overwritten(::Config{<:Any, <:Any, <:Any, Overwritten}) where Overwritten = Overwritten
 
+"""
+    AugmentedReturn(primal, shadow, tape)
+
+Augment the primal return value of a function with its shadow, as well as any additional saved information
```
gaurav-arya, line 43: Don't really understand tape, so was vague

wsmoses: perhaps "any additional information needed to correctly compute the reverse pass. Unless specified by the config that a variable is not overwritten, rules must assume any arrays/data structures/etc are overwritten between the forward and the reverse pass. Any floats or variables passed by value are always preserved as is (as are the arrays themselves, just not necessarily the values in the array)."

examples/custom_rule.jl
wsmoses: Yes, it is correct. The one thing to add is "If a custom rule is specified for the correct function/argument types, but not the correct activity annotation, a runtime error will be thrown alerting the user to the missing activity rule rather than silently ignoring the rule."

lib/EnzymeCore/src/rules.jl
```diff
@@ -23,9 +23,9 @@ function forward end
     ConfigWidth{Width} = Config{<:Any,<:Any, Width}
 
 Configuration type to dispatch on in custom reverse rules (see [`augmented_primal`](@ref) and [`reverse`](@ref)).
-* `NeedsPrimal` and `NeedsShadow` are boolean values specifying whether the primal and shadow (resp.) should be returned. 
-* `Width` is an integer that specifies the number of adjoints/shadows simultaneously being propagated.
-* `Overwritten` is ??? (TODO) 
+* `NeedsPrimal` and `NeedsShadow`: boolean values specifying whether the primal and shadow (resp.) should be returned. 
+* `Width`: an integer that specifies the number of adjoints/shadows simultaneously being propagated.
+* `Overwritten`: a collection of booleans of whether each argument is modified between state (true if to be automatically derived).
```
wsmoses, line 28: a tuple of booleans. and true if potentially modified between (rather than "true if to be automatically derived").

lib/EnzymeCore/src/rules.jl
```diff
@@ -58,7 +79,7 @@ function augmented_primal end
     reverse(::Config, func::Annotation{typeof(f)}, ::Type{<:Annotation), tape, args::Annotation...)
 
 Takes gradient of derivative, activity annotation, and tape. If there is an active return dret is passed
-as Active{T} with the active return val. Otherwise dret is passed as Type{Duplicated{T}}, etc.
+as Active{T} with the derivative of the active return val. Otherwise dret is passed as Type{Duplicated{T}}, etc.
```
gaurav-arya, line 82: Is this a correct correction?

wsmoses: yup

examples/custom_rule.jl
```diff
@@ -0,0 +1,191 @@
...
+
+import .EnzymeRules: forward, reverse, augmented_primal
+using .EnzymeRules
+
+# In this section, we write a simple forward rule to start out:
+
+function forward(func::Const{typeof(f)}, ::Type{<:Duplicated}, y::Duplicated, x::Duplicated)
+    println("Using custom rule!")
+    out = func.val(y.val, x.val)
+    y.dval .= 2 .* x.val .* x.dval
```
gaurav-arya, line 55: is using `x.val` here incorrect, because it could have been mutated since? Do I need to use the tape if I want the rule to handle possible overwrites?

examples/custom_rule.jl
```diff
@@ -0,0 +1,191 @@
...
+# Our function returns an `AugmentedReturn` object specifying the return value, its shadow, and any extra tape information (none in this csae).
+# Note that we return a shadow of `nothing` since the return value is marked [`Active`](@ref).
+
+# Now, we write a method for [`EnzymeRules.reverse`](@ref):
+
+function reverse(config::ConfigWidth{1}, func::Const{typeof(f)}, out::Active, tape,
+                 y::Duplicated, x::Duplicated)
+    println("In custom reverse rule.")
+    y.dval .+= out.val 
+    x.dval .+= 2 .* x.val .* y.dval # accumulate into shadow, don't assign!
```
gaurav-arya, line 171: is using `x.val` here incorrect, because it could have been mutated since? Do I need to use the tape if I want the rule to handle possible overwrites?

wsmoses: It's correct if the modifiedBetween is false for that value, otherwise yeah you need the tape.

You can actually test this by modifying x in place after f in an overall function.

examples/custom_rule.jl
```diff
@@ -0,0 +1,191 @@
...
+# Finally, let's look at how to write a simple reverse-mode rule! 
+# First, we write a method for [`EnzymeRules.augmented_primal`](@ref):
+
+function augmented_primal(config::ConfigWidth{1}, func::Const{typeof(f)}, ::Type{<:Active},
+                          y::Duplicated, x::Duplicated)
+    println("In custom augmented primal rule.")
+    if needs_primal(config)
+        return AugmentedReturn(func.val(y.val, x.val), nothing, nothing)
+    else
+        return AugmentedReturn(nothing, nothing, nothing)
```
wsmoses, line 150: This isn't right since it won't set y in this case whereas the original code would.

examples/custom_rule.jl
gaurav-arya: For the array to put on the tape, do you recommend just allocating it inside `augmented_primal`, or doing something else?

wsmoses: Yeah exactly that.

examples/custom_rule.jl
```diff
@@ -0,0 +1,242 @@
...
+    x .= x.^2
+    return ret
+end
+
+x  = [3.0, 1.0]
+y  = [0.0, 0.0]
+dx .= 0
+dy .= 0
+
+## autodiff(Reverse, h, Duplicated(y, dy), Duplicated(x, dx))
```
gaurav-arya, line 240: needs #706

wsmoses:
It might also be nice to add some words on inactive . In essence it forcibly marks a function as not mattering for the sake of differentiation.

gaurav-arya:
I think docs should build now

examples/custom_rule.jl
```diff
@@ -0,0 +1,262 @@
...
+# (i.e. not by side effects or through its return values), we can simply use [`EnzymeRules.inactive`](@ref).
+# So long as there exists a matching dispatch to [`EnzymeRules.inactive`](@ref), the function will be considered inactive.
+# For example:
+
+function printhi()
+    println("Hi!")
+end
+
+EnzymeRules.inactive(::typeof(printhi), args...) = nothing
+
```
wsmoses, line 256: Maybe add another example like the following

myidentity(x) = x

inactive(typeof(x)) = ...

function tester(x)
  return x + myidentity(x)
 end
 
 ^ the above should return a derivative of 1.0 not 2.0

examples/custom_rule.jl
```diff
@@ -0,0 +1,266 @@
...
+function forward(func::Const{typeof(f)}, ::Type{<:Duplicated}, y::Duplicated, x::Duplicated)
+    println("Using custom rule!")
+    ret = func.val(y.val, x.val)
+    y.dval .= 2 .* x.val .* x.dval
+    return Duplicated(ret, sum(y.dval)) 
+end
+
+# In the signature of our rule, we have made use of `Enzyme`'s activity annotations. Let's break down each one:
+# - the [`Const`](@ref) annotation on `f` indicates that we accept a function `f` that does not have a derivative component,
+#   which makes sense since `f` itself does not depend on any parameters.
```
wsmoses, line 61: change to "since f itself is not a closure with data that could be differentiated".

examples/custom_rule.jl
```diff
@@ -0,0 +1,266 @@
...
+# * We accepted a [`EnzymeRules.Config`](@ref) object with a specified width of 1, which means that our rule does not support batched reverse mode.
+# * We annotated `f` with [`Const`](@ref) as usual.
+# * We dispatched on an [`Active`](@ref) annotation for the return value. This is a special annotation for scalar values, such as our return value,
+#   that indicates that that we care about the value's derivative but we need not explicitly allocate a mutable shadow since it is a scalar value.
+# * We annotated `x` and `y` with [`Duplicated`](@ref), similar to our first simple forward rule.
+
+# Now, let's unpack the body of our `augmented_primal` rule:
+# * We checked if the `config` requires the primal. If not, we need not compute the return value, but we make sure to mutate `y` in all cases.
+# * We checked if `x` could possibly be overwritten using the `Overwritten` attribute of [`EnzymeRules.Config`](@ref). 
+#   If so, we save the value of `x` on the `tape` of the returned [`EnzymeRules.AugmentedReturn`](@ref) objet.
```
wsmoses, line 192: save the value -> save the elements

examples/custom_rule.jl
```diff
@@ -0,0 +1,266 @@
...
+    println("In custom reverse rule.")
+    ## retrieve x value, either from original x or from tape if x may have been overwritten.
+    xval = overwritten(config)[3] ? tape : x.val 
+    ## accumulate into x's shadow, don't assign!
+    x.dval .+= 2 .* xval .* dret.val 
+    return ()
+end
+
+# The activities used in the signature match what we used for `augmented_primal`. 
+# One key difference is that we now receive an *instance* `dret` of [`Active`](@ref) for the return type, not just a type annotation.
```
wsmoses, line 208: This is only applicable for active returns. Duplicated/Const/etc still receives the type.

examples/custom_rule.jl
```diff
@@ -0,0 +1,266 @@
...
+
+# Now, we write a method for [`EnzymeRules.reverse`](@ref):
+
+function reverse(config::ConfigWidth{1}, func::Const{typeof(f)}, dret::Active, tape,
+                 y::Duplicated, x::Duplicated)
+    println("In custom reverse rule.")
+    ## retrieve x value, either from original x or from tape if x may have been overwritten.
+    xval = overwritten(config)[3] ? tape : x.val 
+    ## accumulate into x's shadow, don't assign!
+    x.dval .+= 2 .* xval .* dret.val
```
wsmoses, line 203: you should also add the following:

x.dval .+= 2 .* xval .* y.dval
y.dval = 0;

This is because in reverse mode AD you have to sum up the derivatives from all users. Your case accounts for the use inside of sum(...). However if y was read after our function, we need to consider derivatives from that use as well. (accounted for above).

lib/EnzymeCore/src/rules.jl
wsmoses: That change is fine with me.

examples/custom_rule.jl
gaurav-arya: So for Duplicated returns, would I put a ref to the shadow on the tape, if it's not provided to `reverse?

wsmoses: if you need it. If for example it was just returning an input, you wouldn't need to put it on the tape

lib/EnzymeCore/src/rules.jl
gaurav-arya: Shall I do that in a PR immediately after this gets merged? To keep this change purely docs

gaurav-arya:
@vchuravy we were wondering whether this literate example could also somehow be turned into doctests. Don't know if Literate.jl supports conventional doctests -- one solution is to just add hidden @assert's

PR #702, opened 2023-04-07
support invalidation for user-defined rules

### LINKED ISSUE #696: Custom rule not detected if defined after call to `autodiff`
MWE on main:
```julia
julia> using Enzyme

julia> import .EnzymeRules: forward

julia> f(x) = x^2

julia> autodiff(Forward, f, Duplicated(1.0, 1.0))
(2.0,)

julia> forward(::Const{typeof(f)}, ::Type{<:DuplicatedNoNeed}, x::Duplicated) = 10+2*x.val*x.dval
forward (generic function with 1 method)

julia> autodiff(Forward, f, Duplicated(1.0, 1.0)) # answer unchanged
(2.0,)

julia> f(x) = x^2 # redefine f and we get the correct answer

julia> autodiff(Forward, f, Duplicated(1.0, 1.0))
(12.0,)
```
The reason for the behaviour might be this line:
https://github.com/EnzymeAD/Enzyme.jl/blob/939f9b4086d62b07eca8107db9523a2f8fe043d3/src/Enzyme.jl#L313

aviatesk:
fixes #696

### DIFF
--- a/lib/EnzymeCore/src/rules.jl
+++ b/lib/EnzymeCore/src/rules.jl
@@ -82,23 +82,74 @@ function _annotate_tt(@nospecialize(TT0))
     return ft, tt
 end
 
-function has_frule_from_sig(@nospecialize(TT); world=Base.get_world_counter())
+function has_frule_from_sig(@nospecialize(TT);
+                            world::UInt=Base.get_world_counter(),
+                            method_table::Union{Nothing,Core.Compiler.MethodTableView}=nothing,
+                            caller::Union{Nothing,Core.MethodInstance}=nothing)
     ft, tt = _annotate_tt(TT)
     TT = Tuple{<:Annotation{ft}, Type{<:Annotation}, tt...}
-    isapplicable(forward, TT; world)
+    return isapplicable(forward, TT; world, method_table, caller)
 end
 
-function has_rrule_from_sig(@nospecialize(TT); world=Base.get_world_counter())
+function has_rrule_from_sig(@nospecialize(TT);
+                            world::UInt=Base.get_world_counter(),
+                            method_table::Union{Nothing,Core.Compiler.MethodTableView}=nothing,
+                            caller::Union{Nothing,Core.MethodInstance}=nothing)
     ft, tt = _annotate_tt(TT)
     TT = Tuple{<:Config, <:Annotation{ft}, Type{<:Annotation}, tt...}
-    isapplicable(augmented_primal, TT; world)
+    return isapplicable(augmented_primal, TT; world, method_table, caller)
 end
 
-# Base.hasmethod is a precise match we want the broader query.
-function isapplicable(@nospecialize(f), @nospecialize(TT); world=Base.get_world_counter())
+# `hasmethod` is a precise match using `Core.Compiler.findsup`,
+# but here we want the broader query using `Core.Compiler.findall`.
+# Also add appropriate backedges to the caller `MethodInstance` if given.
+function isapplicable(@nospecialize(f), @nospecialize(TT);
+                      world::UInt=Base.get_world_counter(),
+                      method_table::Union{Nothing,Core.Compiler.MethodTableView}=nothing,
+                      caller::Union{Nothing,Core.MethodInstance}=nothing)
     tt = Base.to_tuple_type(TT)
     sig = Base.signature_type(f, tt)
-    return !isempty(Base._methods_by_ftype(sig, -1, world)) # TODO cheaper way of querying?
+    @static if VERSION < v"1.7.0"
+        return !isempty(Base._methods_by_ftype(sig, -1, world))
+    end
+    mt = ccall(:jl_method_table_for, Any, (Any,), sig)
+    mt isa Core.MethodTable || return false
+    if method_table === nothing
+        method_table = Core.Compiler.InternalMethodTable(world)
+    end
+    result = Core.Compiler.findall(sig, method_table; limit=-1)
+    (result === nothing || result === missing) && return false
+    @static if isdefined(Core.Compiler, :MethodMatchResult)
+        (; matches) = result
+    else
+        matches = result
+    end
+    fullmatch = Core.Compiler._any(match::Core.MethodMatch->match.fully_covers, matches)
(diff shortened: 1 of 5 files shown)


lib/EnzymeCore/src/rules.jl
```diff
@@ -82,23 +82,79 @@ function _annotate_tt(@nospecialize(TT0))
...
+function isapplicable(@nospecialize(f), @nospecialize(TT);
+                      world::UInt=Base.get_world_counter(),
+                      method_table::MaybeMethodTableView=nothing,
+                      caller::Union{Nothing,Core.MethodInstance}=nothing)
     tt = Base.to_tuple_type(TT)
     sig = Base.signature_type(f, tt)
-    return !isempty(Base._methods_by_ftype(sig, -1, world)) # TODO cheaper way of querying?
+    mt = ccall(:jl_method_table_for, Any, (Any,), sig)
+    mt isa Core.MethodTable || return false
+    @static if isdefined(Core.Compiler, :method_table)
```
aviatesk, line 118: Keeping compatibility with Julia<1.8 is not very fun... Can Enzyme drop 1.6/1.7 compatibility?

vchuravy: 1.6 won't support custom rules. But folks have asked that we keep supporting 1.6 until the next LTS

lib/EnzymeCore/src/rules.jl
```diff
@@ -82,23 +82,79 @@ function _annotate_tt(@nospecialize(TT0))
...
+    end
+end
+
+function add_backedge!(caller::Core.MethodInstance, callee::Core.MethodInstance, @nospecialize(sig))
+    ccall(:jl_method_instance_add_backedge, Cvoid, (Any, Any, Any), callee, sig, caller)
+    return nothing
+end
+
+function add_mt_backedge!(caller::Core.MethodInstance, mt::Core.MethodTable, @nospecialize(sig))
+    ccall(:jl_method_table_add_backedge, Cvoid, (Any, Any, Any), mt, sig, caller)
```
wsmoses, line 156: Can you also add this for the `is_inactive_from_sig` below?

lib/EnzymeCore/src/rules.jl
wsmoses: Yeah so we need rules on 1.7, 1.6 support here is fine to drop [since it won't have rules anyways]

test/ruleinvalidation.jl
```diff
@@ -0,0 +1,27 @@
+module RuleInvalidation
+
+using Test
+using Enzyme
+import .EnzymeRules: forward
+
+issue696(x) = x^2
+
+@test autodiff(Forward, issue696, Duplicated(1.0, 1.0))[1] ≈ 2.0
```
vchuravy, line 9: We should also add `f(x) = issue696(x)` and see that `f` is invalidated.

aviatesk: Added.

lib/EnzymeCore/src/rules.jl
aviatesk: Done.

src/compiler.jl
```diff
@@ -7659,9 +7676,11 @@ end
             handleCustom("enz_noop", [StringAttribute("enzyme_inactive"; ctx), EnumAttribute("readonly"; ctx)])
             continue
         end
-        if EnzymeRules.is_inactive_from_sig(mi.specTypes; world)
-            handleCustom("enz_noop", [StringAttribute("enzyme_inactive"; ctx), StringAttribute("nofree"; ctx)])
-            continue
+        @static if EnzymeRules.supported()
```
aviatesk, line 7679:
```suggestion
        @static if EnzymeRules.issupported()
```

src/compiler.jl
```diff
@@ -7569,15 +7583,18 @@ end
...
-            has_custom_rule = EnzymeRules.has_frule_from_sig(specTypes; world)
-            if has_custom_rule
-                @safe_debug "Found frule for" mi.specTypes
-            end
-        else
-            has_custom_rule = EnzymeRules.has_rrule_from_sig(specTypes; world)
-            if has_custom_rule
-                @safe_debug "Found rrule for" mi.specTypes
+        caller = mi
+        @static if EnzymeRules.supported()
```
aviatesk, line 7587:
```suggestion
        @static if EnzymeRules.issupported()
```

aviatesk:
This is ready to go.

src/compiler.jl
```diff
@@ -7753,9 +7770,11 @@ end
             handleCustom("enz_noop", [StringAttribute("enzyme_inactive"; ctx), EnumAttribute("readonly"; ctx)])
             continue
         end
-        if EnzymeRules.is_inactive_from_sig(mi.specTypes; world)
-            handleCustom("enz_noop", [StringAttribute("enzyme_inactive"; ctx), StringAttribute("nofree"; ctx)])
-            continue
+        @static if EnzymeRules.issupported()
```
wsmoses, line 7773: Unfortunately, per the 1.6 test which we have disabled custom rules, we still need the activity checks to be available.

test/runtests.jl
```diff
@@ -47,6 +47,10 @@ if Enzyme.EnzymeRules.issupported()
     include("rrules.jl")
     include("kwrules.jl")
     include("kwrrules.jl")
+    if VERSION ≥ v"1.9-"
+        # XXX invalidation does not work on Julia 1.8
```
wsmoses, line 51: Is this check still needed?

aviatesk: Yes. The backedges are added properly on 1.7/1.8, but the invalidation doesn't work as expected. Probably because of some (already fixed) bugs within the Julia runtime system.

lib/EnzymeCore/src/rules.jl
```diff
@@ -82,23 +82,74 @@ function _annotate_tt(@nospecialize(TT0))
...
```
vtjnash, line 151: This is internal Base machinery of the NativeInterpreter, and should never be accessed by external clients.

Issue #709, opened 2023-04-09
Gradient of matrix determinant errors


jgreener64:
I am on Julia 1.8.5 and Enzyme main (73285ce).
```julia
using Enzyme, LinearAlgebra

x = rand(5, 5)
dx = zero(x)

autodiff(Reverse, det, Active, Duplicated(x, dx))
```
Error message, truncated to fit in the issue limit:
```
ERROR: LoadError: Enzyme compilation failed.
Current scope:
; ModuleID = 'text'
source_filename = "text"
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128-ni:10:11:12:13"
target triple = "x86_64-w64-mingw32"

; Function Attrs: uwtable
define internal fastcc noundef i8 @julia_istril_5503({} addrspace(10)* nonnull align 16 dereferenceable(40) %0) unnamed_addr #0 !dbg !36 {
top:
  %1 = alloca { [2 x i64], i64 }, align 8
  %2 = alloca [1 x i64], align 8
  %3 = alloca { {} addrspace(10)*, { [2 x i64], i64 }, i64, i64 }, align 8
  %4 = call {}*** @julia.get_pgcstack()
  %5 = bitcast {} addrspace(10)* %0 to {} addrspace(10)* addrspace(10)*, !dbg !38
  %6 = addrspacecast {} addrspace(10)* addrspace(10)* %5 to {} addrspace(10)* addrspace(11)*, !dbg !38
  %7 = getelementptr inbounds {} addrspace(10)*, {} addrspace(10)* addrspace(11)* %6, i64 3, !dbg !38
  %8 = bitcast {} addrspace(10)* addrspace(11)* %7 to i64 addrspace(11)*, !dbg !38
  %9 = load i64, i64 addrspace(11)* %8, align 8, !dbg !38, !tbaa !43, !range !47, !invariant.load !4
  %10 = getelementptr inbounds {} addrspace(10)*, {} addrspace(10)* addrspace(11)* %6, i64 4, !dbg !38
  %11 = bitcast {} addrspace(10)* addrspace(11)* %10 to i64 addrspace(11)*, !dbg !38
  %12 = load i64, i64 addrspace(11)* %11, align 16, !dbg !38, !tbaa !43, !range !47, !invariant.load !4
  %.not = icmp sgt i64 2, %12, !dbg !48
[...]

Issue #708, opened 2023-04-09
Sum of Int with floats fails


jgreener64:
I am on Julia 1.8.5 and Enzyme main (73285ce). The following works:
```julia
using Enzyme

f1(x) = sum([ifelse(i > 0, i, zero(i)) for i in x])

x = randn(5)
dx = zero(x)

autodiff(Reverse, f1, Active, Duplicated(x, dx))
```
However if I specify integer zero, it errors:
```julia
f2(x) = sum([ifelse(i > 0, i, 0) for i in x])

autodiff(Reverse, f2, Active, Duplicated(x, dx))
```
This is not ideal coding style since the intermediate vector is the non-concrete `Vector{Real}`, however the function does run fine without gradients. The error:
```
ERROR: Enzyme compilation failed due to illegal type analysis.
Current scope:
; Function Attrs: mustprogress willreturn
define nonnull {} addrspace(10)* @preprocess_julia_f2_5740_inner.1({} addrspace(10)* nonnull writeonly align 16 dereferenceable(40) %0) local_unnamed_addr #24 !dbg !741 {
entry:
  %1 = call {}*** @julia.get_pgcstack()
  %2 = bitcast {}*** %1 to {}**
  %3 = getelementptr inbounds {}*, {}** %2, i64 -12
  %4 = getelementptr inbounds {}*, {}** %3, i64 14
  %5 = bitcast {}** %4 to i8**
  %6 = load i8*, i8** %5, align 8
  %7 = call noalias nonnull dereferenceable(8) dereferenceable_or_null(8) {} addrspace(10)* @jl_gc_alloc_typed(i8* %6, i64 8, {} addrspace(10)* addrspacecast ({}* inttoptr (i64 1907595642928 to {}*) to {} addrspace(10)*)), !enzyme_fromstack !742
  %8 = bitcast {} addrspace(10)* %7 to { {} addrspace(10)* } addrspace(10)*, !enzyme_caststack !4
  %9 = bitcast { {} addrspace(10)* } addrspace(10)* %8 to i8 addrspace(10)*
  %10 = addrspacecast i8 addrspace(10)* %9 to i8*
  call void @llvm.lifetime.start.p0i8(i64 noundef 8, i8* noundef nonnull align 8 dereferenceable(8) %10) #25
  %11 = call {}*** @julia.get_pgcstack() #25
[...]

PR #710, opened 2023-04-09
Activate gradient tests

jgreener64:
I went to write some tests for Enzyme.jl and noticed that there are useful tests in DiffTests.jl which are currently disabled. This PR writes a testing function for vector/matrix to number functions and turns on those tests. It also turns on a number to number test that now works.

A couple of possible issues were uncovered and have been opened (#708 and #709). Those tests are turned off for now.

Tests pass on my machine. I did notice one stochastic failure due to the `1e-9` tolerance but I couldn't reproduce.

If turning these tests on sounds like a good strategy I can proceed with more.

### DIFF
--- a/test/DiffTests.jl
+++ b/test/DiffTests.jl
@@ -115,8 +115,9 @@ end
 
 self_weighted_logit(x) = inv(1.0 + exp(-dot(x, x)))
 
+# vec2num_6 fails due to #708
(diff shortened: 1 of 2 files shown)


wsmoses:
On CI various things fail. Perhaps change that error bound to 1e-6 ?

jgreener64:
I increased the tolerance, and also moved the random number generation away from zero to prevent domain errors when finite differencing chooses non-zero values.

I think CI should pass on the latest commit.

jgreener64:
Does this need anything else?

Issue #706, opened 2023-04-08
Calling convention mismatch error with custom reverse rule


gaurav-arya:
MWE:

```julia
using Enzyme
import .EnzymeRules: forward, reverse, augmented_primal
using .EnzymeRules

function f(y, x)
    y .= x.^2
    return sum(y) 
end

function augmented_primal(config::ConfigWidth{1}, func::Const{typeof(f)}, ::Type{<:Active},
                          y::Duplicated, x::Duplicated)
    println("In custom augmented primal rule.")
    # Compute primal
    if needs_primal(config)
        primal = func.val(y.val, x.val)
    else
        y.val .= x.val.^2 # y still needs to be mutated even if primal not needed!
        primal = nothing
    end
    # Save x in tape if x will be overwritten
    if overwritten(config)[3]
        tape = copy(x.val) 
    else
        tape = nothing
    end
    # Return an AugmentedReturn object with shadow = nothing
    return AugmentedReturn(primal, nothing, tape)
end

function reverse(config::ConfigWidth{1}, func::Const{typeof(f)}, out::Active, tape,
                 y::Duplicated, x::Duplicated)
    println("In custom reverse rule.")
    # retrieve x value, either from original x or from tape if x may have been overwritten.
    xval = overwritten(config)[3] ? tape : x.val 
    # accumulate into x's shadow, don't assign!
    x.dval .+= 2 .* xval .* out.val 
    return ()
end

x  = [3.0, 1.0]
y  = [0.0, 0.0]
dx = [0.0, 0.0]
dy = [0.0, 0.0]

function h(y, x)
    out = f(y, x)
    x .= x.^2
    return out
end

autodiff(Reverse, h, Duplicated(y, dy), Duplicated(x, dx))
```
And then I get a logged (not thrown) error:
```
[...]

PR #713, opened 2023-04-10
Fix custom reverse rule ordering bug

wsmoses:


wsmoses:
Fixes https://github.com/EnzymeAD/Enzyme.jl/issues/706

PR #715, opened 2023-04-10
Rename Config -> ReverseConfig

gaurav-arya:
Pros:

- Makes it more clear from the name that `Config` is for reverse mode

Cons:

- More typing
- Does this make it inconsistent with terminology in enzyme proper? If that's somewhat ingrained the change might not be worth it

### DIFF
--- a/examples/custom_rule.jl
+++ b/examples/custom_rule.jl
@@ -159,7 +159,7 @@ g(y, x) = f(y, x)^2 # function to differentiate
 # Let's look at how to write a simple reverse-mode rule! 
 # First, we write a method for [`EnzymeRules.augmented_primal`](@ref):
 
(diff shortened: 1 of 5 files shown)


wsmoses:
I'm fine with this, though don't forget to update some recently written example docs :P

wsmoses:
The config state is an artifical construct for Julia user-level information, since within Enzyme proper it's currently just queried on demand inside a `GradientUtils` meta structure (that also describes other analysis states/instruction and value mappings/etc).

gaurav-arya:
that makes sense 🙂 docs are now updated

vchuravy:
I don't think we ever are expecting to have a `ForwardConfig` and I don't like the length of the method definitions...

PR #728, opened 2023-04-13
Change `::Int64` to `::Int`

devmotion:
This PR is a simple find-replace of `::Int64` with `::Int`.

The restriction to `::Int64` breaks support for 32bit systems: https://github.com/TuringLang/Turing.jl/actions/runs/4687453137/jobs/8306774283?pr=1887#step:6:478

### DIFF
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -869,7 +869,7 @@ struct Return3
     ret2::Any
     ret3::Any
 end
-AnyArray(Length::Int64) = NamedTuple{ntuple(i->Symbol(i), Val(Length)),NTuple{Length,Any}}
+AnyArray(Length::Int) = NamedTuple{ntuple(i->Symbol(i), Val(Length)),NTuple{Length,Any}}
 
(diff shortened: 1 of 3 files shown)


tgymnich (approved):
OK from my side. Enzyme C API uses `unsigned` for the width.

wsmoses:
@devmotion this is sadly still broken per an Int64 used in an allocation function (see CI).

wsmoses:
@devmotion here https://github.com/EnzymeAD/Enzyme.jl/blob/02cefb14b5ad4c57a7d4f7f57750030f71068dc7/src/compiler.jl#L371 can you add:
```julia
if value_type(Size) != T_size_t
  Size = trunc!(B, Size, T_size_t
end
```

wsmoses:
@devmotion this still fails in an abi.jl test, but the fix isn't obvious from the CI log. If you have an X86 machine, would you be able to isolate the test which fails (and possibly add a Enzyme.API.printall!(true))

Issue #318, opened 2022-05-23
Better error for `Active{Vector{Float64}}`


vchuravy:
```
~/.julia/packages/GPUCompiler/XyxTy/src/utils.jl:35
ERROR: LoadError: Conversion of boxed type Vector{Float64} is not allowed
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:33
  [2] convert(::Type{LLVM.LLVMType}, typ::Type; ctx::LLVM.Context, allow_boxed::Bool)
    @ LLVM.Interop ~/.julia/packages/LLVM/szqwr/src/interop/base.jl:92
  [3] create_abi_wrapper(enzymefn::LLVM.Function, F::Type, argtypes::Vector{DataType}, rettype::Type, actualRetType::Type, Mode::Enzyme.API.CDerivativeMode, augmented::Nothing, dupClosure::Bool, width::Int64)
    @ Enzyme.Compiler ~/.julia/packages/Enzyme/aFZx5/src/compiler.jl:3234
  [4] enzyme!(job::GPUCompiler.CompilerJob{Enzyme.Compiler.EnzymeTarget, Enzyme.Compiler.EnzymeCompilerParams, GPUCompiler.FunctionSpec{typeof(loss), Tuple{Vector{Float64}, Vector{Vector{Float64}}, Vector{Vector{Float64}}, Vector{Vector{Float64}}, Vector{Vector{Float64}}, Matrix{Float64}, Matrix{Float64}, Matrix{Float64}}}}, mod::LLVM.Module, primalf::LLVM.Function, adjoint::GPUCompiler.FunctionSpec{typeof(loss), Tuple{Active{Vector{Float64}}, Const{Vector{Vector{Float64}}}, Const{Vector{Vector{Float64}}}, Const{Vector{Vector{Float64}}}, Const{Vector{Vector{Float64}}}, Const{Matrix{Float64}}, Const{Matrix{Float64}}, Const{Matrix{Float64}}}}, mode::Enzyme.API.CDerivativeMode, width::Int64, parallel::Bool, actualRetType::Type, dupClosure::Bool, wrap::Bool, modifiedBetween::Bool)
```

found by @dallasfoster

wsmoses:
This particularly will now throw an error for any abstractarray

PR #732, opened 2023-04-14
Better active of vector error

wsmoses:
Fixes https://github.com/EnzymeAD/Enzyme.jl/issues/318

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -17,7 +17,7 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 
 [compat]
 CEnum = "0.4"
-EnzymeCore = "0.3"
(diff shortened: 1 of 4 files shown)


simsurace:
I'm a little worried that this breaks `Active(::Tuple)` or `Active(::NamedTuple)`, which seemed to work fine as long as all values were e.g. `Float64`s. Shouldn't this keep working?

wsmoses:
Yeah I concur, which is why this hasn't landed.

wsmoses:
In reality we need a check that active only contains elements allocated inline/not vectors/ec

simsurace:
I guess `StaticVectors` would be fine. Maybe some sort of recursive `ismutable`?

wsmoses:
Superceeded elsewhere

Issue #735, opened 2023-04-16
cispi gradient is incorrect


ptiede:
I just noticed I was getting incorrect gradients when using the cispi function in some of my code. For some reason (missing a rule?) the gradient is zero, but only for cispi.

Here is a MWE
```julia
using Enzyme

f1(x) = cispi(x)
f2(x) = cis(x*π)
f3(x) = exp(x*1im*π)

autodiff(Reverse, f1, Active, Active(0.5)) #returns ((0.0,),) <-- Wrong
autodiff(Reverse, f2, Active, Active(0.5)) #returns ((-3.141592653589793,),) Correct
autodiff(Reverse, f3, Active, Active(0.5)) #returns ((-3.141592653589793,),) Correct
```

I'm currently using Julia 1.8.5 

```julia
julia> versioninfo()
Julia Version 1.8.5
Commit 17cfb8e65ea (2023-01-08 06:45 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 32 × AMD Ryzen 9 7950X 16-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, znver3)
  Threads: 1 on 32 virtual cores
Environment:
  JULIA_EDITOR = code
  JULIA_NUM_THREADS = 1
```

and I am using the main branch of Enzyme but this also occurs for 0.10 and 0,.11.
```
  [7da242da] Enzyme v0.11.0 `https://github.com/EnzymeAD/Enzyme.jl.git#main`
  [f151be2c] EnzymeCore v0.3.0 `https://github.com/EnzymeAD/Enzyme.jl.git:lib/EnzymeCore#main`
```

wsmoses:
Diving deeper, looks like the culprit is sincospi

PR #736, opened 2023-04-17
add sincospi

wsmoses:


wsmoses:
Fixes https://github.com/EnzymeAD/Enzyme.jl/issues/735 requires https://github.com/EnzymeAD/Enzyme/pull/1114

PR #739, opened 2023-04-18
Add rules for BLAS.dot, BLAS.dotc, and BLAS.dotu

sethaxen:
In an attempt to learn Enzyme's rule system and speed up AD of BLAS calls, this PR adds a rule for `BLAS.dot` (and `BLAS.dotc` and `BLAS.dotu`). On an input of length 10,000, this is 6x faster than the fallback in forward mode and 60x faster than the fallback in reverse mode.

### DIFF
--- a/test/rules/LinearAlgebra/blas.jl
+++ b/test/rules/LinearAlgebra/blas.jl
@@ -45,6 +45,13 @@ using Test
                 elseif Tret <: DuplicatedNoNeed
                     @test only(ret) ≈ dexp
                 end
+
+                if pfun === identity && sz == n && inc == 1
+                    @testset "consistency of 2-arg version" begin
+                        ret2 = autodiff(Forward, fun, Tret, x_annot, y_annot)
+                        @test ret2 == ret
+                    end
+                end
             end
 
             @testset for Tret in (BatchDuplicated, BatchDuplicatedNoNeed),
@@ -89,6 +96,12 @@ using Test
                 y[1] = 0
                 return d
             end
+            function fun_overwrite!(x, y)
+                d = fun(x, y)
+                x[1] = 0
+                y[1] = 0
+                return d
+            end
 
             @testset for Tret in (Const, Active),
                 Tx in (Const, Duplicated),
@@ -106,9 +119,9 @@ using Test
                 xcopy, ycopy, ∂xcopy, ∂ycopy = map(copy, (x, y, ∂x, ∂y))
 
                 x_annot =
-                    Tx <: Const ? Const(pfun(x)) : Duplicated(pfun(xcopy), pfun(∂xcopy))
+                    Tx <: Const ? Const(pfun(xcopy)) : Duplicated(pfun(xcopy), pfun(∂xcopy))
                 y_annot =
-                    Ty <: Const ? Const(pfun(y)) : Duplicated(pfun(ycopy), pfun(∂ycopy))
+                    Ty <: Const ? Const(pfun(ycopy)) : Duplicated(pfun(ycopy), pfun(∂ycopy))
                 activities = (Const(n), x_annot, Const(inc), y_annot, Const(inc))
 
                 vexp = fun(n, x, inc, y, inc)
@@ -138,6 +151,39 @@ using Test
                 @test ∂ycopy ≈
                     dexp[2] .* !(Ty <: Const || Tret <: Const) .+
                       ∂y .* ((Ty <: Const) .| (y .== ycopy))
+
+                if pfun === identity && sz == n && inc == 1
+                    @testset "consistency of 2-arg version" begin
+                        xcopy2, ycopy2, ∂xcopy2, ∂ycopy2 = map(copy, (x, y, ∂x, ∂y))
+                        x_annot = if Tx <: Const
+                            Const(pfun(xcopy2))
+                        else
+                            Duplicated(pfun(xcopy2), pfun(∂xcopy2))
+                        end
+                        y_annot = if Ty <: Const
+                            Const(pfun(ycopy2))
+                        else
+                            Duplicated(pfun(ycopy2), pfun(∂ycopy2))
+                        end
+                        activities = (x_annot, y_annot)
+                        fwd, rev = autodiff_thunk(
+                            ReverseSplitWithPrimal,
+                            Const{typeof(f)},
+                            Tret,
+                            map(typeof, activities)...,
+                        )
+                        tape, val2, shadow_val = fwd(Const(f), activities...)
+                        if Tret <: Const
+                            dval2, = rev(Const(f), activities..., tape)
+                        else
+                            dval2, = rev(Const(f), activities..., dret, tape)
+                        end
+                        @test all(isnothing, dval2)
+                        @test val2 == val
+                        @test ∂xcopy2 == ∂xcopy
+                        @test ∂ycopy2 == ∂ycopy
+                    end
+                end
             end
         end
     end


src/rules/LinearAlgebra/blas.jl
```diff
@@ -0,0 +1,72 @@
...
+    incy::Const{<:Integer},
+)
+    if EnzymeRules.needs_primal(config)
+        primal = func.val(n.val, X.val, incx.val, Y.val, incy.val)
+    else
+        primal = nothing
+    end
+    _, _, Xow, _, Yow = EnzymeRules.overwritten(config)
+    # copy only the elements we need
+    Xtape = Xow ? BLAS.blascopy!(n.val, X.val, incx.val, similar(X.val, n.val), 1) : nothing
```
sethaxen, line 48: Maybe it makes more sense to use the fallback instead of copying? Is that even possible?

sethaxen:
How does Enzyme treat complex numbers? e.g. if I wanted to also support `BLAS.dotc`, would the `Duplicated`s all store complex derivative vectors, and would the `Active` store a complex derivative?

wsmoses:
> How does Enzyme treat complex numbers? e.g. if I wanted to also support `BLAS.dotc`, would the `Duplicated`s all store complex derivative vectors, and would the `Active` store a complex derivative?

yup

src/rules/LinearAlgebra/blas.jl
```diff
@@ -0,0 +1,72 @@
+const ConstOrDuplicated{T} = Union{Const{T},Duplicated{T}}
+
```
wsmoses, line 2: Can you also add batch duplicated for these as well?

Otherwise someone successfully using the fallback right now in vector mode will now get a runtime rule not found error?

sethaxen: Is this related at all to `ConfigWidth`? i.e. if a `BatchDuplicated` is used, then will `ConfigWidth` indicate the number of shadows contained in the `BatchDuplicated`?

src/rules/LinearAlgebra/blas.jl
```diff
@@ -0,0 +1,115 @@
+module BLASRules
+
+using ..Enzyme
+using LinearAlgebra.BLAS
+
+const ConstOrDuplicated{T} = Union{Const{T},Duplicated{T}}
+
+_safe_similar(x::AbstractArray, n::Integer) = similar(x, n)
```
sethaxen, line 8: These utilities should be reusable for and greatly simplify the rules for all other Level 1 BLAS functions.

src/rules/LinearAlgebra/blas.jl
wsmoses: Yes, all duplicated will be of the same width. Here it must be passed as such in case there are No duplicated arguments

ZuseZ4:
Hi @sethaxen,
great to see some work on EnzymeBlas from another approach!
I worked on BlasEnzyme last year, but didn't got around merging it back then. 
Namely I do have tablegen implementations for 
```
handling: asum
handling: axpy
handling: copy
handling: dot
handling: scal
```
but I am lacking the tests for it.
Do you have a couple of tests for dot that I could use? Then I can finish my refactoring 
work [here](https://github.com/EnzymeAD/Enzyme/pull/1117) and re-use your tests once that's done. 
It would be great if we end up having support at all three levels, bitcode, enzyme proper and trough enzyme rules,
so we can compare those.

sethaxen:
> Do you have a couple of tests for dot that I could use? 

Sure, I just added some tests for `dot`, `dotu`, and `dotc`. Currently the `dot` tests pass on my machine, but the complex ones segfault on my machine, and it's not clear why. (I'm still a bit confused on how rules with complex numbers are supposed to be defined, see #744). All tests pass on the fallbacks.

>  It would be great if we end up having support at all three levels, bitcode, enzyme proper and julia level,
so we can compare those.

Yeah, it'd be nice if we could make that comparison when this PR is finished before I tackle BLAS rules for the other functions you listed.

src/rules/LinearAlgebra/blas.jl
```diff
@@ -0,0 +1,128 @@
...
+                tape = nothing
+            end
+
+            return EnzymeRules.AugmentedReturn(primal, shadow, tape)
+        end
+
+        function EnzymeRules.reverse(
+            config::EnzymeRules.ConfigWidth{1},
+            ::Const{typeof(BLAS.$fname)},
+            dret::Union{Active,Type{<:Const}},
```
sethaxen, line 104: @wsmoses the rules for `dot` and `dotc` are defined identically (the former for real inputs, the latter for complex inputs), but for `dotu` Enzyme wants to provide `dret` as a `Type{<:Duplicated}`, which wouldn't contain derivative information (i.e. `.dval`). Why is this?

```julia
julia> using Enzyme, LinearAlgebra

julia> n = 10;

julia> x, y = randn(n), randn(n);

julia> ∂x, ∂y = map(zero, (x, y));

julia> autodiff(Reverse, BLAS.dot, Active, Const(n), Duplicated(x, ∂x), Const(1), Duplicated(y, ∂y), Const(1))
┌ Warning: Using fallback BLAS replacements, performance may be degraded
└ @ Enzyme.Compiler ~/.julia/packages/GPUCompiler/BxfIW/src/utils.jl:56
((nothing, nothing, nothing, nothing, nothing),)

julia> x, y = randn(ComplexF64, n), randn(ComplexF64, n);

julia> ∂x, ∂y = map(zero, (x, y));

julia> autodiff(Reverse, BLAS.dotc, Active, Const(n), Duplicated(x, ∂x), Const(1), Duplicated(y, ∂y), Const(1))
┌ Warning: Using fallback BLAS replacements, performance may be degraded
└ @ Enzyme.Compiler ~/.julia/packages/GPUCompiler/BxfIW/src/utils.jl:56
ERROR: Enzyme execution failed.
Enzyme: No custom reverse rule was appliable for Tuple{EnzymeCore.EnzymeRules.ConfigWidth{1, true, true, (false, false, false, false, false, false)}, Const{typeof(LinearAlgebra.BLAS.dotc)}, Type{Duplicated{ComplexF64}}, Tuple{Nothing, Nothing}, Const{Int64}, Duplicated{Vector{ComplexF64}}, Const{Int64}, Duplicated{Vector{ComplexF64}}, Const{Int64}}
```

ZuseZ4:
I just build Enzyme.jl on top of Enzyme with Blas-Tblgen, removing the julia code that calls the Blas Fallback.
It seems like tablegen dot implementation passes the existing Julia tests and also the Enzyme LLVM-IR tests.
I need to clean it up a bit because I have multiple blas functions in my PR and not all of them are that well tested, but I hope to merge it soon.

sethaxen:
> It seems like tablegen dot implementation passes the existing Julia tests and also the Enzyme LLVM-IR tests.
> I need to clean it up a bit because I have multiple blas functions in my PR and not all of them are that well tested, but I hope to merge it soon.

@ZuseZ4 I'm interested to hear more about this tablegen approach. It seems to produce extremely terse expressions of the rules, which would in some ways be preferable to this approach here. But do you have do define both forward- and reverse-mode rules? And I'm a little confused by the existing rules. e.g. `axpy` should I believe have a forward-model rule
```julia
axpy!(∂a, X, ∂Y)
axpy!(a, ∂X, ∂Y)
```

and the reverse-mode rule
```julia
∂a = dot(X, ∂Y)
axpy!(conj(a), ∂Y, ∂X)
scal!(0, ∂Y)
```

But the tablegen rule looks like neither of these and even uses `asum`: https://github.com/EnzymeAD/Enzyme/blob/d679b813efb21bf038d240b8780eaaef3b08b3d8/enzyme/Enzyme/targets/BlasDerivatives.td#L80-L87

ZuseZ4:
Hi @sethaxen.
So the general approach is to declare reverse mode rules, fwd. should be able to be handled using the primal function itself. I agree that tablegen is "nicer" than the julia level for now, but this is just due to the fact that I wrote both sides. The Blas rules and the new tablegen.cpp code which parses this Blas rules. The julia rules are more generic and not fine-tuned for blas.

The incorrect reverse blas rules are mostly due to a design decision of my last approach. If say dot uses axpy for the reverse pass, tablegen checks the number and types of axpy arguments based on the axpy rules, which in turn required me to declare and implement all blas functions used for the reverse axpy pass. That recursively caused a typical "big bang" approach where I needed to handle multiple rules at once. Since I didn't had the tests back then I just ended up using placeholder (/incorrect) reverse rules to compile it and thus never merged it. I do have a better solution for it now, I'll push more of it on Monday

sethaxen:
@wsmoses the only failing tests are due to #778. Otherwise this should be ready for review.

src/rules/LinearAlgebra/blas.jl
```diff
@@ -0,0 +1,138 @@
...
+            incy::Const{<:Integer},
+        ) where {T<:BLAS.$Ttype}
+            primal, shadow = _maybe_primal_shadow(
+                config, func.val, (n.val, X.val, incx.val, Y.val, incy.val)
+            )
+
+            # build tape
+            if !(RT <: Const)
+                _, _, Xow, _, Yow = EnzymeRules.overwritten(config)
+                if Xow || BLAS.$fname === BLAS.dotu
```
sethaxen, line 89: An alternative to always taping for `dotu` would be to to do what we do hear when `X` or `Y` is overwritten but otherwise broadcast instead of using `axpy!` in `reverse` to include the conjugation operation.

sethaxen:
It seems the BLAS fallback warnings are raised even if the the BLAS fallbacks are not being hit?

wsmoses:
Yeah the BLAS fallback injection is done prior to any custom rules, so will always occur. However if a custom rule is hit it will use that implementation rather than the injected fallback.

ZuseZ4:
> > 
> 
> Ah, that makes sense. It'll be interesting to see how you handle `asum`, whose reverse-mode rule isn't computable just with BLAS functions.

That is indeed a bit annoying. We do have Blas-Tablegen and Instruction-Tablegen. The second one could handle this, but we will probably develop it further in the future, while Blas-Tablegen hopefully can stay unchanged in the future. So I might actually copy over a bit of the logic, such that we don't introduce a dependency there.
dot is btw. ready to be merged, so I'm now trying to get some more lv. 1 functions ready.

src/rules/LinearAlgebra/blas.jl
```diff
@@ -0,0 +1,148 @@
...
+                    Const,DuplicatedNoNeed,Duplicated,BatchDuplicatedNoNeed,BatchDuplicated
+                },
+            },
+            n::Const{<:Integer},
+            X::ConstOrBatchDuplicated{<:Union{Ptr{T},AbstractArray{T}}},
+            incx::Const{<:Integer},
+            Y::ConstOrBatchDuplicated{<:Union{Ptr{T},AbstractArray{T}}},
+            incy::Const{<:Integer},
+        ) where {T<:BLAS.$Ttype}
+            RT <: Const && return func.val(n.val, X.val, incx.val, Y.val, incy.val)
```
sethaxen, line 75: For some reason calling [the 2-arg `dot`](https://github.com/JuliaLang/julia/blob/7fc3e356e05ba46617c9b8fbbe228101eec0b6e6/stdlib/LinearAlgebra/src/blas.jl#L392), which forwards to the 5-arg `dot`, now segfaults with `Const` return type:
```suggestion
x, y, ∂x, ∂y = ntuple(_ -> randn(5), 4);
autodiff(Forward, Duplicated, Duplicated(x, ∂x), Duplicated(y, ∂y))  # fine
autodiff(Forward, Duplicated, Const(x), Duplicated(y, ∂y))  # fine
autodiff(Forward, Const, Duplicated(x, ∂x), Duplicated(y, ∂y))  # segfault
```
This is strange because only this line of code should be hit in this case, and all it does is call the primal function.

test/rules/LinearAlgebra/blas.jl
```diff
@@ -0,0 +1,132 @@
...
+                x_annot =
+                    Tx <: Const ? Const(pfun(x)) : Duplicated(pfun(xcopy), pfun(∂xcopy))
+                y_annot =
+                    Ty <: Const ? Const(pfun(y)) : Duplicated(pfun(ycopy), pfun(∂ycopy))
+
+                vexp = fun(n, x, inc, y, inc)
+                dexp = FiniteDifferences.j′vp(
+                    fdm, (x, y) -> fun(n, x, inc, y, inc), one(T), x, y
+                )
+                ret = autodiff(
```
sethaxen, line 113: Should use `autodiff_thunk` so that we can test pullback of a non-unit differential, but ran into #791.

src/rules/LinearAlgebra/blas.jl
sethaxen: The 2-arg methods also error for reverse-mode. These are the only remaining failures in the test suite.

wsmoses: Open an issue?

sethaxen:
All tests seem to pass for me. Some of the jobs in CI seem to not be picked up for some reason, but as far as I can tell, the same tests pass here that pass on main.

Here's an updated benchmark. The main takeaways are that the forward mode rules give a 2.5-7.3x speed-up vs the versions hit by main, and the reverse mode rules give a 13-60x speed-up when no tape is needed and a 5.7-8.7x speed-up when a tape is needed. In this latter case, the trade-off is that these rules allocate a tape, which in a case like the one benchmarked here (where future operations mutate only a single entry in the array) is wasteful.

```julia
using BenchmarkTools, Enzyme, LinearAlgebra, Random

Random.seed!(42)

n = 1_000
x = randn(n)
y = randn(n)
∂x = randn(eltype(x), size(x))
∂y = randn(eltype(y), size(y))
incx = incy = 1

# version that triggers tape allocation
function f_overwite!(f, n, x, incx, y, incy)
    s = f(n, x, incx, y, incy)
    x[1] = 0
    y[1] = 0
    return s
end

## BLAS.dot

@btime autodiff(
    Forward, $(BLAS.dot), $n, $(Duplicated(x, ∂x)), $incx, $(Duplicated(y, ∂y)), $incy,
)
# main: 731.130 ns (0 allocations: 0 bytes)
# here: 100.365 ns (2 allocations: 64 bytes)

# no tape needed
@btime autodiff(
    ReverseWithPrimal, $(BLAS.dot), Active, $n, Dx, $incx, Dy, $incy,
) setup=(Dx=Duplicated(x, copy(∂x)); Dy=Duplicated(y, copy(∂y)))
# main: 11.139 μs (2 allocations: 32 bytes)
# here: 192.827 ns (7 allocations: 208 bytes)

# tape needed
@btime autodiff(
    ReverseWithPrimal, $(f_overwite!), Active, $(BLAS.dot), $n, Dx, $incx, Dy, $incy,
) setup=(Dx=Duplicated(copy(x), copy(∂x)); Dy=Duplicated(copy(y), copy(∂y)))
# main: 11.624 μs (2 allocations: 32 bytes)
# here: 1.342 μs (9 allocations: 16.08 KiB)

T = ComplexF64
x = randn(T, n)
y = randn(T, n)
∂x = randn(eltype(x), size(x))
∂y = randn(eltype(y), size(y))
incx = incy = 1

## BLAS.dotu

@btime autodiff(
    Forward, $(BLAS.dotu), $n, $(Duplicated(x, ∂x)), $incx, $(Duplicated(y, ∂y)), $incy,
)
# main: 1.599 μs (0 allocations: 0 bytes)
# here: 650.117 ns (2 allocations: 64 bytes)

# no tape needed
@btime autodiff(
    ReverseWithPrimal, $(BLAS.dotu), Active, $n, Dx, $incx, Dy, $incy,
) setup=(Dx=Duplicated(x, copy(∂x)); Dy=Duplicated(y, copy(∂y)))
# main: 21.493 μs (2 allocations: 48 bytes)
# here: 1.706 μs (8 allocations: 256 bytes)

# tape needed
@btime autodiff(
    ReverseWithPrimal, $(f_overwite!), Active, $(BLAS.dotu), $n, Dx, $incx, Dy, $incy,
) setup=(Dx=Duplicated(copy(x), copy(∂x)); Dy=Duplicated(copy(y), copy(∂y)))
# main: 22.231 μs (2 allocations: 48 bytes)
# here: 3.885 μs (10 allocations: 31.75 KiB)

## BLAS.dotc

@btime autodiff(
    Forward, $(BLAS.dotc), $n, $(Duplicated(x, ∂x)), $incx, $(Duplicated(y, ∂y)), $incy,
)
# main: 1.584 μs (0 allocations: 0 bytes)
# here: 647.920 ns (2 allocations: 64 bytes)

# no tape needed
@btime autodiff(
    ReverseWithPrimal, $(BLAS.dotc), Active, $n, Dx, $incx, Dy, $incy,
) setup=(Dx=Duplicated(x, copy(∂x)); Dy=Duplicated(y, copy(∂y)))
# main: 21.220 μs (2 allocations: 48 bytes)
# here: 758.145 ns (8 allocations: 256 bytes)

# tape needed
@btime autodiff(
    ReverseWithPrimal, $(f_overwite!), Active, $(BLAS.dotc), $n, Dx, $incx, Dy, $incy,
) setup=(Dx=Duplicated(copy(x), copy(∂x)); Dy=Duplicated(copy(y), copy(∂y)))
# main: 22.251 μs (2 allocations: 48 bytes)
# here: 2.977 μs (10 allocations: 31.75 KiB)
```

@ZuseZ4 it would be interesting to see how the tablegen versions compare.

Issue #745, opened 2023-04-20
Error with LoopVectorization.jl


jgreener64:
I am on Julia 1.8.5, Enzyme main (4c30ed8b693a7a1d41f03bb4ef295201e26435bb) and LoopVectorization 0.12.157. This works:
```julia
using Enzyme, LoopVectorization

function f1(a, b)
    s = 0.0
    @inbounds @simd for i ∈ eachindex(a,b)
        s += a[i]*b[i]
    end
    s
end

a, b = rand(1000), rand(1000)
da, db = zero(a), zero(b)
autodiff(Reverse, f1, Active, Duplicated(a, da), Duplicated(b, db))
```
But this errors:
```julia
function f2(a, b)
    s = 0.0
    @turbo for i ∈ eachindex(a,b)
        s += a[i]*b[i]
    end
    s
end

da, db = zero(a), zero(b)
autodiff(Reverse, f2, Active, Duplicated(a, da), Duplicated(b, db))
```
```
ERROR: MethodError: no method matching length(::LLVM.VectorType)
Closest candidates are:
  length(::Union{Base.KeySet, Base.ValueIterator}) at abstractdict.jl:58
  length(::Union{ArrayInterfaceCore.BidiagonalIndex, ArrayInterfaceCore.TridiagonalIndex}) at ~/.julia/packages/ArrayInterfaceCore/o8YXM/src/ArrayInterfaceCore.jl:666
  length(::Union{ArrayInterface.BidiagonalIndex, ArrayInterface.TridiagonalIndex}) at ~/.julia/packages/ArrayInterface/RKU7b/src/ArrayInterface.jl:808
  ...
Stacktrace:
  [1] zero_single_allocation(builder::LLVM.IRBuilder, jlType::Type, LLVMType::LLVM.VectorType, nobj::LLVM.AllocaInst, zeroAll::Bool, idx::LLVM.ConstantInt, ctx::LLVM.Context)
    @ Enzyme.Compiler ~/.julia/dev/Enzyme/src/compiler.jl:5579
  [2] zero_allocation(B::Ptr{LLVM.API.LLVMOpaqueBuilder}, LLVMType::Ptr{LLVM.API.LLVMOpaqueType}, obj::Ptr{LLVM.API.LLVMOpaqueValue}, isTape::UInt8)
    @ Enzyme.Compiler ~/.julia/dev/Enzyme/src/compiler.jl:5546
[...]

PR #747, opened 2023-04-20
Fix llvm vector

wsmoses:
Fixes https://github.com/EnzymeAD/Enzyme.jl/issues/745

### DIFF
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -5575,8 +5575,16 @@ function zero_single_allocation(builder, jlType, LLVMType, nobj, zeroAll, idx, c
                 end
                 continue
             end
-            if isa(ty, LLVM.ArrayType) || isa(ty, LLVM.VectorType)
+            if isa(ty, LLVM.ArrayType) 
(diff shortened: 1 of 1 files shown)


src/compiler.jl
```diff
@@ -5575,14 +5575,22 @@ function zero_single_allocation(builder, jlType, LLVMType, nobj, zeroAll, idx, c
...
-            if isa(ty, LLVM.ArrayType) || isa(ty, LLVM.VectorType)
+            if isa(ty, LLVM.ArrayType) 
                 for i=1:length(ty)
                     npath = copy(path)
                     push!(npath, LLVM.ConstantInt(LLVM.IntType(32; ctx), i-1))
                     push!(todo, (npath, eltype(ty), eltype(jlty)))
                 end
                 continue
             end
+            if isa(ty, LLVM.VectorType)
```
vchuravy, line 5586: Instead of copying the entire code we should just branch on `size` vs `length`

vchuravy:
Also should have added a test.

PR #760, opened 2023-04-23
Bug Fix and use of new GPUCompiler persistent caching (8x improvement to loading time)

collinwarner:
Add functionality of GPUCompiler  https://github.com/JuliaGPU/GPUCompiler.jl/pull/425. Sees 8x improvement of cached loading times of the enzyme package.

Additionally latent bug found and fixed related to precompilation of Enzyme functions

### DIFF
--- a/src/Enzyme.jl
+++ b/src/Enzyme.jl
@@ -178,13 +178,11 @@ Enzyme.autodiff(ReverseWithPrimal, x->x*x, Active(3.0))
     ModifiedBetween = Val(falses_from_args(Val(1), args...))
 
     tt    = Tuple{map(T->eltype(Core.Typeof(T)), args′)...}
-    world = GPUCompiler.codegen_world_age(Core.Typeof(f.val), tt)
-    
     if A <: Active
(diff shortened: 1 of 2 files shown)


collinwarner:
image

collinwarner:
Bug Found:

The following script will error out with a segfault in the current Enzyme implementation. This occurs because the generated functions for thunk and deferred_codegen do no retrigger GPUCompiler and instead use stale values. Simply dropping the generated tag fixes the segfault but we are no longer able to look up unique method instances as the returned world ages are incorrect. Instead we should switch to using GPUCompiler.methodinstance directly dropping the need to filter on world ages.



```
module EnzymeTest                                                               
using Enzyme                                                                    
                                                                                
f1(x) = x*x                                                                     
                                                                                
@show autodiff(Reverse, f1, Active(1.0))                                        
                                                                                
                                                                                
function __init__()                                                             
    @show autodiff(Reverse, f1, Active(1.0))                                    
end                                                                             
                                                                                
end # module EnzymeTest 
```

vchuravy:
Can you split that last change out into a new PR?

collinwarner:
Changing to new PR, cleaning

PR #782, opened 2023-04-26
Add EnzymeTestUtils package for testing Enzyme rules

### LINKED ISSUE #780: Adding testing utilities
It would be handy to have test functions `test_forward` and `test_reverse` to simplify testing rule implementations, similar in spirit to `test_frule` and `test_rrule` in ChainRulesTestUtils (see e.g. https://juliadiff.org/ChainRulesTestUtils.jl/dev/#Canonical-example).

The main thing these testers would do are:
1. given an input argument and an activity type, generate a random (co)tangent and construct the activity
2. construct functions to pass to FiniteDifferences' `jvp` and `j′vp` so that arguments with a `Const` activity are not differentiated against
3. compare computed differentials with those simulated with FiniteDifferences, handling especially `Active` arguments in reverse-mode and `BatchDuplicated` in forward-mode

While none of these are very complicated, they involve a heavy amount of boilerplate that would be repeated for many custom rules, and they could be implemented generically to reduce testing a function to just point generation and calling the tester.

While these could initially just be used in Enzyme's test suite for simplifying its tests (e.g. those in #739), they could eventually be placed into their own package to standardize testing of rules.

sethaxen:
This PR adds functions `test_forward` and `test_reverse`, which support testing results of `autodiff` against FiniteDifferences. Fixes #780

Edit: `test_forward` and `test_reverse` are patterned off of `test_frule` and `test_rrule` in ChainRulesTestUtils.


test/testers.jl
```diff
@@ -0,0 +1,540 @@
...
+        T <: Union{BatchDuplicated,BatchDuplicatedNoNeed}
+    end
+    all_batch_or_const = all(activities) do T
+        T <: Union{BatchDuplicated,BatchDuplicatedNoNeed,Const}
+    end
+    return all_batch_or_const || no_batch
+end
+
+# meta-testing functions adapted from https://github.com/JuliaDiff/ChainRulesTestUtils.jl/blob/v1.10.1/test/meta_testing_tools.jl
+# Copyright (c) 2020 JuliaDiff
```
sethaxen, line 241: These are just here temporarily and will be removed once https://github.com/sethaxen/MetaTesting.jl is registered.

test/testers.jl
```diff
@@ -0,0 +1,816 @@
...
+    end
+end
+
+"""
+    test_reverse(f, Activity, args...; kwargs...)
+
+Test `Enzyme.autodiff` of `f` in `Reverse`-mode against finite differences.
+
+`f` has the same constraints as `Enzyme.autodiff_thunk`. If it mutates one of its arguments,
+it must not also return that argument. If the return value is a struct, then all floating
```
sethaxen, line 305: This limitation comes from us not being able to tell the difference between a shadow that is allocated in `augmented_primal` and the shadow of one of the inputs whose reference is returned by `augmented_primal`. Otherwise when constructing the `j′vp` call, we would take care to not add that input to the list of outputs. We could remove this limitation if the output is one of the inputs, but if instead the input is a field or element of the output, we could not detect that. So for clarity I've left this limitation as is.

sethaxen:
The test suite currently fails due to #874 #875 and #877 but otherwise `test_forward` and `test_reverse` are complete.

Two topics for discussion:
- These testers would be strictly more useful if we had some way of inferring all allowed activities for the inputs and outputs. Then the user would just pass the inputs to the tester, instead of explicitly looping through all activity types, and within the tester we would perform the looping. In particular, this would prevent cases where by adding a rule, a user broke batch forward-mode, but they don't know because they didn't test with `BatchDuplicated` activities. Is there functionality in Enzyme to enumerate this list of activities?
- Should these testers be moved into their own EnzymeTestUtils.jl package so that developers can use them in the test suites for their packages?

vchuravy:
> Is there functionality in Enzyme to enumerate this list of activities?

I think we used to have some code that guessed at the activity. But I think we deleted that code. Do you mean enumerate in the sense of "these are the activities that make sense for this type" or "these are all the activities Enzyme has"

sethaxen:
> Do you mean enumerate in the sense of "these are the activities that make sense for this type" or "these are all the activities Enzyme has"


The first one, but also filtering out all disallowed combinations. e.g. `Duplicated` with `BatchDuplicated`.

test/testers.jl
sethaxen: Sure!

sethaxen:
> That sounds like a great idea, put it next to EnzymeCore?

Done! I also added the docstrings to the Enzyme API docs, though that seems to maybe not be the right place for them?

Also, I can't see where the EnzymeCore tests are run in CI. Where do you recommend running the EnzymeTestUtils tests?

sethaxen:
> Also, I can't see where the EnzymeCore tests are run in CI. Where do you recommend running the EnzymeTestUtils tests?

I just added a new stage in the CI workflow

vchuravy:
So @wsmoses and I actually ended up using this to check correctness for some custom rules we were working on and quite liked it.

My one comment would be that because it is using `@test` the "failure" message is not that helpful. Maybe we could provide some better diagnostics?

sethaxen:
> My one comment would be that because it is using `@test` the "failure" message is not that helpful. Maybe we could provide some better diagnostics?

I've adapted ChainRulesTestUtils's mechanism for returning informative failure messages, and I've also tried to reduce the amount of unuseful information. How's this?

vchuravy:
Looks great! 

Except for 
```
 buildworker/worker/package_linux64/build/src/llvm-late-gc-lowering.cpp:870: std::vector<int, std::allocator<int> > LateLowerGCFrame::NumberAllBase(State&, llvm::Value*): Assertion `Tracked.size() == BaseNumbers.size()' failed.
```

That the tests are hitting xD

wsmoses:
Would you mind isolating the failing part of the test for us to remedy?

sethaxen:
> Except for
> 
> ```
>  buildworker/worker/package_linux64/build/src/llvm-late-gc-lowering.cpp:870: std::vector<int, std::allocator<int> > LateLowerGCFrame::NumberAllBase(State&, llvm::Value*): Assertion `Tracked.size() == BaseNumbers.size()' failed.
> ```
> 
> That the tests are hitting xD

On Julia v1.7 yes, but don't forget all these errors on v1.9: https://github.com/EnzymeAD/Enzyme.jl/actions/runs/5283317531/jobs/9559405514?pr=782#step:6:129

> Would you mind isolating the failing part of the test for us to remedy?

They should all be captured in #857, #874, and #877

wsmoses:
Yeah I'm not sure that `julia: /buildworker/worker/package_linux64/build/src/llvm-late-gc-lowering.cpp:870: std::vector<int, std::allocator<int> > LateLowerGCFrame::NumberAllBase(State&, llvm::Value*): Assertion `Tracked.size() == BaseNumbers.size()' failed.` error has an issue (all the others are distinct).

vchuravy:
@wsmoses it would be great if we could land this.

lib/EnzymeTestUtils/src/finite_difference_calls.jl
```diff
@@ -0,0 +1,144 @@
...
+# Arguments
+- `fdm::FiniteDifferenceMethod`: How to numerically differentiate `f`.
+- `f`: The function to differentiate.
+- `ȳ`: The cotangent of the primal output `y=f(xs...)`.
+- `activities`: activities that would be passed to `Enzyme.autodiff`
+
+# Returns
+- `x̄s`: Derivatives of output `s` w.r.t. `xs` estimated by finite differencing.
+=#
+function _make_j′vp_call(fdm, f, ȳ, activities)
```
wsmoses, line 58: Can we not have a backtick in this name?

wsmoses:
Given the CI burden atm, perhaps this makes sense to be a package in a different repo (and therefore doesn't run internal EnzymeTestUtils tests on all Enzyme.jl commits)?

thoughts @vchuravy 

regardless supportive of the PR and LGTM besides removing backtick in fn name.

lib/EnzymeTestUtils/src/finite_difference_calls.jl
sethaxen: The backtick is present in the FiniteDifferences function it is calling, after which it is named.

vchuravy:
I think we need to have it in the mono-repo. Eventually we might want to migrate the Enzyme.jl tests to use EnzymeTestUtils.

sethaxen:
An alternative then is to only run these tests on PRs marked with a specific label. Then on the rare PRs that modify EnzymeTestUtils, that label can be set to run the test suite, but other PRs don't waste the compute on them.

The downside is that if an Enzyme PR accidentally breaks EnzymeTestUtils, that would be missed.

vchuravy:
The EnzymeTestUtils test take 3min x3 so I  don't think they are a burden.

sethaxen:
Okay, finally have a version that passes tests on the latest release! Now to make sure issues are opened for all exceptions, that we test BatchDuplicated on reverse -mode, and that tests pass on v1.7.

sethaxen:
@wsmoses @vchuravy the tests now pass on v1.7 and v1.9. I had to add a lot of exceptions (tests marked as skip or broken) to work around version-specific errors and failures that occur when Enzyme differentiates Julia code. Not all are captured by issues, but I don't really have time right now to open issues with MWEs and errors for them all.

Reverse-mode tests could use more tests for batch mode, but these could be added in a later PR.

I suggest that this is ready for merge and initial package registration.

wsmoses (approved):

sethaxen:
Thanks @wsmoses! Looks like it stilled needs a registration with `@JuliaRegistrator register subdir="lib/EnzymeTestUtils"`

PR #783, opened 2023-04-26
call `__init__` during precompilation to make sure it gets precompiled

KristofferC:
The `__init__` method does not run during precompilation but it is attempted to get compiled. However, for this package it seems that there is some dynamic dispatch which means that precompilation cannot follow the call chain all the way. Calling the method explicitly helps and since it is quite expensive, gives a good improvement to load time.

Before:

```julia
julia> @time using Enzyme
[ Info: Precompiling Enzyme [7da242da-08ed-463a-9acd-ee780be4f1d9]
 44.283140 seconds (8.50 M allocations: 585.708 MiB, 0.36% gc time, 11.54% compilation time)

 julia> @time using Enzyme
 5.449457 seconds (8.47 M allocations: 583.942 MiB, 3.33% gc time, 92.65% compilation time)
```

After:

```julia
julia> @time using Enzyme
[ Info: Precompiling Enzyme [7da242da-08ed-463a-9acd-ee780be4f1d9]
 47.372864 seconds (798.57 k allocations: 49.376 MiB, 0.06% gc time, 0.08% compilation time)

julia> @time using Enzyme
  0.420285 seconds (775.94 k allocations: 47.622 MiB, 4.59% gc time, 3.96% compilation time)
```

I don't think the `__init__` method caches anything that will end up in the precompile file. Another way to fix this might be to improve the type inference for the methods in `__init__`.

### DIFF
--- a/src/Enzyme.jl
+++ b/src/Enzyme.jl
@@ -179,7 +179,7 @@ Enzyme.autodiff(ReverseWithPrimal, x->x*x, Active(3.0))
 
     tt    = Tuple{map(T->eltype(Core.Typeof(T)), args′)...}
     world = GPUCompiler.codegen_world_age(Core.Typeof(f.val), tt)
-    
+
     if A <: Active
         tt    = Tuple{map(T->eltype(Core.Typeof(T)), args′)...}
         rt = Core.Compiler.return_type(f.val, tt)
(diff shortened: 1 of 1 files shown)


src/Enzyme.jl
```diff
@@ -989,5 +989,7 @@ end
     mapreduce(LinearAlgebra.adjoint, vcat, rows)
 end
 
+# Precompile the init
+# Compiler.__init__()
```
vchuravy, line 993: Is this meant to be commented out?

KristofferC: Oh, hah.. that was for my benchmark without this

vchuravy:
@wsmoses we could just do:

```
    current_task_offset()
@static if VERSION < v"1.7.0"
else
    current_ptls_offset()
end
```

During precompilation instead of running it in `__init__`. Ideally Julia would just expose an API for us to use.

PR #792, opened 2023-04-27
Canonicalize TT

wsmoses:
Requires https://github.com/EnzymeAD/Enzyme/pull/1136


vchuravy (approved):

wsmoses:
@vchuravy can you investigate the CUDA.jl stuff prior to merge. I'm guessing its related to the DI set I commented out.

The reason why the particular DI was commented out is because it is illegal to have the same DI node for multiple functions, and if the wrapped function is recursive it wouldn't get inlined leading to that error hence the removal.

IMO the correct way of fixing is to clone the DI/construct a new one, but LLVM.jl doesn't seem up for the task yet.

PR #794, opened 2023-04-27
add blas attrs on the julia side

ZuseZ4:
Just experimenting, doesn't fix the julia-tblgen-blas failure for some reason yet.
- We should also set attrs per arg, not only per function.
- Already set by blas-tblgen in the AdjointGenerator, but that's pretty late. Probably better to factor this out once it's working, so we can call it during AA and don't have to re-implement it for julia.


ZuseZ4:
@wsmoses requires an enzyme bump, but other than that it seems good to merge?

PR #807, opened 2023-05-04
Fix thunk/deferred_codegen getting incorrect world age and segfaulting

collinwarner:
Bug Found:

The following script will error out with a segfault in the current Enzyme implementation. This occurs because the generated functions for thunk and deferred_codegen do no retrigger GPUCompiler and instead use stale values. Simply dropping the generated tag fixes the segfault but we are no longer able to look up unique method instances as the returned world ages are incorrect. Instead we should switch to using GPUCompiler.methodinstance directly dropping the need to filter on world ages.

```
module EnzymeTest                                                               
using Enzyme                                                                    
                                                                                
rosenbrock(x, y) = (1.0 - x)^2 + 100.0 * (y - x^2)^2                            
autodiff(Reverse, rosenbrock, Active, Active(1.0), Active(2.0))                 
                                                                                
function __init__()                                                             
    autodiff(Reverse, rosenbrock, Active, Active(1.0), Active(2.0))             
end                                                                             
                                                                                
end # module EnzymeTest   
```

With the following segfault:


```
julia> using EnzymeTest                                                                                                                            
[ Info: Precompiling EnzymeTest [87ed9522-ec2b-4441-9f14-f86404eed9ba]
                                                                                                                                                                                                                         
[1273377] signal (11.1): Segmentation fault                                                                                                        
in expression starting at REPL[1]:1                                                                                                                
unknown function (ip: 0x7f6f218d4050)                                                                                                              
macro expansion at /home/collinw/.julia/dev/Enzyme/src/compiler.jl:8971 [inlined]                                                                  
enzyme_call at /home/collinw/.julia/dev/Enzyme/src/compiler.jl:8663 [inlined]                                                                      
CombinedAdjointThunk at /home/collinw/.julia/dev/Enzyme/src/compiler.jl:8626 [inlined]                                                             
autodiff at /home/collinw/.julia/dev/Enzyme/src/Enzyme.jl:205 [inlined]                                                                            
autodiff at /home/collinw/.julia/dev/Enzyme/src/Enzyme.jl:214 [inlined]                                                                            
runGradients at /home/collinw/julia_master/EnzymeTest/src/EnzymeTest.jl:9 [inlined]                                                                
__init__ at /home/collinw/julia_master/EnzymeTest/src/EnzymeTest.jl:30                                                                             
Allocations: 7890557 (Pool: 7878530; Big: 12027); GC: 11                                                                                           
Segmentation fault (core dumped)                              
```

### DIFF
--- a/src/Enzyme.jl
+++ b/src/Enzyme.jl
@@ -178,13 +178,12 @@ Enzyme.autodiff(ReverseWithPrimal, x->x*x, Active(3.0))
     ModifiedBetween = Val(falses_from_args(Val(1), args...))
 
     tt    = Tuple{map(T->eltype(Core.Typeof(T)), args′)...}
-    world = GPUCompiler.codegen_world_age(Core.Typeof(f.val), tt)
-    
+
     if A <: Active
         tt    = Tuple{map(T->eltype(Core.Typeof(T)), args′)...}
         rt = Core.Compiler.return_type(f.val, tt)
         if !allocatedinline(rt) || rt isa Union
-            forward, adjoint = Enzyme.Compiler.thunk(Val(world), FA, Duplicated{rt}, tt′, #=Split=# Val(API.DEM_ReverseModeGradient), Val(width), ModifiedBetween, #=ReturnPrimal=#Val(ReturnPrimal), #=ShadowInit=#Val(true))
+            forward, adjoint = Enzyme.Compiler.thunk(FA, Duplicated{rt}, tt′, #=Split=# Val(API.DEM_ReverseModeGradient), Val(width), ModifiedBetween, #=ReturnPrimal=#Val(ReturnPrimal), #=ShadowInit=#Val(true))
             res = forward(f, args′...)
             tape = res[1]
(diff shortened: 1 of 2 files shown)


collinwarner:
To fix the segfault the `@generated` tags are removed so that GPUCompilation can occur whenever this function is called. Previously precompilation caching was storing stale function pointers. With this change we no longer get segfaults but instead are unable to look up the correct methodinstance on initialization due to mismatching world ages. To fix this` GPUCompiler.codegen_world_age()` is no longer used infavor of calling `GPUCompiler.methodinstance` inside thunk/deferred_codgen. Tested on failing testcase and achieve desired results. No segfaults and run of both autodiff calls.

wsmoses:
to deduplicate conversatoin channels cc @vtjnash @maleadt

wsmoses (changes requested):
Per slack, this does not resolve underlying issue.

wsmoses:
Superceeded by https://github.com/EnzymeAD/Enzyme.jl/pull/1575

PR #808, opened 2023-05-04
add persistent caching 8x speedup

collinwarner:
Add functionality of GPUCompiler https://github.com/JuliaGPU/GPUCompiler.jl/pull/425. Sees 8x improvement of cached loading times of the enzyme package.


collinwarner:
image

PR #824, opened 2023-05-09
Iterate to_tuple_type

wsmoses:
Fails to fix https://github.com/EnzymeAD/Enzyme.jl/issues/823

### DIFF
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -4914,7 +4914,7 @@ function boxfloat_augfwd(B::LLVM.API.LLVMBuilderRef, OrigCI::LLVM.API.LLVMValueR
     B = LLVM.IRBuilder(B)
 
     flt = value_type(origops[1])
-    TT = to_tape_type(flt)
+    TT = tape_type(flt)
 
     if width == 1
         obj = emit_allocobj!(B, TT)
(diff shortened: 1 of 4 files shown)


src/compiler.jl
```diff
@@ -5904,48 +5904,58 @@ base_type(T::UnionAll) = base_type(T.body)
...
+        for i in 1:nelems
+            e = LLVM.API.LLVMStructGetTypeAtIndex(Type, i-1)
             T, sub = to_tape_type(e)
             containsAny |= sub
             push!(tys, T)
         end
         Tup = Tuple{tys...}
         if containsAny
-            return AnonymousStruct(Tup), false
+            return NamedTuple{ntuple(i->Symbol(i), Val(Int(nelems))), Tup}, false
```
vchuravy, line 5921: That shouldn't be faster?

wsmoses:
Yeah I'm not sure how to fix (various attempts bove). This is clearly stuck in Julia type inference (I presume for the tuple/ntuple/namedtuple parts).

src/compiler.jl
wsmoses: now, with the explicit creation of the tuple via an array of symbols, it is many many orders of magnitude faster

PR #842, opened 2023-05-13
Add tests for BLAS.dot, BLAS.dotc, and BLAS.dotu

sethaxen:
As requested by @ZuseZ4, this PR just copies the tests from #739 but not the EnzymeRules.


ZuseZ4:
Didn't intend to close, will merge it once I added coverage for dotu, dotc

test/runtests.jl
```diff
@@ -76,6 +76,7 @@ include("typetree.jl")
         # XXX invalidation does not work on Julia 1.8
         include("ruleinvalidation.jl")
     end
+    include("rules/LinearAlgebra/blas.jl")
```
wsmoses, line 79: Can we have this run, even if enzymerules are not supported (thus to test tablegen and fallback blas)?

wsmoses: similarly can we move the test file into just tests rather than rules?

wsmoses (approved):

sethaxen:
I updated this PR to use EnzymeTestUtils.

wsmoses:
sgtm and still approved to merge (once tests pass). @ZuseZ4 was also wondering if you'd be able to add tests for some other BLAS functions as well.

sethaxen:
Sure, which ones?

wsmoses:
@ZuseZ4 mentioned gemv/gemm/scal/axpy/spmv

also it looks like CI fails?

ZuseZ4:
> Sure, which ones?

I did implement scal, axpy, gemv, spmv, gemm,
since these are the relevant ones for Julia.
So tests for these would be the most helpful for me.

sethaxen:
> also it looks like CI fails?

Pretty sure the failure is due to https://github.com/JuliaLang/Pkg.jl/issues/1585. One way to workaround this is to not have a separate `Project.toml` in `test/`. But trying a different workaround.

> I did implement scal, axpy, gemv, spmv, gemm,

Okay, I'll add this once I fix the CI

ZuseZ4:
thanks! Will try how tablegen does against the testcases tomorrow.

sethaxen:
> thanks! Will try how tablegen does against the testcases tomorrow.

Are the tablegen rules currently not being used? That could explain why when I added rules for `scal!`, I got a number of failures.

PR #883, opened 2023-05-30
Use Ubuntu 20.04 and renable assert builds

vchuravy:


vchuravy:
Assertions are green, which is nice.

Issue #810, opened 2023-05-07
Implementation in DynamicExpressions.jl


MilesCranmer:
I'm close to getting a working implementation of Enzyme.jl-generated derivatives for use in [DynamicExpressions.jl](https://github.com/SymbolicML/DynamicExpressions.jl) (and by extension [SymbolicRegression.jl](http://github.com/MilesCranmer/SymbolicRegression.jl) and [PySR](https://github.com/MilesCranmer/PySR)). I was wondering if you could help me with a couple of things related to this?

## 1. Disabling warnings

How do I turn off the `Recursive type` warnings?

```
┌ Warning: Recursive type
│   T = Node{Float64}
└ @ Enzyme /dev/shm/.julia/packages/GPUCompiler/BxfIW/src/utils.jl:56
```

I get about 100 of these on the first compile. It hasn't seemed to cause issues though (though maybe see 3. below), so I'm comfortable just turning it off for downstream users so people using the Python frontend aren't confused.

## 2. Improving speed

Currently it seems that the hand-rolled autodiff I wrote internally seems to have slightly better performance. The code I wrote isn't bad per se, but because the forward evaluation is so much more optimized in the code, my hope would be that Enzyme.jl could beat the internal autodiff by taking advantage of that.

Here is what I have so far. `eval_tree_array` takes in a matrix of shape (nfeatures, nrows), and returns a vector of shape (nrows,). 

```julia
function _eval_tree_array!(
    result::AbstractVector{T},
    tree::Node{T},
    X::AbstractMatrix{T},
    operators::OperatorEnum,
) where {T}
    out, completed = eval_tree_array(tree, X, operators; turbo=false)
    if !completed
        result .= T(NaN)
    else
        result .= out
    end
    return nothing
end

function enzyme_forward_diff_variables(
    i::Int, tree::Node{T}, X::AbstractMatrix{T}, operators::OperatorEnum
) where {T}
    nfeatures, nrows = size(X)
    result = zeros(T, nrows)
    dresult = ones(T, nrows)
    dX = zeros(T, nfeatures, nrows)
    dX[i, :] .= T(1)
    ###############################
    # Enzyme code: ################
    ###############################
    Enzyme.autodiff(
        Forward,
        _eval_tree_array!,
        Duplicated(result, dresult),
        Const(tree),
        Duplicated(X, dX),
        Const(operators),
    )
    ###############################
    ###############################
    return dresult
end

function enzyme_forward_gradient(
[...]

PR #891, opened 2023-06-04
Avoid recursive warning on recursive types

MilesCranmer:
Right now when you perform autodiff on a recursive type, you see a warning about it. However, this warning is printed recursively at each node in the type hierarchy, so you can get thousands of warning messages about the same type.

This implements a simple fix to only warn about a recursive type *once*. Later warnings are skipped.

Fixes part 1 of #810

### DIFF
--- a/src/typetree.jl
+++ b/src/typetree.jl
@@ -143,13 +143,27 @@ else
     ismutabletype(T) = isa(T, DataType) && T.mutable
 end
 
+const AlreadyWarned = Dict{Symbol,Bool}()
+const AlreadyWarnedLock = Threads.SpinLock()
+
+function recursive_type_warning(@nospecialize(T))
+    haskey(AlreadyWarned, Symbol(T)) && return nothing
+    lock(AlreadyWarnedLock) do
+        haskey(AlreadyWarned, Symbol(T)) && return nothing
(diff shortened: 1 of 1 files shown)


src/typetree.jl
```diff
@@ -143,13 +143,27 @@ else
...
+end
+
 function typetree(@nospecialize(T), ctx, dl, seen=nothing)
     if T isa UnionAll || T isa Union || T == Union{} || Base.isabstracttype(T)
         return TypeTree()
     end
 
     if seen !== nothing && T ∈ seen
-        GPUCompiler.@safe_warn "Recursive type" T
+        recursive_type_warning(T)
```
vchuravy, line 166: Maybe we could simply demote the warning to a debug statement? It's not like something the user will be able to do anything about it

wsmoses: I'm also debating if we just get rid of the warning. The only implication of it is that we don't fill out a recursive type tree, so slightly less info.

MilesCranmer: FWIW I haven't had issues with recursive types despite the warning.

Feel free to edit my PR directly by the way.

vchuravy (commented):
remove warning entirely

PR #893, opened 2023-06-04
Better errors

vchuravy:


wsmoses (approved):

src/compiler.jl
```diff
@@ -9219,12 +9219,23 @@ end
...
-            @assert length(argtypes) + is_adjoint + needs_tape == length(argexprs)
+            nexpected = nargs + is_adjoint + needs_tape
         elseif rettype <: Const
-            @assert length(argtypes)              + needs_tape == length(argexprs)
+            nexpected = nargs              + needs_tape
         else
-            @assert length(argtypes)              + needs_tape == length(argexprs)
+            nexpected = nargs              + needs_tape
+        end
+        if nexpected != nprovided
```
wsmoses, line 9231: since this is generated, do we want to quote this, then return out of the generated function.

Likewise would this be good to be a methoderror?

wsmoses:
Closed in favor of https://github.com/EnzymeAD/Enzyme.jl/pull/1444

PR #897, opened 2023-06-07
Adapt to new custom error return

wsmoses:


lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -149,4 +149,38 @@ function autodiff_deferred_thunk end
...
+@inline guaranteed_inactive(::Type{String}) = true
+@inline guaranteed_inactive(::Type{Symbol}) = true
+@inline guaranteed_inactive(::Type{T}) where T<:Array = guaranteed_inactive(eltype(T))
+
+@inline zero_internal(x::T, stackdict) where T<:Integer = x
+@inline zero_internal(x::T, stactdict) where T<:AbstractFloat = T(0)
+@inline zero_internal(x::String, stactdict) where = x
+@inline zero_internal(x::Symbol, stactdict) where = x
+
+@inline function zero_internal(x::T, stactdict) where T<:Array
```
vchuravy, line 170: Hm... Not the biggest fan.

I think the API is first and foremost `zero` which would call `Base.zero` so that we can define `Duplicated(x) = Duplicated(x, zero(x))`

And then a second API that takes the stack dict and calls zero if not present. Maybe call it `shadowfor`

wsmoses: Similarly we want this zero to do the structural zero equivalent (e.g. correct support for sparse datastructures) not the semantic zero that might otherwise be perscribed.

wsmoses: This was the particular inspiration for the API: https://github.com/JuliaLang/julia/blob/d7d7ffa7f964a9acf9261833bb0c98d442586ef9/base/deepcopy.jl#LL16C2-L18C20

wsmoses:
Superceeded by existing functionality

PR #899, opened 2023-06-09
Mark safepoint inactive and nofree

wsmoses:


vchuravy (approved):

wsmoses:
Nightly now gets further, but still fails at an activity mismatch. But hopefully soon we'll be able to see backtraces for that!

PR #907, opened 2023-06-23
reinsert gc marker before lowering

wsmoses:


src/compiler/optimize.jl
```diff
@@ -1006,6 +1006,7 @@ end
 
 function addJuliaLegalizationPasses!(pm, lower_intrinsics=true)
     if lower_intrinsics
+        add!(pm, FunctionPass("ReinsertGCMarker", reinsert_gcmarker_pass!))
```
vchuravy, line 1009: After barrier noop?

vchuravy (approved):

Issue #909, opened 2023-06-24
Handle the new calling convention that julia emits for cheaper ptls


motabbara:
Looks like something has gone quite wrong here with the presence of swiftcc and swiftself as a first argument. 

MWE on Julia 1.10 7b565e3d8dc. toml/manifest attached

```
using Enzyme

function foo(a::Float32)
    a*40.0
end

autodiff(Reverse, foo, Active, Active(3.0f0))

```

---

mod = ; ModuleID = 'start'
source_filename = "start"
target datalayout = "e-m:o-i64:64-i128:128-n32:64-S128-ni:10:11:12:13"
target triple = "arm64-apple-darwin22.4.0"

; Function Attrs: alwaysinline returns_twice
define internal swiftcc double @julia_foo_5929_inner({}*** nonnull swiftself %0, float %1) #0 !dbg !4 {
top:
  %2 = call {}*** @julia.get_pgcstack()
  %3 = bitcast {}*** %2 to {}**
  %current_task = getelementptr inbounds {}*, {}** %3, i64 -14
  %4 = bitcast {}** %current_task to i64*
  %world_age = getelementptr inbounds i64, i64* %4, i64 15
  %5 = bitcast {}*** %2 to {}**
  %current_task1 = getelementptr inbounds {}*, {}** %5, i64 -14
  %ptls_field = getelementptr inbounds {}*, {}** %current_task1, i64 16
  %ptls_load = load {}*, {}** %ptls_field, align 8, !tbaa !8
  %ptls = bitcast {}* %ptls_load to {}**
  %6 = bitcast {}** %ptls to i64**
  %7 = getelementptr inbounds i64*, i64** %6, i64 2
  %safepoint = load i64*, i64** %7, align 8, !tbaa !12, !invariant.load !7
  fence sIncorrect number of arguments passed to called function!
  %2 = call swiftcc double @julia_foo_5929_inner({}*** %1), !dbg !31
LLVM.API.LLVMVerifyFunction(wrapper_f, LLVM.API.LLVMPrintMessageAction) = 1
wrapper_f = ; Function Attrs: returns_twice
define double @julia_foo_5929_inner.1({}** %0) #4 !dbg !4 {
entry:
  %1 = alloca {}**, align 8, !dbg !31
  store {}** %0, {}*** %1, align 8, !dbg !31
  %2 = call swiftcc double @julia_foo_5929_inner({}*** %1), !dbg !31
  ret double %2, !dbg !31
}

[...]

PR #912, opened 2023-06-25
GPUCompiler + LLVM 6 adapt

motabbara:
Not sure if all the active(ctx) are needed.


motabbara:
Going to fail all of CI because of CUDA incompatability with GPUCompiler.

src/Enzyme.jl
```diff
@@ -178,7 +178,7 @@ Enzyme.autodiff(ReverseWithPrimal, x->x*x, Active(3.0))
     ModifiedBetween = Val(falses_from_args(Val(1), args...))
 
     tt    = Tuple{map(T->eltype(Core.Typeof(T)), args′)...}
-    world = GPUCompiler.codegen_world_age(Core.Typeof(f.val), tt)
+    world = GPUCompiler.tls_world_age() #(Core.Typeof(f.val), tt).min_world
```
wsmoses, line 181: I don't think this replacement is equivalent here. I think we need to import the codegen_world_age that was dropped here https://github.com/JuliaGPU/GPUCompiler.jl/commit/3de799afb1e145a72aaa30dc7c29222d32802693

wsmoses: Which, relatedly, can we split off the codegen_world_age refit into its own PR (which presumably should add support for both GPUCompiler 0.19 and 0.20), from the other dependency bumps?

motabbara: So PR1 is GPUCompiler bump and codegen_world_age and PR2 is rest of the stuff?

wsmoses: Yeah, [though the rest may also make sense to split off, depending on how invasive the LLVM stuff is]

motabbara:
Will close and submit two PRs. This doesn't resolve #909 as that needs GPUCompiler 0.21.2 which depends on LLVM 6.

PR #915, opened 2023-06-28
Enable later versions of GPUCompiler

motabbara:


src/utils.jl
```diff
@@ -12,3 +12,256 @@ if VERSION <= v"1.6"
...
     import Base: allocatedinline
 end
+
+#Excerpt from https://github.com/JuliaGPU/GPUCompiler.jl/blob/v0.19.4/src/jlgen.jl
+
+using Core: MethodInstance
+# Julia compiler integration
+
+
+## world age lookups
```
vchuravy, line 22: This was removed from GPUCompiler since it can give incorrect answers. Can you add a warning that this is fundamentally unsound.

motabbara: As a comment or when invoked?

Project.toml
```diff
@@ -19,7 +19,7 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 CEnum = "0.4"
 EnzymeCore = "0.3"
 Enzyme_jll = "0.0.70"
-GPUCompiler = "0.19"
+GPUCompiler = "0.19 - 0.21.2"
```
vchuravy, line 22:
```suggestion
GPUCompiler = "0.19, 0.20, 0.21"
```

src/utils.jl
```diff
@@ -12,3 +12,258 @@ if VERSION <= v"1.6"
...
+        # HACK: dealing with a closure or something... let's do somthing really invalid,
+        #       which works because MethodError doesn't actually use the function
+        Ref{ft}()[]
+    end
+end
+function MethodError(ft::Type, tt::Type, world::Integer=typemax(UInt))
+    Base.MethodError(unsafe_function_from_type(ft), tt, world)
+end
+
+"""
```
wsmoses, line 233: Wait this wasn't dropped, was it? We should still use GPUCompiler's version.

src/utils.jl
```diff
@@ -12,3 +12,258 @@ if VERSION <= v"1.6"
...
+# `tls_world_age` should be used to look up the current world age. in most cases, this is
+# what you should use to invoke the compiler with.
+#
+# `codegen_world_age` is a special function that returns the world age in which the passed
+# method instance (identified by its function and argument types) is to be compiled. the
+# returned constant is automatically invalidated when the method is redefined, and as such
+# can be used to drive cached compilation. it is unlikely that you should use this function
+# directly, instead use `cached_compilation` which handles invalidation for you.
+
+tls_world_age() = ccall(:jl_get_tls_world_age, UInt, ())
```
wsmoses, line 35: Same here we shouldn't need this, instead using GPUCompilers, if need be.

src/utils.jl
```diff
@@ -12,3 +12,254 @@ if VERSION <= v"1.6"
...
+"""
+    methodinstance(ft::Type, tt::Type, [world::UInt])
+
+Look up the method instance that corresponds to invoking the function with type `ft` with
+argument typed `tt`. If the `world` argument is specified, the look-up is static and will
+always return the same result. If the `world` argument is not specified, the look-up is
+dynamic and the returned method instance will automatically be invalidated when a relevant
+function is redefined.
+"""
+function methodinstance(ft::Type, tt::Type, world::Integer=tls_world_age())
```
wsmoses, line 238: This looks like its still in GPUCompiler (https://github.com/JuliaGPU/GPUCompiler.jl/blob/c0c9ab7d2604c44dfb90a99332ba3eadad3679d7/src/jlgen.jl#L57) can we remove it from here?

motabbara:
@wsmoses Failing nightly as expected.

PR #917, opened 2023-07-01
enable LLVM6

motabbara:


src/api.jl
```diff
@@ -34,11 +34,11 @@ IntList() = IntList(Ptr{Int64}(0),0)
 
 function EnzymeConcreteTypeIsFloat(cc::CConcreteType, ctx)
   if cc == DT_Half
```
wsmoses, line 36: I assume this shouldn't take a ctx anymore?

wsmoses:
@vchuravy I think you'll need to give this a once over, since the intermixing with being called from a C ABI, and which LLVM ctx will need to be validated to avoid issues.

Unfortunately @vchuravy is out on vacation at present.

src/compiler/utils.jl
```diff
@@ -20,7 +20,11 @@ function get_function!(builderF, mod::LLVM.Module, name)
 end
 
 
-T_ppjlvalue(ctx) = LLVM.PointerType(LLVM.PointerType(LLVM.StructType(LLVMType[]; ctx)))
+function T_ppjlvalue(ctx)
```
wsmoses, line 23: I think this makes sense to not take a context, and use the surrounding ctx

motabbara: I've moved the ctx wrapper to declare_pgcstack! otherwise there's no active context to T_ppjlvalue.

luraess:
Running the following MWE using Enzyme from this PR (incl. LLVM 6 bump) on latest `AMDGPU#pxl-th` and `CUDA@v4.4.0` results in an error: `GPUCompiler.jl encountered an unexpected internal error` (see full trace here [err.txt](https://github.com/EnzymeAD/Enzyme.jl/files/11950210/err.txt))

```julia
using Enzyme
using CUDA

function fun_cpu!(A, B, a)
    for ix ∈ axes(A, 1)
        A[ix] += a * B[ix] * 100.65
    end
    return
end

function fun_gpu!(A, B, a)
    ix = (blockIdx().x - 1) * blockDim().x + threadIdx().x
    A[ix] += a * B[ix] * 100.65
    return
end

function ∇_fun_cpu!(A, Ā, B, B̄, a)
    Enzyme.autodiff_deferred(Enzyme.Reverse, fun_cpu!, Duplicated(A, Ā), Duplicated(B, B̄), Const(a))
    return
end

function ∇_fun_gpu!(A_d, Ā_d, B_d, B̄_d, a)
    Enzyme.autodiff_deferred(Enzyme.Reverse, fun_gpu!, Duplicated(A_d, Ā_d), Duplicated(B_d, B̄_d), Const(a))
    return
end

function main()
    N = 16

    A = rand(Float64, N)
    B = rand(Float64, N)
    a = 6.5

    A_d = CuArray(copy(A))
    B_d = CuArray(copy(B))

    Ā   = ones(Float64, size(A))
    B̄   = ones(Float64, size(B))
    Ā_d = CUDA.ones(Float64, size(A_d))
    B̄_d = CUDA.ones(Float64, size(B_d))

    @info "running on CPU"
    ∇_fun_cpu!(A, Ā, B, B̄, a)

    @info "running on CUDA GPU"
    CUDA.@sync @cuda threads=N blocks=1 ∇_fun_gpu!(A_d, Ā_d, B_d, B̄_d, a)

    return
end

main()
```

wsmoses:
I think the proper way to resolve this and other things is to only (and every time) do the context scope for all Julia code we create an @ cfunction from, when we GPUCompiler.codegen, and whenever a context is created

All other code should not take a context and/or set it.

Thoughts @motabbara ?

motabbara:
Agree @wsmoses. Will attempt something in that direction. 

Specifically for the example by @luraess, I don't have a CUDA machine to repro but perhaps I can trigger with Metal.

motabbara:
At first glance, the wrong context deactivation comes about from the exception thrown from GPUCompiler/lower_byval but I think the original cause of the exception might be unrelated to contexts.

@luraess, do you see the same thing if you set the return activity to Const in autodiff_deferred rather than use the version that guesses activity?

luraess:
> Specifically for the example by @luraess, I don't have a CUDA machine to repro but perhaps I can trigger with Metal.

For now, this Enzyme PR branch is incompatible with Metal v0.5.0 because of `ObjectFile` compat (0.3 here vs 0.4 Metal).

Also, until recently, Enzyme was failing to auto diff Metal kernels (this may have changed though).

wsmoses:
I added objectfile 0.4 recently so hopefully that doesn’t block.

luraess:
> @luraess, do you see the same thing if you set the return activity to Const in autodiff_deferred rather than use the version that guesses activity?

Yes, it also fails with similar error, although the stack trace may look slightly different (see [err.txt](https://github.com/EnzymeAD/Enzyme.jl/files/11954644/err.txt))

luraess:
> I added objectfile 0.4 recently so hopefully that doesn’t block.
> […](#)

I guess this blocks for Metal.jl which v0.5.0 is still on `ObjectFIle = 0.3`

motabbara:
@wsmoses Can we merge this in that failing CI is now unrelated to the PR as far as I can tell. There is a context mismatch warning that I will find in a follow-up.

wsmoses:
this seems a bit both invasive as well as prone to error in all of the sub handlers. 

Simultaneously, I've been working on a code cleanup/simplification that puts various ABI calls in a utility class, as well as making a macro for all of the handlers.

Perhaps we could land this: https://github.com/EnzymeAD/Enzyme.jl/pull/927/files first, and then inside of the three macros augfunc, fwdfunc and revfunc, we add the context! setting, simplifying much of this change, at least in compiler.jl?

wsmoses:
@motabbara landed the other one

src/compiler.jl
```diff
@@ -439,13 +439,13 @@ macro safe_show(exs...)
 end
 
 declare_allocobj!(mod) = get_function!(mod, "julia.gc_alloc_obj") do ctx
```
wsmoses, line 441: We should remove the context arg from get_function

motabbara:
@wsmoses

GC/No decay test is disabled

src/compiler.jl
5799:         gutils = API.EnzymeGradientUtilsRef(data). ###data is void here
5800:         newb = new_from_original(gutils, val)
5801:         while isa(newb, LLVM.PHIInst)
5802:             newb = LLVM.Instruction(LLVM.API.LLVMGetNextInstruction(newb))
5803:         end

I can repro the issue reported by @luraess . Metal/CUDA creates its own JuliaContext in compile but haven't worked out how to use that.

wsmoses:
I think it's probably being passed in as the ctx argument to codegen presumably?

Cc @vchuravy

wsmoses:
re the data is void there, do you have a reproducer outside of the context stuff? Because looking at the other end [eg the code that calls it], that should never happen...

src/compiler.jl
```diff
@@ -896,54 +900,55 @@ function array_shadow_handler(B::LLVM.API.LLVMBuilderRef, OrigCI::LLVM.API.LLVMV
     b = LLVM.IRBuilder(B)
     orig = LLVM.Value(OrigCI)
 
-    vals = LLVM.Value[]
```
wsmoses, line 899: Can we restore the indentation here? It'll both look prettier and also reduce the size of the diff

src/compiler.jl
```diff
@@ -1610,26 +1614,26 @@ function common_generic_fwd(offset, B, orig, gutils, normalR, shadowR)
...
-            gep = LLVM.inbounds_gep!(B, AT, sret, [LLVM.ConstantInt(0; ctx), LLVM.ConstantInt(1; ctx)])
-            shadow = LLVM.load!(B, T_prjlvalue, gep)
-        else
-            ST = LLVM.LLVMType(API.EnzymeGetShadowType(width, value_type(orig)))
-            shadow = LLVM.UndefValue(ST)
-            for i in 1:width
-                gep = LLVM.inbounds_gep!(B, AT, sret, [LLVM.ConstantInt(0; ctx), LLVM.ConstantInt(i; ctx)])
-                ld = LLVM.load!(B, T_prjlvalue, gep)
-                shadow = insert_value!(B, shadow, ld, i-1)
+        sret = generic_setup(orig, runtime_generic_fwd, AnyArray(1+Int(width)), gutils, #=start=#offset, ctx, B, false)
```
wsmoses, line 1617: and heere

src/compiler.jl
```diff
@@ -1658,28 +1661,28 @@ function common_generic_augfwd(offset, B, orig, gutils, normalR, shadowR, tapeR)
...
-            gep = LLVM.inbounds_gep!(B, AT, sret, [LLVM.ConstantInt(0; ctx), LLVM.ConstantInt(1; ctx)])
-            shadow = LLVM.load!(B, T_prjlvalue, gep)
-        else
-            ST = LLVM.LLVMType(API.EnzymeGetShadowType(width, value_type(orig)))
-            shadow = LLVM.UndefValue(ST)
-            for i in 1:width
-                gep = LLVM.inbounds_gep!(B, AT, sret, [LLVM.ConstantInt(0; ctx), LLVM.ConstantInt(i; ctx)])
-                ld = LLVM.load!(B, T_prjlvalue, gep)
-                shadow = insert_value!(B, shadow, ld, i-1)
+        if shadowR != C_NULL
```
wsmoses, line 1664: etc

src/compiler.jl
```diff
@@ -1723,31 +1726,31 @@ function common_apply_latest_fwd(offset, B, orig, gutils, normalR, shadowR)
     mod = LLVM.parent(LLVM.parent(LLVM.parent(orig)))
     ctx = LLVM.context(orig)
 
-    T_jlvalue = LLVM.StructType(LLVMType[]; ctx)
+    T_jlvalue = LLVM.StructType(LLVMType[])
     T_prjlvalue = LLVM.PointerType(T_jlvalue, Tracked)
 
     width = get_width(gutils)
     AT = LLVM.ArrayType(T_prjlvalue, 1+Int(width))
     sret = generic_setup(orig, runtime_generic_fwd, AnyArray(1+Int(width)), gutils, #=start=#offset+1, ctx, B, false)
```
wsmoses, line 1734: generic_setup should no longer take a ctx

luraess:
#925 still fails after landing #930 on M1

motabbara:
> re the data is void there, do you have a reproducer outside of the context stuff? Because looking at the other end [eg the code that calls it], that should never happen...

I've enabled the test again and passes CI on everything but nightly.

luraess:
> > re the data is void there, do you have a reproducer outside of the context stuff? Because looking at the other end [eg the code that calls it], that should never happen...
> 
> I've enabled the test again and passes CI on everything but nightly.

Trying the MWE on the latest commit of this branch still fails for me (`ERROR: Deactivating wrong context`).

src/compiler.jl
```diff
@@ -2144,8 +2145,8 @@ function common_jl_getfield_rev(offset, B, orig, gutils, tape)
         inps = [inp]
     end
 
-    vals = LLVM.Value[]
-    push!(vals, inps[1])
+        vals = LLVM.Value[]
```
wsmoses, line 2148: also here re indent

src/compiler.jl
```diff
@@ -2154,15 +2155,15 @@ function common_jl_getfield_rev(offset, B, orig, gutils, tape)
     sym = emit_apply_type!(B, Base.Val, [sym])
     push!(vals, sym)
 
-    push!(vals, unsafe_to_llvm(Val(is_constant_value(gutils, orig)), ctx))
+    push!(vals, unsafe_to_llvm(Val(is_constant_value(gutils, orig))))
 
-    for v in inps[2:end]
-        push!(vals, v)
-    end
+        for v in inps[2:end]
```
wsmoses, line 2160: and here

src/compiler.jl
```diff
@@ -2943,18 +2943,21 @@ end
     # TODO:
     #  - GPU support
     #  - When OrcV2 only use a MaterializationUnit to avoid mutation of the module here
+    #activate(ctx)
```
wsmoses, line 2946: remove?

src/compiler.jl
```diff
@@ -3185,19 +3192,22 @@ end
 else
             jctx = ctxToThreadSafe[jctx]
 end
-            cmod, adjointnm, augfwdnm, _, TapeType = _thunk(ejob, jctx, #=postopt=#false)
+            cmod, adjointnm, augfwdnm, TapeType = context!(jctx) do
```
wsmoses, line 3195: I don't think we should need this since we should preumably already be in a context?

src/compiler.jl
```diff
@@ -3185,19 +3192,22 @@ end
 else
             jctx = ctxToThreadSafe[jctx]
```
wsmoses, line 3193: we can presumably get rid of the jctx stuff here

src/compiler.jl
```diff
@@ -3387,13 +3397,13 @@ function threadsfor_augfwd(B, orig, gutils, normalR, shadowR, tapeR)
     normal = (unsafe_load(normalR) != C_NULL) ? LLVM.Instruction(unsafe_load(normalR)) : nothing
     shadow = (unsafe_load(shadowR) != C_NULL) ? LLVM.Instruction(unsafe_load(shadowR)) : nothing
 
-    byRef, sname, dfuncT, vals, thunkTy, _, copies = threadsfor_common(orig, gutils, B, API.DEM_ReverseModePrimal)
+        byRef, sname, dfuncT, vals, thunkTy, _, copies = threadsfor_common(orig, gutils, B, API.DEM_ReverseModePrimal)
```
wsmoses, line 3400: Indent

src/compiler.jl
```diff
@@ -3387,13 +3397,13 @@ function threadsfor_augfwd(B, orig, gutils, normalR, shadowR, tapeR)
...
 
-@static if VERSION < v"1.8-"
-    if byRef
-        emit_error(B, orig, "Enzyme: active variable in Threads.@threads closure "*(string(eltype(eltype(dfuncT))))*" not supported")
+    @static if VERSION < v"1.8-"
```
wsmoses, line 3402: Indent

src/compiler.jl
```diff
@@ -3403,21 +3413,21 @@ end
     mode = get_mode(gutils)
     world = enzyme_extract_world(LLVM.parent(position(B)))
     entry = nested_codegen!(mode, mod, runtime_pfor_augfwd, tt, world)
-    push!(function_attributes(entry), EnumAttribute("alwaysinline"; ctx))
+    push!(function_attributes(entry), EnumAttribute("alwaysinline"))
 
-    pval = const_ptrtoint(functions(mod)[sname], convert(LLVMType, Ptr{Cvoid}; ctx))
-    pval = LLVM.ConstantArray(value_type(pval), [pval])
-    store!(B, pval, vals[1])
+        pval = const_ptrtoint(functions(mod)[sname], convert(LLVMType, Ptr{Cvoid}))
```
wsmoses, line 3418: Indent

src/compiler.jl
```diff
@@ -3403,21 +3413,21 @@ end
...
+        store!(B, pval, vals[1])
 
     tape = LLVM.call!(B, LLVM.function_type(entry), entry, vals)
     debug_from_orig!(gutils, tape, orig)
 
-    if !any_jltypes(get_tape_type(thunkTy))
-        if value_type(tape) != convert(LLVMType, Ptr{Cvoid}; ctx)
-            tape = LLVM.ConstantInt(0; ctx)
-            GPUCompiler.@safe_warn "Illegal calling convention for threadsfor augfwd"
+        if !any_jltypes(get_tape_type(thunkTy))
```
wsmoses, line 3425: Indent

src/compiler.jl
```diff
@@ -3455,11 +3465,11 @@ else
 end
     mode = get_mode(gutils)
     entry = nested_codegen!(mode, mod, runtime_pfor_rev, tt, world)
-    push!(function_attributes(entry), EnumAttribute("alwaysinline"; ctx))
+    push!(function_attributes(entry), EnumAttribute("alwaysinline"))
 
-    pval = const_ptrtoint(functions(mod)[sname], convert(LLVMType, Ptr{Cvoid}; ctx))
-    pval = LLVM.ConstantArray(value_type(pval), [pval])
-    store!(B, pval, vals[1])
+        pval = const_ptrtoint(functions(mod)[sname], convert(LLVMType, Ptr{Cvoid}))
```
wsmoses, line 3470: Indent

src/compiler.jl
```diff
@@ -3468,6 +3478,7 @@ end
         ld = load!(B, pllty, val)
         store!(B, ld, pv)
     end
+
```
wsmoses, line 3481: extra newline

src/compiler.jl
```diff
@@ -3478,6 +3489,7 @@ function newtask_fwd(B, orig, gutils, normalR, shadowR)
         return true
     end
     ctx = LLVM.context(orig)
+
```
wsmoses, line 3492: extra newline

src/compiler.jl
```diff
@@ -3478,6 +3489,7 @@ function newtask_fwd(B, orig, gutils, normalR, shadowR)
         return true
     end
     ctx = LLVM.context(orig)
```
wsmoses, line 3491: can we get rid of this ctx var?

src/compiler.jl
```diff
@@ -3524,9 +3536,9 @@ function newtask_augfwd(B, orig, gutils, normalR, shadowR, tapeR)
     shadow = (unsafe_load(shadowR) != C_NULL) ? LLVM.Instruction(unsafe_load(shadowR)) : nothing
     mod = LLVM.parent(LLVM.parent(LLVM.parent(orig)))
     ctx = LLVM.context(orig)
-
-    T_jlvalue = LLVM.StructType(LLVMType[]; ctx)
-    T_prjlvalue = LLVM.PointerType(T_jlvalue, Tracked)
+    
+        T_jlvalue = LLVM.StructType(LLVMType[])
```
wsmoses, line 3540: indent

src/compiler.jl
```diff
@@ -3488,19 +3500,19 @@ function newtask_fwd(B, orig, gutils, normalR, shadowR)
...
                        (sizeof(Int) == sizeof(Int64) ? emit_box_int64! : emit_box_int32!)(B, new_from_original(gutils, ops[3])),
-                       unsafe_to_llvm(Val(width), ctx),
+                       unsafe_to_llvm(Val(width)),
                       ]
 
     ntask = emit_apply_generic!(B, vals)
     debug_from_orig!(gutils, ntask, orig)
 
-    # TODO: GC, ret
+        # TODO: GC, ret
```
wsmoses, line 3515: indent

src/compiler.jl
```diff
@@ -3759,9 +3771,9 @@ function enzyme_custom_setup_args(B, orig, gutils, mi, RT, reverse, isKWCall)
     sret = sret !== nothing
     returnRoots = returnRoots !== nothing
 
-	jlargs = classify_arguments(mi.specTypes, called_type(orig), sret, returnRoots, parmsRemoved)
+    jlargs = classify_arguments(mi.specTypes, called_type(orig), sret, returnRoots, parmsRemoved)
```
wsmoses, line 3774: indent

src/compiler.jl
```diff
@@ -3941,6 +3953,7 @@ function enzyme_custom_setup_ret(gutils, orig, mi, RealRt)
     origNeedsPrimal = needsPrimal
 
     ctx = LLVM.context(orig)
+
```
wsmoses, line 3956: remove newline (also the ctx var?)

src/compiler.jl
```diff
@@ -3983,91 +3996,91 @@ function enzyme_custom_fwd(B, orig, gutils, normalR, shadowR)
 
     width = get_width(gutils)
 
-    if shadowR != C_NULL
-        unsafe_store!(shadowR,UndefValue(LLVM.LLVMType(API.EnzymeGetShadowType(width, value_type(orig)))).ref)
-    end
+        if shadowR != C_NULL
```
wsmoses, line 3999: indent

src/compiler.jl
```diff
@@ -3983,91 +3996,91 @@ function enzyme_custom_fwd(B, orig, gutils, normalR, shadowR)
...
-        @assert kwtup !== nothing
-        insert!(tt, 1, kwtup)
-        insert!(tt, 2, Core.typeof(EnzymeRules.forward))
-        insert!(tt, 4, Type{RT})
-    else
-        @assert kwtup === nothing
-        insert!(tt, 2, Type{RT})
-    end
-    TT = Tuple{tt...}
+        tt = copy(activity)
```
wsmoses, line 4024: indent

src/compiler.jl
```diff
@@ -3983,91 +3996,91 @@ function enzyme_custom_fwd(B, orig, gutils, normalR, shadowR)
...
-            @safe_debug "Applying custom forward rule (kwcall)" TT
-            llvmf = nested_codegen!(mode, mod, kwfunc, TT, world)
-            fwd_RT = Core.Compiler.return_type(kwfunc, TT, world)
-        end
-    else
-        if EnzymeRules.isapplicable(EnzymeRules.forward, TT; world)
-            @safe_debug "Applying custom forward rule" TT
-            llvmf = nested_codegen!(mode, mod, EnzymeRules.forward, TT, world)
-            fwd_RT = Core.Compiler.return_type(EnzymeRules.forward, TT, world)
+        # TODO get world
```
wsmoses, line 4043: indent

src/compiler.jl
```diff
@@ -3983,91 +3996,91 @@ function enzyme_custom_fwd(B, orig, gutils, normalR, shadowR)
...
-    _, sret, returnRoots = get_return_info(enzyme_custom_extract_mi(llvmf)[2], ctx)
-    if sret !== nothing
-        sret = alloca!(alloctx, convert(LLVMType, eltype(sret); ctx))
-        pushfirst!(args, sret)
-        if returnRoots !== nothing
-            returnRoots = alloca!(alloctx, convert(LLVMType, eltype(returnRoots); ctx))
-            insert!(args, 2, returnRoots)
+    push!(function_attributes(llvmf), EnumAttribute("alwaysinline", 0))
+
+        _, sret, returnRoots = get_return_info(enzyme_custom_extract_mi(llvmf)[2], ctx)
```
wsmoses, line 4071: indent

src/compiler.jl
```diff
@@ -4086,24 +4099,24 @@ function enzyme_custom_fwd(B, orig, gutils, normalR, shadowR)
     res = LLVM.call!(B, LLVM.function_type(llvmf), llvmf, args)
     debug_from_orig!(gutils, res, orig)
 
-    hasNoRet = any(map(k->kind(k)==kind(EnumAttribute("noreturn"; ctx)), collect(function_attributes(llvmf))))
+        hasNoRet = any(map(k->kind(k)==kind(EnumAttribute("noreturn")), collect(function_attributes(llvmf))))
```
wsmoses, line 4102: indent

src/compiler.jl
```diff
@@ -4086,24 +4099,24 @@ function enzyme_custom_fwd(B, orig, gutils, normalR, shadowR)
...
     if hasNoRet
         return false
     end
 
-    if sret !== nothing
-        if LLVM.version().major >= 12
-            attr = TypeAttribute("sret", eltype(value_type(parameters(llvmf)[1])); ctx)
-        else
-            attr = EnumAttribute("sret"; ctx)
+        if sret !== nothing
```
wsmoses, line 4108: Indent

src/compiler.jl
```diff
@@ -4086,24 +4099,24 @@ function enzyme_custom_fwd(B, orig, gutils, normalR, shadowR)
...
+            LLVM.API.LLVMAddCallSiteAttribute(res, LLVM.API.LLVMAttributeIndex(1), attr)
+            res = load!(B, eltype(value_type(parameters(llvmf)[1])), sret)
         end
-        LLVM.API.LLVMAddCallSiteAttribute(res, LLVM.API.LLVMAttributeIndex(1), attr)
-        res = load!(B, eltype(value_type(parameters(llvmf)[1])), sret)
-    end
 
-    shadowV = C_NULL
-    normalV = C_NULL
+        shadowV = C_NULL
```
wsmoses, line 4118: Indent

src/compiler.jl
```diff
@@ -4172,16 +4185,16 @@ function enzyme_custom_fwd(B, orig, gutils, normalR, shadowR)
         end
     end
 
-    if shadowR != C_NULL
-        unsafe_store!(shadowR, shadowV)
-    end
+        if shadowR != C_NULL
```
wsmoses, line 4188: indent

src/compiler.jl
```diff
@@ -4367,12 +4380,12 @@ function enzyme_custom_common_rev(forward::Bool, B, orig::LLVM.CallInst, gutils,
         end
         if RT <: Active
 
-            llty = convert(LLVMType, RT; ctx)
+                llty = convert(LLVMType, RT)
```
wsmoses, line 4383: indent

src/compiler.jl
```diff
@@ -4383,27 +4396,27 @@ function enzyme_custom_common_rev(forward::Bool, B, orig::LLVM.CallInst, gutils,
                 end
             end
 
-            al0 = al = emit_allocobj!(B, RT)
-            al = bitcast!(B, al, LLVM.PointerType(llty, addrspace(value_type(al))))
-            al = addrspacecast!(B, al, LLVM.PointerType(llty, Derived))
+                al0 = al = emit_allocobj!(B, RT)
```
wsmoses, line 4399: indent

src/compiler.jl
```diff
@@ -4458,8 +4470,8 @@ function enzyme_custom_common_rev(forward::Bool, B, orig::LLVM.CallInst, gutils,
         ST = EnzymeRules.AugmentedReturn{needsPrimal ? RealRt : Nothing, needsShadow ? ShadT : Nothing, TapeT}
         if aug_RT != ST
             if aug_RT <: EnzymeRules.AugmentedReturnFlexShadow
-                if convert(LLVMType, EnzymeRules.shadow_type(aug_RT); ctx, allow_boxed=true) !=
-                   convert(LLVMType, EnzymeRules.shadow_type(ST)    ; ctx, allow_boxed=true)
+                if convert(LLVMType, EnzymeRules.shadow_type(aug_RT); allow_boxed=true) !=
+                convert(LLVMType, EnzymeRules.shadow_type(ST)    ; allow_boxed=true)
```
wsmoses, line 4474: ok stopping indent stuff now, but hopefully is clear to continue that

src/compiler.jl
```diff
@@ -6104,7 +6117,8 @@ function julia_undef_value_for_type(Ty::LLVM.API.LLVMTypeRef, forceZero::UInt8):
         end
     end
     if isa(ty, LLVM.PointerType)
-        val = unsafe_to_llvm(nothing, LLVM.context(ty))
+        #LLVM.context(ty)
```
wsmoses, line 6120: remove?

src/compiler.jl
```diff
@@ -6172,123 +6187,122 @@ function zero_allocation(B, LLVMType, obj, isTape::UInt8)
...
-
 end
 
 
 function zero_allocation(B::LLVM.IRBuilder, jlType, LLVMType, obj, AlignedSize, Size, zeroAll::Bool)::LLVM.API.LLVMValueRef
     func = LLVM.parent(position(B))
     mod = LLVM.parent(func)
     ctx = context(mod)
-    T_int8 = LLVM.Int8Type(ctx)
+        T_int8 = LLVM.Int8Type()
```
wsmoses, line 6264: like here

src/compiler.jl
```diff
@@ -7220,6 +7256,7 @@ function enzyme!(job, mod, primalf, TT, mode, width, parallel, actualRetType, wr
     args_known_values = API.IntList[]
 
     ctx = LLVM.context(mod)
+    #activate(ctx)
```
wsmoses, line 7259: remove

src/compiler.jl
```diff
@@ -8552,7 +8593,24 @@ function GPUCompiler.codegen(output::Symbol, job::CompilerJob{<:EnzymeTarget};
         primal_job = CompilerJob(primal, config2, job.world) # TODO EnzymeInterp params, etc
     end
 
-    mod, meta = GPUCompiler.codegen(:llvm, primal_job; optimize=false, toplevel=toplevel, cleanup=false, validate=false, parent_job=parent_job, ctx)
+    ctx = nothing
+    c = nothing
+
+    @static if VERSION < v"1.9-"
```
wsmoses, line 8599: Why do we need this? Re below can we not get the corresponding context from mod?

E.g. ctx = LLVM.context(mod)

src/compiler.jl
```diff
@@ -9016,7 +9077,9 @@ end
     end
 
     adjointf = functions(mod)[adjointf_name]
-    push!(function_attributes(adjointf), EnumAttribute("alwaysinline", 0; ctx=context(mod)))
+    ctx=context(mod)
```
wsmoses, line 9080: remove ctx?

src/compiler.jl
```diff
@@ -559,7 +563,7 @@ function emit_jl!(B, val)::LLVM.Value
     fn = LLVM.parent(curent_bb)
     mod = LLVM.parent(fn)
     ctx = LLVM.context(mod)
```
wsmoses, line 565: Can we get rid of all of these ctx = LLVM.context?

src/compiler.jl
```diff
@@ -9231,8 +9296,9 @@ end
     end
 
     ctx = LLVM.Context()
```
wsmoses, line 9298: This needs to be activated, no?

src/compiler.jl
```diff
@@ -9416,18 +9482,18 @@ end
...
 
     # Now invoke the JIT
     jitted_mod = JIT.add!(mod)
-    if VERSION >= v"1.9.0-DEV.115"
-        LLVM.dispose(ctx)
-    else
-        # we cannot dispose of the global unique context
-    end
+    #if VERSION >= v"1.9.0-DEV.115"
+    #    LLVM.dispose(ctx)
```
wsmoses, line 9493: I think we still need to dispose this, cc @vchuravy

src/compiler.jl
```diff
@@ -9448,33 +9514,47 @@ function _link(job, (mod, adjoint_name, primal_name, ctx, TapeType))
...
+        end
+        return (mod, adjoint_name, primal_name, meta.TapeType)
     end
- 
-    LLVM.ModulePassManager() do pm
-        add!(pm, FunctionPass("ReinsertGCMarker", reinsert_gcmarker_pass!))
-        run!(pm, mod)
+
+@static if VERSION < v"1.9.0"
+    JuliaContext() do ctx
```
wsmoses, line 9545: Why don't we just make the context at the code that calls _thunk without it (e.g. from the generators)?

src/compiler/orcv2.jl
```diff
@@ -128,9 +129,15 @@ function move_to_threadsafe(ir)
     buf = convert(MemoryBuffer, ir)
 
     # 2. deserialize and wrap by a ThreadSafeModule
-    return ThreadSafeContext() do ctx
-        mod = parse(LLVM.Module, buf; ctx=context(ctx))
-        ThreadSafeModule(mod; ctx)
+    ts_ctx = ThreadSafeContext()
+    ctx = context(ts_ctx)
+    activate(ctx)
+    try
```
wsmoses, line 135: @vchuravy @maleadt  is there a nicer way to do this?

src/compiler/utils.jl
```diff
@@ -113,7 +115,8 @@ end
 function reinsert_gcmarker!(func, PB=nothing)
 	pgs = get_pgcstack(func)
     if pgs === nothing
-        B = IRBuilder(context(LLVM.parent(func)))
+        context(LLVM.parent(func))
```
wsmoses, line 118: We should instead ensure that the code that calls this has the context set.

wsmoses: Oh just kidding this is dead code, just remove.

src/compiler/validation.jl
```diff
@@ -258,33 +257,30 @@ function check_ir!(job, errors, imported, inst::LLVM.CallInst, calls)
...
             if mfn == C_NULL
-                ptr8 = LLVM.PointerType(LLVM.IntType(8; ctx))
+                ptr8 = LLVM.PointerType(LLVM.IntType(8))
                 mfn = LLVM.API.LLVMAddFunction(mod, "malloc", LLVM.FunctionType(ptr8, [value_type(LLVM.Value(LLVM.LLVM.API.LLVMGetOperand(inst, 0)))]))
             end
             mfn2 = LLVM.Function(mfn)
             nval = ptrtoint!(b, call!(b, LLVM.function_type(mfn2), mfn2, [LLVM.Value(LLVM.LLVM.API.LLVMGetOperand(inst, 0))]), value_type(inst))
             replace_uses!(inst, nval)
             LLVM.API.LLVMInstructionEraseFromParent(inst)
+
```
wsmoses, line 272: extra line?

src/compiler/validation.jl
```diff
@@ -308,169 +304,169 @@ function check_ir!(job, errors, imported, inst::LLVM.CallInst, calls)
...
         elseif fn == "jl_lazy_load_and_lookup" || fn == "ijl_lazy_load_and_lookup"
             ofn = LLVM.parent(LLVM.parent(inst))
             mod = LLVM.parent(ofn)
-            ctx = context(mod)
 
-            ops = collect(operands(inst))[1:end-1]
-            @assert length(ops) == 2
-            flib = ops[1]
-            fname = ops[2]
+                ops = collect(operands(inst))[1:end-1]
```
wsmoses, line 312: mode indent stuff

src/utils.jl
```diff
@@ -5,24 +5,31 @@ const Tracked = 10
...
+    context(ctx::LLVM.Context) = ctx
+else
+    context!(f, ctx::LLVM.ThreadSafeContext) = LLVM.ts_context!(f, ctx)
 end
 
-export makeInstanceOf
+@static if VERSION < v"1.9-"
+    has_context() = LLVM._has_context()
+else
+    has_context() = LLVM._has_ts_context() ###??
```
wsmoses, line 31: @vchuravy @maleadt 

separately it would be nice if we can refactor to not need this

test/Manifest.toml
```diff
@@ -0,0 +1,521 @@
+# This file is machine-generated - editing it directly is not advised
+
```
wsmoses, line 2: This file should not be committed

test/metal.jl
```diff
@@ -0,0 +1,52 @@
+using CUDA
```
wsmoses, line 1: Why is this "using CUDA" not "using Metal"

test/runtests.jl
```diff
@@ -264,12 +264,6 @@ euroad′(x) = first(autodiff(Reverse, euroad, Active, Active(x)))[1]
 test_scalar(euroad, 0.5)
 end
 
-@testset "Nested AD" begin
-    tonest(x,y) = (x + y)^2
```
wsmoses, line 268: can we not move this?

test/runtests.jl
```diff
@@ -1271,6 +1271,7 @@ typeunknownvec = Float64[]
 end
 
 @testset "No Decayed / GC" begin
+
```
wsmoses, line 1274: remove the extra line

wsmoses:
@vchuravy I'm not seeing buildkite results to confirm cuda tests were run

src/compiler.jl
motabbara: if this matches the Context() in enzyme_call then it's now disposed on exit of enzyme_call

src/compiler/orcv2.jl
motabbara: On Julia < 1.9, JuliaContext() returns a Context() rather than a ThreadSafeContext() so ThreadSafeModule fails. The code snippet here is the same as that used in GPUCompiler for the >= 1.9 case.

src/compiler.jl
motabbara: I tried several permutations of this and always break with a missing ThreadSafeContext() in orcv2/get_trampoline.

src/compiler.jl
```diff
@@ -990,27 +987,27 @@ end
...
     array = LLVM.pointercast!(B, array, LLVM.PointerType(ST, LLVM.addrspace(LLVM.value_type(array))))
-    v = inbounds_gep!(B, ST, array, LLVM.Value[LLVM.ConstantInt(Int32(0); ctx), LLVM.ConstantInt(Int32(1); ctx)])
-    sizeT = LLVM.IntType(8*sizeof(Csize_t); ctx)
+    v = inbounds_gep!(B, ST, array, LLVM.Value[LLVM.ConstantInt(Int32(0)), LLVM.ConstantInt(Int32(1))])
+    sizeT = LLVM.IntType(8*sizeof(Csize_t))
     return LLVM.load!(B, sizeT, v)
 end
 
 function get_array_nrows(B, array)
     ctx = LLVM.context(array)
```
wsmoses, line 1006: can you remove the other ctx's that aren't used, like here? Just search for `context(` and that should probably find most of them

src/compiler.jl
```diff
@@ -3873,8 +3865,8 @@ function enzyme_custom_setup_args(B, orig, gutils, mi, RT, reverse, isKWCall)
 
             push!(args, al)
 
-            push!(activity, Ty)
-            push!(actives, op)
+        push!(activity, Ty)
```
wsmoses, line 3868: indentation?

src/compiler.jl
```diff
@@ -5962,12 +5951,12 @@ function tape_type(LLVMType::LLVM.LLVMType)
     return TT
 end
 
-from_tape_type(::Type{T}, ctx) where T<:AbstractFloat = convert(LLVMType, T; ctx)
-from_tape_type(::Type{T}, ctx) where T<:Integer = convert(LLVMType, T; ctx)
+from_tape_type(::Type{T}, ctx) where T<:AbstractFloat = convert(LLVMType, T)
```
wsmoses, line 5954: these shouldn't take a ctx

src/compiler.jl
```diff
@@ -7467,8 +7469,8 @@ function enzyme!(job, mod, primalf, TT, mode, width, parallel, actualRetType, wr
             uncacheable_args))
         augmented_primalf = nothing
         if wrap
-          pf = adjointf
-          adjointf = create_abi_wrapper(adjointf, TT, rt, actualRetType, API.DEM_ForwardMode, nothing, width, returnUsed, shadow_init, world, interp)
+        pf = adjointf
```
wsmoses, line 7472: indent?

src/compiler.jl
```diff
@@ -7313,38 +7315,38 @@ function enzyme!(job, mod, primalf, TT, mode, width, parallel, actualRetType, wr
 
     rules = Dict{String, API.CustomRuleType}(
         "jl_apply_generic" => @cfunction(ptr_rule,
-                                           UInt8, (Cint, API.CTypeTreeRef, Ptr{API.CTypeTreeRef},
-                                                   Ptr{API.IntList}, Csize_t, LLVM.API.LLVMValueRef)),
+                                        UInt8, (Cint, API.CTypeTreeRef, Ptr{API.CTypeTreeRef},
```
wsmoses, line 7318: indent here and down

src/compiler.jl
```diff
@@ -8387,18 +8388,18 @@ function lower_convention(functy::Type, mod::LLVM.Module, entry_f::LLVM.Function
...
                 else
                     push!(nops, parm)
                 end
             end
             res = call!(builder, LLVM.function_type(wrapper_f), wrapper_f, nops)
             callconv!(res, callconv(wrapper_f))
             if sret
-              @assert value_type(res) == eltype(value_type(ops[1]))
-              store!(builder, res, ops[1])
+            @assert value_type(res) == eltype(value_type(ops[1]))
```
wsmoses, line 8399: indent

src/compiler/pmap.jl
```diff
@@ -238,36 +237,36 @@ end
         shadow = if is_constant_value(gutils, ops[i]) == 0
           invert_pointer(gutils, ops[i], B)
         else
-          nothing
+        nothing
```
wsmoses, line 240: indent?

src/compiler/pmap.jl
```diff
@@ -238,36 +237,36 @@ end
...
         end
 
         codegen_typ = value_type(parameters(entry)[length(vals)+1])
 
         if codegen_typ == value_type(primal)
             push!(vals, primal)
             if shadow !== nothing
-              push!(vals, shadow)
+            push!(vals, shadow)
```
wsmoses, line 248: indent?

src/compiler/pmap.jl
```diff
@@ -238,36 +237,36 @@ end
...
             end
         elseif codegen_typ isa LLVM.PointerType && issized(eltype(codegen_typ)) &&
-               !(source_typ <: Ptr) && !(source_typ <: Core.LLVMPtr)
+            !(source_typ <: Ptr) && !(source_typ <: Core.LLVMPtr)
             if !GPUCompiler.deserves_argbox(source_typ)
-              primA = alloca!(EB, value_type(primal))
-              store!(B, primal, primA)
-              primal = addrspacecast!(B, primA, codegen_typ)
+            primA = alloca!(EB, value_type(primal))
```
wsmoses, line 253: indent?

src/compiler/pmap.jl
```diff
@@ -238,36 +237,36 @@ end
...
+            store!(B, primal, primA)
+            primal = addrspacecast!(B, primA, codegen_typ)
             end
             push!(vals, primal)
             if shadow !== nothing
-              if !GPUCompiler.deserves_argbox(source_typ) 
+            if !GPUCompiler.deserves_argbox(source_typ)
```
wsmoses, line 259: indent?

src/compiler/pmap.jl
```diff
@@ -238,36 +237,36 @@ end
...
             if shadow !== nothing
-              if !GPUCompiler.deserves_argbox(source_typ) 
+            if !GPUCompiler.deserves_argbox(source_typ) 
                 shadowA = alloca!(EB, value_type(shadow))
                 store!(B, shadow, shadowA)
                 shadow = addrspacecast!(B, shadowA, codegen_typ)
-              end
-              push!(vals, shadow)
+            end
+            push!(vals, shadow)
```
wsmoses, line 264: etc

src/compiler/pmap.jl
```diff
@@ -238,36 +237,36 @@ end
...
             end
             # push!(args, (cc=GPUCompiler.BITS_REF, typ=source_typ,
             #             codegen=(typ=codegen_typ, i=codegen_i)))
         else
-			@assert false
+            @assert false
```
wsmoses, line 269: etc

test/Manifest.toml
wsmoses: bump, please remove

src/compiler.jl
```diff
@@ -500,23 +499,22 @@ function emit_allocobj!(B, T::DataType)
     curent_bb = position(B)
     fn = LLVM.parent(curent_bb)
     mod = LLVM.parent(fn)
-	ctx = context(mod)
 
-	T_jlvalue = LLVM.StructType(LLVMType[]; ctx)
+	T_jlvalue = LLVM.StructType(LLVMType[])
```
wsmoses, line 503: indent

src/compiler.jl
```diff
@@ -2720,23 +2714,23 @@ function common_invoke_fwd(offset, B, orig, gutils, normalR, shadowR)
     normal = (unsafe_load(normalR) != C_NULL) ? LLVM.Instruction(unsafe_load(normalR)) : nothing
     shadow = (unsafe_load(shadowR) != C_NULL) ? LLVM.Instruction(unsafe_load(shadowR)) : nothing
     ctx = LLVM.context(orig)
```
wsmoses, line 2716: eg removing context here?

src/compiler.jl
```diff
@@ -3132,11 +3125,14 @@ end
 else
             jctx = ctxToThreadSafe[jctx]
 end
-            cmod, fwdmodenm, _, _ = _thunk(ejob, jctx, #=postopt=#false)
+            cmod, fwdmodenm, _, _ = context!(jctx) do
```
wsmoses, line 3128: I don't think this should be needed since the calling code should already be in a contet, from fwdfunc, right?

src/Enzyme.jl
```diff
@@ -176,6 +178,7 @@ Enzyme.autodiff(ReverseWithPrimal, x->x*x, Active(3.0))
     point values, but cannot do so for integer values in tuples and structs.
 """
 @inline function autodiff(::ReverseMode{ReturnPrimal, RABI}, f::FA, ::Type{A}, args...) where {FA<:Annotation, A<:Annotation, ReturnPrimal, RABI<:ABI}
+    JuliaContext() do ctx
```
wsmoses, line 181: This should probably be in the generated function code, not here.

motabbara: It looks like there are multiples calls to these in autodiff* top-level functions and I end up having to spin up some contexts to allow the returned thunks to be called. This way also results in fewer calls to activate/deactivate contexts that reach into task-local storage.

src/compiler.jl
```diff
@@ -357,17 +357,21 @@ end
 end
 
 @inline @generated function active_reg(::Type{T}) where {T}
+    JuliaContext() do ctx
```
vchuravy, line 360: Why do we need JuliaContext here?
Furthermore alignment.

src/compiler.jl
```diff
@@ -450,13 +454,12 @@ function emit_allocobj!(B, tag::LLVM.Value, Size::LLVM.Value, needs_workaround::
     curent_bb = position(B)
     fn = LLVM.parent(curent_bb)
     mod = LLVM.parent(fn)
-	ctx = context(mod)
 
-	T_jlvalue = LLVM.StructType(LLVMType[]; ctx)
+	T_jlvalue = LLVM.StructType(LLVMType[])
```
vchuravy, line 458: Formatting

vchuravy (commented):
Please make sure that formatting is correct. I don't fully understand why the `JuliaContext() do ctx` is necessary. Does this push a new context onto the stack?

motabbara:
> Please make sure that formatting is correct. I don't fully understand why the `JuliaContext() do ctx` is necessary. Does this push a new context onto the stack?

JuliaContext where? In active_reg ?

src/compiler.jl
motabbara: Not needed here and now removed. Also removed in a few other places and only exists in trampolines, thunk, deferred_codegen and enzyme_call. The do bodies are now indented.

src/compiler.jl
motabbara: Isn't the diff here on the removal of ctx rather than whitespace? I've otherwise gone through whitespace diffs and tried to get rid of them apart from a couple of weird indents that seemed out of place.

vchuravy:
Yeah can we now just get rid of `JuliaContext() do ctx`, everywhere?

motabbara:
@vchuravy can you please take a look at whether the usage of GPUCompiler.finish_module! is correct in src/compiler ? process_module! is now deprecated. Is finish_ir! needed somewhere? More generally, I don't have a CUDA machine and Metal produces no IR for a autodiff_deferred kernel per the test so I'm missing something here.

motabbara:
> Yeah can we now just get rid of `JuliaContext() do ctx`, everywhere?

It has to be in at least a few places per the latest commit.

motabbara:
> > Yeah can we now just get rid of `JuliaContext() do ctx`, everywhere?
> 
> It has to be in at least a few places per the latest commit. 

Everywhere that previously took a ctx argument has to sit in the same JuliaContext. Internally, LLVM.jl supplies ctx from task-local storage that is populated and depopulated through JuliaContext()-do. I figure we have to at least a few unless I've misunderstood the setup.

test/runtests.jl
```diff
@@ -722,7 +721,7 @@ end
 end
 
 ## https://github.com/JuliaDiff/ChainRules.jl/tree/master/test/rulesets
-if !Sys.iswindows()
+if !Sys.iswindows() && !(Sys.ARCH == :aarch64 && Sys.isapple())
```
wsmoses, line 724: is this needed?

test/runtests.jl
```diff
@@ -21,7 +21,7 @@ using Enzyme_jll
 @info "Testing against" Enzyme_jll.libEnzyme
 
 # Test against FiniteDifferences
-function test_scalar(f, x; rtol=1e-9, atol=1e-9, fdm=central_fdm(5, 1), kwargs...)
+function test_scalar(f, x; rtol=1e-7, atol=1e-9, fdm=central_fdm(5, 1), kwargs...)
```
wsmoses, line 24: is this needed?

test/runtests.jl
motabbara: This test suite is currently broken on M1/M2 with an assert failure in Enzyme proper. Will remove for now.

test/runtests.jl
motabbara: Will remove.

test/Manifest.toml
```diff
@@ -0,0 +1,521 @@
...
+julia_version = "1.11.0-DEV"
+manifest_format = "2.0"
+project_hash = "0d24394284ba951660989e7593688d13df20c8a4"
+
+[[deps.ANSIColoredPrinters]]
```
wsmoses, line 7: If you remove this manifest and we test this on cuda, this seems reasonable to me.

PR #951, opened 2023-07-15
[CI] Update TagBot setup

giordano:
Switch to comment-based setup, instead of hourly cron jobs.


wsmoses:
@vchuravy since I don't know much about tagbot I'll let you review

PR #996, opened 2023-08-10
Add macro to import rrule from chainrules

wsmoses:


Project.toml
```diff
@@ -5,6 +5,8 @@ version = "0.11.7"
 
 [deps]
 CEnum = "fa961155-64e5-5f13-b03f-caf6b980ea82"
+ChainRules = "082447d4-558c-5d27-93f4-14fc19e9eca2"
```
vchuravy, line 8: Can we avoid the dependency on `ChainRules`?

wsmoses: Yeah I would want to do this as an extension module or something?

vchuravy: I think we need to define `@import_frule` in Enzyme proper, but then we could do the importing of actually rules in a package extension? It is not clear to me how you intend to use this?

wsmoses: Like this:

```julia
julia> using Enzyme, ChainRules; Enzyme.@import_frule(typeof(Base.sort), Any)
julia> x=[1.0, 2.0, 0.0]; dx=[0.1, 0.2, 0.3]; ddx = [0.01, 0.02, 0.03]
julia> Enzyme.autodiff(Forward, sort, Duplicated, BatchDuplicated(x, (dx,ddx)))
(var"1" = [0.0, 1.0, 2.0], var"2" = (var"1" = [0.3, 0.1, 0.2], var"2" = [0.03, 0.01, 0.02]))

julia> Enzyme.autodiff(Forward, sort, DuplicatedNoNeed, BatchDuplicated(x, (dx,ddx)))
(var"1" = (var"1" = [0.3, 0.1, 0.2], var"2" = [0.03, 0.01, 0.02]),)

julia> Enzyme.autodiff(Forward, sort, DuplicatedNoNeed, BatchDuplicated(x, (dx,)))
(var"1" = [0.3, 0.1, 0.2],)

julia> Enzyme.autodiff(Forward, sort, Duplicated, BatchDuplicated(x, (dx,)))
(var"1" = [0.0, 1.0, 2.0], var"2" = [0.3, 0.1, 0.2])
```

wsmoses:
I’d strongly advocate against that. This is essentially a nice fallback for
quickly prototyping things in, but has very specific limitations both for
what it can be used on, and performance.

I’m personally fine hiding it behind a hidden extension package, or maybe
even it’s own package

Issue #880, opened 2023-05-29
partition, type analysis mismatch


jgreener64:
I am on Julia 1.9.0 and Enzyme 16818fd3d39d0583915aee38595b54a7fcce6b58. The following errors, though it works on Julia 1.8.5 (with `warning: didn't implement memmove, using memcpy as fallback which can result in errors`).
```julia
using Enzyme, Statistics
f(x) = median([2.0, 1.0, x])
autodiff(Reverse, f, Active, Active(1.5))
```
```
Assertion failed: isPointerArithmeticInst(cur, true, true), file /workspace/srcdir/Enzyme/enzyme/Enzyme/GradientUtils.cpp, line 9182

[21744] signal (22): SIGABRT
in expression starting at REPL[3]:1
crt_sig_handler at C:/workdir/src\signals-win.c:95
raise at C:\WINDOWS\System32\msvcrt.dll (unknown line)
abort at C:\WINDOWS\System32\msvcrt.dll (unknown line)
assert at C:\WINDOWS\System32\msvcrt.dll (unknown line)
needsCacheWholeAllocation at /workspace/srcdir/Enzyme/enzyme/Enzyme\GradientUtils.cpp:9181
calculateUnusedValuesInFunction at /workspace/srcdir/Enzyme/enzyme/Enzyme\EnzymeLogic.cpp:661
CreateAugmentedPrimal at /workspace/srcdir/Enzyme/enzyme/Enzyme\EnzymeLogic.cpp:2340
recursivelyHandleSubfunction at /workspace/srcdir/Enzyme/enzyme/Enzyme\AdjointGenerator.h:8778
visitCallInst at /workspace/srcdir/Enzyme/enzyme/Enzyme\AdjointGenerator.h:12417
delegateCallInst at /opt/x86_64-w64-mingw32/x86_64-w64-mingw32/sys-root/usr/local/include/llvm/IR\InstVisitor.h:302 [inlined]
visitCall at /opt/x86_64-w64-mingw32/x86_64-w64-mingw32/sys-root/usr/local/include/llvm/IR\Instruction.def:209 [inlined]
visit at /opt/x86_64-w64-mingw32/x86_64-w64-mingw32/sys-root/usr/local/include/llvm/IR\Instruction.def:209
visit at /opt/x86_64-w64-mingw32/x86_64-w64-mingw32/sys-root/usr/local/include/llvm/IR\InstVisitor.h:112 [inlined]
[...]

PR #1000, opened 2023-08-14
sort! rules

jgreener64:
I made Enzyme rules for `sort!` as suggested in #880. These are my first Enzyme rules so will need some checking.

I put them in `src/internal_rules.jl` for now, let me know if they should go somewhere else. I have tests too which I can add later.

### DIFF
--- a/src/internal_rules.jl
+++ b/src/internal_rules.jl
@@ -481,4 +481,56 @@ function EnzymeRules.reverse(config, func::Const{typeof(Base.hvcat_fill!)}, ::Ty
     end
     return (nothing, nothing)
 end
+
+function EnzymeRules.forward(
+        ::Const{typeof(sort!)},
+        RT::Type{<:Union{Const, DuplicatedNoNeed, Duplicated}},
+        xs::Duplicated;
+        kwargs...
+    )
+    inds = sortperm(xs.val; kwargs...)
+    xs.val .= xs.val[inds]
+    xs.dval .= xs.dval[inds]
+    if RT <: Const
+        return xs.val
+    elseif RT <: DuplicatedNoNeed
+        return xs.dval
+    else
+        return xs
+    end
+end
+
+function EnzymeRules.augmented_primal(
+        config::EnzymeRules.ConfigWidth{1},
+        ::Const{typeof(sort!)},
+        RT::Type{<:Union{Const, DuplicatedNoNeed, Duplicated}},
+        xs::Duplicated;
+        kwargs...
+    )
+    inds = sortperm(xs.val; kwargs...)
+    if EnzymeRules.needs_primal(config)
+        primal = xs.val[inds]
+    else
+        primal = nothing
+    end
+    if RT <: Const
+        shadow = nothing
+    else
+        shadow = xs.dval[inds]
+    end
+    return EnzymeRules.AugmentedReturn(primal, shadow, inds)
+end
+
+function EnzymeRules.reverse(
+        config::EnzymeRules.ConfigWidth{1},
+        ::Const{typeof(sort!)},
+        RT::Type{<:Union{Const, DuplicatedNoNeed, Duplicated}},
+        tape,
+        xs::Duplicated;
+        kwargs...,
+    )
+    inds = tape
+    back_inds = sortperm(inds)
+    xs.dval .= xs.dval[back_inds]
+    return (nothing,)
 end
--- a/src/internal_rules.jl
+++ b/src/internal_rules.jl
@@ -508,15 +508,17 @@ function EnzymeRules.augmented_primal(
         kwargs...
     )
     inds = sortperm(xs.val; kwargs...)
+    xs.val .= xs.val[inds]
+    xs.dval .= xs.dval[inds]
     if EnzymeRules.needs_primal(config)
-        primal = xs.val[inds]
+        primal = xs.val
     else
         primal = nothing
     end
     if RT <: Const
         shadow = nothing
     else
-        shadow = xs.dval[inds]
+        shadow = xs.dval
     end
     return EnzymeRules.AugmentedReturn(primal, shadow, inds)
 end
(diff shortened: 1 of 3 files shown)


src/internal_rules.jl
```diff
@@ -99,3 +99,56 @@ end
...
 end
+
+function EnzymeRules.forward(
+        ::Const{typeof(sort!)},
+        RT::Type{<:Union{Const, DuplicatedNoNeed, Duplicated}},
+        xs::Duplicated;
+        kwargs...
+    )
+    inds = sortperm(xs.val; kwargs...)
+    xs.val .= xs.val[inds]
```
wsmoses, line 110: Looks like sortperm itself just makes a list of 1:N, then sorts that in place using the original data as the order: https://github.com/JuliaLang/julia/blob/750df9fb5bede16f321f5d5405943d12aec7b83e/base/sort.jl#L1756

Could we change this to first sort the derivative array in place using the primal array as the order, then do the actual sort on the primal? That way we don't have that temporary

src/internal_rules.jl
```diff
@@ -99,3 +99,56 @@ end
...
+    xs.dval .= xs.dval[inds]
+    if RT <: Const
+        return xs.val
```
wsmoses, line 113: Looks good, but mind also adding support for BatchDuplicated/BatchDuplicatedNoNeed?

also tests?

src/internal_rules.jl
```diff
@@ -99,3 +99,56 @@ end
...
+
+function EnzymeRules.augmented_primal(
+        config::EnzymeRules.ConfigWidth{1},
+        ::Const{typeof(sort!)},
+        RT::Type{<:Union{Const, DuplicatedNoNeed, Duplicated}},
+        xs::Duplicated;
+        kwargs...
+    )
+    inds = sortperm(xs.val; kwargs...)
+    if EnzymeRules.needs_primal(config)
```
wsmoses, line 129: we still need to sort the primal and shadow here, like in the forward pass. The only difference is that we should now also cache the permutation, and the reverse pass will invert the permutation

jgreener64:
Added tests and changed the augmented primal. I will address the other points later.

jgreener64:
Added a rule for batched mode.

I had a look at sorting without allocating the index array but couldn't get it to work. The `Perm` order at https://github.com/JuliaLang/julia/blob/750df9fb5bede16f321f5d5405943d12aec7b83e/base/sort.jl#L1756 is specifically for ordering indices, rather than generic ordering of another array.

test/runtests.jl
```diff
@@ -77,6 +77,7 @@ include("typetree.jl")
     include("rrules.jl")
     include("kwrules.jl")
     include("kwrrules.jl")
+    include("internal_rules.jl")
```
wsmoses, line 80: Maybe also move the A \ B rule here (its at the end of the file).

wsmoses (approved):

jgreener64:
Is this ready for merge? Test failures look unrelated.

matinraayai:
@jgreener64 will this rule also cover out-of-place ```sort``` across different dims?

jgreener64:
In general `sort` will call this rule yes since it calls `sort!` internally: https://github.com/JuliaLang/julia/blob/750df9fb5bede16f321f5d5405943d12aec7b83e/base/sort.jl#L1495-L1500.

`sort` with `dims` though maybe not as that goes down a different path: https://github.com/JuliaLang/julia/blob/750df9fb5bede16f321f5d5405943d12aec7b83e/base/sort.jl#L1817. Give it a go and see what happens?

matinraayai:
> In general `sort` will call this rule yes since it calls `sort!` internally: https://github.com/JuliaLang/julia/blob/750df9fb5bede16f321f5d5405943d12aec7b83e/base/sort.jl#L1495-L1500.
> 
> `sort` with `dims` though maybe not as that goes down a different path: https://github.com/JuliaLang/julia/blob/750df9fb5bede16f321f5d5405943d12aec7b83e/base/sort.jl#L1817. Give it a go and see what happens?

This is what I'm getting on Enzyme Master right now:
```
Illegal updateAnalysis prev:{[-1]:Pointer, [-1,-1]:Float@double} new: {[-1]:Pointer, [-1,0]:Integer, [-1,1]:Integer, [-1,2]:Integer, [-1,3]:Integer, [-1,4]:Integer, [-1,5]:Integer, [-1,6]:Integer, [-1,7]:Integer}
val:   %47 = bitcast double addrspace(13)* %46 to i64 addrspace(13)*, !dbg !291 origin=  store i64 %42, i64 addrspace(13)* %47, align 8, !dbg !291, !tbaa !207, !alias.scope !210, !noalias !211

Caused by:
Stacktrace:
 [1] setindex!
   @ ./array.jl:1019
 [2] _setindex_ra!
   @ ./reinterpretarray.jl:537
 [3] setindex!
   @ ./reinterpretarray.jl:505
 [4] _sort!
   @ ./sort.jl:906
```
I was hoping this PR would fix this issue, but if not maybe I can help add something new to the rule to cover the ```dim``` keyword?

jgreener64:
Does that error occur with this PR as well?

jgreener64:
It looks like a specific rule needs to be written for this path: https://github.com/JuliaLang/julia/blob/750df9fb5bede16f321f5d5405943d12aec7b83e/base/sort.jl#L1817.

Though as a short-term fix you can copy the array and call `sort!`, that should hit the above rules and be fast.

matinraayai:
@jgreener64 I will do that (I don't care about the unsorted array). I thought the rules would apply to the `!` variant too however?

jgreener64:
The rules in this PR will always apply to `sort!` and will apply to `sort` in cases where `dims` is not set. So the workaround for `sort` with `dims` is to copy the array and use `sort!`, which always hits these rules.

PR #1002, opened 2023-08-15
support wide ints in tape

### LINKED ISSUE #998: tape width issue using InlineStrings
Below fails since bar member of struct appears as i512. 

`
using Enzyme
using InlineStrings

struct Foo
    x::Float64
    bar::String63
end

function g(f::Foo)
    return f.x*5.0
end

function test()
    autodiff(Reverse, g, Active, Active(Foo(3.0, "a")))
end

test()`

motabbara:
Fixes #998


src/compiler.jl
```diff
@@ -6059,10 +6059,8 @@ function to_tape_type(Type::LLVM.API.LLVMTypeRef)::Tuple{DataType,Bool}
             return UInt32, false
         elseif N == 64
             return UInt64, false
-        elseif N == 128
-            return UInt128, false
         else
-            error("Can't construct tape type for integer of width $N")
```
vchuravy, line 6065: Why delete the branch for UInt128?

motabbara: Otherwise we assert in LLVM with getActiveBits() <= 64 && "Too many bits for uint64_t"

wsmoses:
I'm a little worried that the mismatch here of int28->tuple of int8's may create an error similar to https://github.com/EnzymeAD/Enzyme.jl/issues/999

src/compiler.jl
```diff
@@ -6072,10 +6072,8 @@ function to_tape_type(Type::LLVM.API.LLVMTypeRef)::Tuple{DataType,Bool}
...
+            return NTuple{Int(N), UInt8}, false
```
wsmoses, line 6076: N/8 presumably, but per other comment I'm somewhat worried about this

motabbara: I think you would have seen that before since Julia 1.9.2 and earlier will have LLVM assert on the i128 like in my reply to @vchuravy. The other approach is to keep the UInt128 branch and accept failing on 1.9.2 and below.

wsmoses: let's try to leave the i128 version and track down and fix the assertion

motabbara: It's this https://github.com/llvm/llvm-project/issues/23879

wsmoses: Hm okay in whch case yeah I think I'd prefer to version gate the test personally.

wsmoses: Also can we restore the old error if other int type size, also N should be N == 128 not N = 128 here

motabbara: The inline string looks i512 or i1024 and we break here even on Julia 1.10 and nightly. These types can be defined by e.g., primitive type VeryLongInt <: Integer 512 end
so I don't think an error is the right thing here.

wsmoses: I mean the thing that matters here is that convert(LLVMType, output) must be the same llvm type that this takes as an input. So if we return VeryLongInt for an i512, VeryLongInt must lower to an i512

motabbara: Are the CI failures for type preservation related to this PR?

src/compiler.jl
```diff
@@ -6101,7 +6103,24 @@ function to_tape_type(Type::LLVM.API.LLVMTypeRef)::Tuple{DataType,Bool}
         elseif N == 128
             return UInt128, false
         else
-            error("Can't construct tape type for integer of width $N")
+            T = get(WideTypes, N, nothing)
+            if isnothing(T)
+                symname = Symbol(:UInt,N)
+                emptysvec = Core.svec()
+                _module = Main
+                Tp = GC.@preserve emptysvec symname @ccall jl_new_primitivetype(symname::Symbol,
```
wsmoses, line 6111: is there a reason we have to do this, rather than some sort of @ eval thing cc @vchuravy

motabbara: Because eval cannot be used in a generated function and we're inside one at this point. I guess the ccall shouldn't really be valid for the same reason. A bit stuck here unless we can go back to returning NTuple{N >> 6, UInt64} for N > 128. The only way to get to that branch in Julia is from new primitive types and there are no architectures that Julia supports where anything larger than i128 is supported so I don't think this impacts anything that has the possibility of performing arithmetic.

wsmoses:
I suppose we can turn this into an error then and just make our own
primitive type for any other type we see. I kind of assume we won’t get
arbitrary bit widths so having at most a few extra seems fine to me.

motabbara:
> I suppose we can turn this into an error then and just make our own primitive type for any other type we see. I kind of assume we won’t get arbitrary bit widths so having at most a few extra seems fine to me.
> […](#)

I've changed to that approach now.

wsmoses (approved):

PR #1029, opened 2023-08-25
Only lock cache for writing

MilesCranmer:
Currently the `cached_compilation` method locks the cache for both reads and writes.

Since the compilation cache is not modified once written-to, it only needs to be locked for writes. This PR implements the change.

It also switches to `haskey(cache, key)` from `get(cache, key, nothing) === nothing` which improves type stability.

### DIFF
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -9739,16 +9739,18 @@ const cache_lock = ReentrantLock()
 @inline function cached_compilation(@nospecialize(job::CompilerJob))::CompileResult
     key = hash(job)
 
+    haskey(cache, key) && return cache[key]
     # NOTE: no use of lock(::Function)/@lock/get! to keep stack traces clean
     lock(cache_lock)
     try
-        obj = get(cache, key, nothing)
-        if obj === nothing
+        if haskey(cache, key)
+            cache[key]
+        else
             asm = _thunk(job)
             obj = _link(job, asm)
             cache[key] = obj
+            obj
         end
-        obj
     finally
         unlock(cache_lock)
     end


vchuravy:
cc: @maleadt

I am not a 100% sure that is legal. What happens on a growth of the underlying dict, due to a concurrent write.

MilesCranmer:
> What happens on a growth of the underlying dict, due to a concurrent write.

If you are modifying the cache outside of the locked region, then the code would already be illegal...

MilesCranmer:
Actually maybe I misunderstood what you mean? Do you mean, what if:

1. Thread 1 reads cache[key_a], around the time
2. Thread 2 writes cache[key_b]

which is fine. Or if you mean:

1. Thread 1 checks `haskey(cache, key_a)`, around the time
2. Thread 2 writes `cache[key_a]`

which I'm pretty sure is also fine. If it's already written, `haskey(cache, key_a) == true` and there's no race condition since `cache[key_a]` exists. If it's still being written, then `haskey(cache, key_a) == false`, and it would move to the locked portion, thus waiting for thread 2 to finish. i.e., no race condition.

vchuravy:
I think:

```
    Thread 1 cache[key_a]
    Thread 2 cache[key_b] = 
```

Can lead to `Thread 1` getting wrong answers

MilesCranmer:
Oops, you're totally right. `setindex!` has this:

```julia
    # Rehash now if necessary
    if h.ndel >= ((3*sz)>>2) || h.count*3 > sz*2
        # > 3/4 deleted or > 2/3 full
        rehash!(h, h.count > 64000 ? h.count*2 : h.count*4)
    end
```

in `dict.jl`. Which would mean it might change the entire hash table while thread 1 is reading. Thus breaking both `haskey` and `getindex`...

PR #1030, opened 2023-08-25
Don't show ❌ if failure on nightly

MilesCranmer:
This makes the unittests avoid generating a failure when just the Julia nightly tests fail (which often happens due to upstream errors, rather than any error in a package). It also:

1. Excludes windows from `1.10-nightly` which doesn't seem to exist.
2. Excludes the `1.10-nightly` and `assertions=true` build, since `release-1.10-nightly` is not a valid tag on the julia github.

---

Motivation: when I'm browsing GitHub packages and see a project with an ❌ on the main branch like this:

Screenshot 2023-08-25 at 12 25 57

I might assume it's unmaintained or perhaps testing is not a priority. This would make me less likely to use the package.

However Enzyme.jl is obviously well-maintained and testing is robust:

Screenshot 2023-08-25 at 12 29 49

It's just the nightly tests that fail – and these are out of your control. So this PR makes the CI for nightly not fail the whole build. But they will still run so you can get testing information out of them.


.github/workflows/CI.yml
```diff
@@ -132,6 +129,7 @@ jobs:
         env:
           JULIA_PKG_SERVER_REGISTRY_PREFERENCE: eager
       - uses: julia-actions/julia-runtest@v1
+        continue-on-error: ${{ matrix.version == 'nightly' || matrix.version == '1.10-nightly' }}
```
wsmoses, line 132: we should consider 1.10 errors as errors

MilesCranmer: Are 1.10-nightly builds stable? Or is it similar to nightly?

wsmoses: it should be stable yeah

MilesCranmer:
Moved to #1240

PR #1046, opened 2023-09-11
Test more BLAS functions

sethaxen:
This PR adds tests using EnzymeTestUtils for
- [ ] `BLAS.scal!`
- [ ] `BLAS.axpy!`
- [ ] `BLAS.gemv!`
- [ ] `BLAS.spmv!`
- [ ] `BLAS.gemm!`

### DIFF
--- a/test/blas.jl
+++ b/test/blas.jl
@@ -19,15 +19,24 @@ using Test
                 ),
                 Tx in (Const, Duplicated, BatchDuplicated),
                 Ty in (Const, Duplicated, BatchDuplicated),
-                T in (fun == BLAS.dot ? RTs : RCs),
-                (sz, inc) in ((10, 1), ((2, 20), -2))
+                T in (fun == BLAS.dot ? RTs : RCs)
 
                 are_activities_compatible(Tret, Tx, Ty) || continue
-
-                x = randn(T, sz)
-                y = randn(T, sz)
                 atol = rtol = sqrt(eps(real(T)))
-                test_forward(fun, Tret, n, (x, Tx), inc, (y, Ty), inc; atol, rtol)
+
+                @testset "$fun(n, x, incx, y, incy)" begin
+                    @testset for (sz, inc) in ((10, 1), ((2, 20), -2))
+                        x = randn(T, sz)
+                        y = randn(T, sz)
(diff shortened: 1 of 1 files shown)


ZuseZ4:
Answering the question from your old PR, I have not enabled the tablegen rules by default yet.
Mainly I was not convinced that the tests I wrote were enough, which is why I was looking for this larger testset.
We have a flag which allows to enable tablegen though:
`Enzyme.Compiler.bitcode_replacement!(false)`
But if this testset pass for the fallback then I can work myself on fixing remaining tablegen failures and turn it on afterwards.

wsmoses:
@sethaxen can you mark the currently failing tests as skipped (and open corresponding issues)

wsmoses:
@sethaxen 1.6 fails for unrelated reasons.

All of the "local" ones have the fix (aka it requires a new jll with the fix, which we haven't yet released). That fixed your issue for me.

1.7+ have different GC assertion errors. If you can isolate the cause, I can try to include a fix in the same jll release.

wsmoses:
jll bump has landed, so updating this, CI should have that fix. The remaining failures still need investigation if you're able @sethaxen

sethaxen:
Thanks! I'll check later. For some reason dotu and dotc started failing and not I think from changes I made. Still trying to isolate the cause.

wsmoses:
bumping @sethaxen if you've had a chance to isolate?

sethaxen:
> bumping @sethaxen if you've had a chance to isolate?

No, this turned out to be much more time-consuming than anticipated, and I don't think I'll be able to finish it soon. It should be fairly straightforward to finish if someone else wants to take it on, just slow to identify which tests will cause Julia to crash, which need to be marked broken, and what the MWE is for an issue.

ZuseZ4:
```
julia: /buildworker/worker/package_linux64/build/src/llvm-late-gc-lowering.cpp:855: std::vector<int, std::allocator<int> > LateLowerGCFrame::NumberAllBase(State&, llvm::Value*): Assertion `Tracked.size() == Numbers.size()' failed.
```
Julia getting incorrect DiffeUse info, causing gc to free too early? Or is that something else?

wsmoses:
GC related for sure, how or what causes is unclear

PR #1052, opened 2023-09-13
Set active_reg for mutable structs to be duplicated rather than active

wsmoses:


motabbara:
Should immutable structs with container fields also be marked dup or is that already correctly handled?

wsmoses:
That should already be handled.

PR #1079, opened 2023-09-24
[EnzymeTestUtils] Fix batch duplicated return and aliasing return/argument

wsmoses:


sethaxen:
@wsmoses, can you please add a test that was failing before this PR but now passes? It's not clear to me what edge case this solves.

wsmoses:
Yeah will add a test later today for post-commit review (sorry for quick push, am trying to push new version in fix to fix the accidentally failing NNlib tests which turn out to be correct: https://github.com/FluxML/NNlib.jl/pull/536)

PR #1084, opened 2023-09-26
Add enzyme activity rule

wsmoses:


src/compiler.jl
```diff
@@ -279,101 +281,206 @@ end
...
     if isghostty(T) || Core.Compiler.isconstType(T)
         return AnyState
     end
+
+    inactivety = if typeof(world) === Nothing
+        EnzymeCore.EnzymeRules.inactive_type(T)
+    else
+        inmi = GPUCompiler.methodinstance(typeof(EnzymeCore.EnzymeRules.inactive_type), Tuple{Type{T}}, world)
+        args = Any[EnzymeCore.EnzymeRules.inactive_type, T];
+        ccall(:jl_invoke, Any, (Any, Ptr{Any}, Cuint, Any), EnzymeCore.EnzymeRules.inactive_type, args, length(args), inmi)
```
vchuravy, line 383: @wsmoses what did you want to do here? Did you want to do an `invoke_in_world`?

wsmoses: yeah basically to run inactive_type on the given world (which may almost certainly be later than the given world of creation of that method)

PR #1104, opened 2023-10-19
Add tape_type query given a parent compiler job context

wsmoses:
Still needs a test, but pinging @vchuravy for review/thoughts

### DIFF
--- a/src/Enzyme.jl
+++ b/src/Enzyme.jl
@@ -615,6 +615,62 @@ end
     return TapeType
 end
 
+const tape_cache = Dict{UInt, Compiler.CompileResult}()
+
+const tape_cache_lock = ReentrantLock()
+
+@inline function tape_type(parent_job::Union{GPUCompiler.CompilerJob,Nothing}, ::ReverseModeSplit{ReturnPrimal,ReturnShadow,Width,ModifiedBetweenT, RABI}, ::Type{FA}, ::Type{A}, args...) where {FA<:Annotation, A<:Annotation, ReturnPrimal,ReturnShadow,Width,ModifiedBetweenT, RABI<:ABI}
+    width = if Width == 0
+        w = same_or_one(args...)
+        if w == 0
+            throw(ErrorException("Cannot differentiate with a batch size of 0"))
+        end
+        w
+    else
+        Width
+    end
+
+    if ModifiedBetweenT === true
+        ModifiedBetween = falses_from_args(Val(1), args...)
+    else
+        ModifiedBetween = ModifiedBetweenT
+    end
+
+    @assert ReturnShadow
+    TT = Tuple{args...}
+   
+    primal_tt = Tuple{map(eltype, args)...}
+
+    # world = codegen_world_age(eltype(FA), primal_tt)
+
+    mi = fspec(eltype(FA), TT)
+
+    target = Compiler.EnzymeTarget()
+    params = Compiler.EnzymeCompilerParams(Tuple{FA, TT.parameters...}, API.DEM_ReverseModeGradient, width, remove_innerty(A), true, #=abiwrap=#false, ModifiedBetweenT, ReturnPrimal, #=ShadowInit=#false, UnknownTapeType, RABI)
+    job    = Compiler.CompilerJob(mi, CompilerConfig(target, params; kernel=false))
+
+
+    key = hash(hash(job), parent_job)
+
+    # NOTE: no use of lock(::Function)/@lock/get! to keep stack traces clean
+    lock(tape_cache_lock)
+
+    try
+        obj = get(tape_cache, key, nothing)
+        if obj === nothing
+
+            JuliaContext() do ctx
+                _, meta = codegen(:llvm, job; optimize=false, parent_job)
+            end
+            obj = meta.TapeType
+            cache[key] = meta.TapeType
+        end
+        obj
+    finally
+        unlock(tape_cache_lock)
+    end
+end
+
 """
     autodiff_deferred_thunk(::ReverseModeSplit, ftype, Activity, argtypes...)
 
--- a/src/Enzyme.jl
+++ b/src/Enzyme.jl
@@ -723,10 +723,12 @@ result, ∂v, ∂A
 (7.26, 2.2, [3.3])
 ```
 """
-@inline function autodiff_deferred_thunk(::ReverseModeSplit{ReturnPrimal,ReturnShadow,Width,ModifiedBetweenT, RABI}, ::Type{FA}, ::Type{A}, args...) where {FA<:Annotation, A<:Annotation, ReturnPrimal,ReturnShadow,Width,ModifiedBetweenT, RABI<:ABI}
+@inline function autodiff_deferred_thunk(
+    ::ReverseModeSplit{ReturnPrimal,ReturnShadow,Width,ModifiedBetweenT, RABI},
+    ::Type{FA}, ::Type{A}, args...
+) where {FA<:Annotation, A<:Annotation, ReturnPrimal,ReturnShadow,Width,ModifiedBetweenT, RABI<:ABI}
     @assert RABI == FFIABI
     # args′  = annotate(args...)
-    args′  = args
     width = if Width == 0
         w = same_or_one(args...)
         if w == 0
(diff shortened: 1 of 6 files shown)


wsmoses:
@michel2323 @vchuravy re parent job/GPU things, if you want to try that out.

vchuravy:
Looks about right.

wsmoses:
@sriharikrishna

michel2323:
@wsmoses Trying to get GPUs in KA going with `https://github.com/JuliaGPU/KernelAbstractions.jl/tree/enz_rev_gpu`. I use the 'reverse_gpu.jl` script for development. Now, the compiler crashes.

In this PR I did this notable (uninformed) changes:
* Uncomment your line `world=...` to use on fspec
* Changes `key=hash((hash(job), parnent_job)` to `key=`hash((job, parent_job))`. I'm not sure where `tape_caches` is used other than here. I guess it should also be used in the reverse.
* `tape_type()` didn't return any type. So I return `meta.TapeType`.

michel2323:
The log.
[out.log](https://github.com/EnzymeAD/Enzyme.jl/files/13689455/out.log)

michel2323:
@vchuravy Does this make any sense: https://github.com/JuliaGPU/KernelAbstractions.jl/blob/d6c5a721068c7baab3a3d32db3d4c723de0c8cfe/ext/EnzymeExt.jl#L119-L120

vchuravy:
> Does this make any sense:

Yeah, you could probably use `f` and the unwrap the argument types, but identity is as good as any.

michel2323:
@wmoses Same error. Here is the updated log against latest main branches.
[out2.log](https://github.com/EnzymeAD/Enzyme.jl/files/13754193/out2.log)

michel2323:
`autodiff_deferred_thunk` doesn't go through GPUCompiler with CUDA. I commented back in the `annotate(args...`, but there seems to be more going on. Now it goes through this function, but the execution of forward fails.

src/Enzyme.jl
```diff
@@ -725,7 +725,8 @@ result, ∂v, ∂A
 """
 @inline function autodiff_deferred_thunk(::ReverseModeSplit{ReturnPrimal,ReturnShadow,Width,ModifiedBetweenT, RABI}, ::Type{FA}, ::Type{A}, args...) where {FA<:Annotation, A<:Annotation, ReturnPrimal,ReturnShadow,Width,ModifiedBetweenT, RABI<:ABI}
     @assert RABI == FFIABI
-    args′  = annotate(args...)
+    # args′  = annotate(args...)
```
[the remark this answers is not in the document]
wsmoses: It doesnt make sense to have annotate here anyways since it has to take the types not the values, since this is the thunk version

src/Enzyme.jl
```diff
@@ -679,20 +747,59 @@ result, ∂v, ∂A
...
+    # XXX: Calling thunk from autodiff_deferred_thunk is invalid.
     nondef = Enzyme.Compiler.thunk(Val(world), FA, A, TT, #=Split=# Val(API.DEM_ReverseModeGradient), Val(width), ModifiedBetween, #=ReturnPrimal=#Val(ReturnPrimal), #=ShadowInit=#Val(false), RABI)
     TapeType = EnzymeRules.tape_type(nondef[1])
     A2 = Compiler.return_type(typeof(nondef[1]))
 
+
     adjoint_ptr, primal_ptr = Compiler.deferred_codegen(Val(world), FA, Val(TT), Val(A2), Val(API.DEM_ReverseModeGradient), Val(width), ModifiedBetween, Val(ReturnPrimal), #=ShadowInit=#Val(false), TapeType)
     AugT = Compiler.AugmentedForwardThunk{Ptr{Cvoid}, FA, A2, TT, Val{width}, Val(ReturnPrimal), TapeType}
-    @assert AugT == typeof(nondef[1])
+    # @assert AugT == typeof(nondef[1])
```
wsmoses, line 763: @michel2323 can you make this version call the latter?

michel2323: There is no `thunk` call anymore, so I'm unsure what to compare it with.

michel2323: Nevermind fixed.

src/Enzyme.jl
```diff
@@ -679,15 +747,17 @@ result, ∂v, ∂A
...
-   
+
     primal_tt = Tuple{map(eltype, args)...}
     world = codegen_world_age(eltype(FA), primal_tt)
 
     # TODO this assumes that the thunk here has the correct parent/etc things for getting the right cuda instructions -> same caching behavior
+    # XXX: Calling thunk from autodiff_deferred_thunk is invalid.
     nondef = Enzyme.Compiler.thunk(Val(world), FA, A, TT, #=Split=# Val(API.DEM_ReverseModeGradient), Val(width), ModifiedBetween, #=ReturnPrimal=#Val(ReturnPrimal), #=ShadowInit=#Val(false), RABI)
     TapeType = EnzymeRules.tape_type(nondef[1])
     A2 = Compiler.return_type(typeof(nondef[1]))
```
wsmoses, line 758: just have this call autodiff_deferred_thunk(TapeType, ...)

src/Enzyme.jl
```diff
@@ -651,14 +714,17 @@ forward, reverse = autodiff_deferred_thunk(ReverseSplitWithPrimal, Const{typeof(
...
-result, ∂v, ∂A 
+result, ∂v, ∂A
 
 # output
 
 (7.26, 2.2, [3.3])
 ```
 """
-@inline function autodiff_deferred_thunk(::ReverseModeSplit{ReturnPrimal,ReturnShadow,Width,ModifiedBetweenT, RABI}, ::Type{FA}, ::Type{A}, args...) where {FA<:Annotation, A<:Annotation, ReturnPrimal,ReturnShadow,Width,ModifiedBetweenT, RABI<:ABI}
+@inline function autodiff_deferred_thunk(
```
vchuravy, line 724: Should we just delete this function? We could provide a helper that calls `tape_type` internally, but that seems less helpful.

src/Enzyme.jl
```diff
@@ -679,23 +745,62 @@ result, ∂v, ∂A
...
+    primal_ptr = Compiler.deferred_codegen(Val(world), FA, Val(TT), Val(A2), Val(API.DEM_ReverseModePrimal), Val(width), ModifiedBetween, Val(ReturnPrimal), #=ShadowInit=#Val(false), TapeType)
+    adjoint_ptr = Compiler.deferred_codegen(Val(world), FA, Val(TT), Val(A2), Val(API.DEM_ReverseModeGradient), Val(width), ModifiedBetween, Val(ReturnPrimal), #=ShadowInit=#Val(false), TapeType)
     AugT = Compiler.AugmentedForwardThunk{Ptr{Cvoid}, FA, A2, TT, Val{width}, Val(ReturnPrimal), TapeType}
     @assert AugT == typeof(nondef[1])
     AdjT = Compiler.AdjointThunk{Ptr{Cvoid}, FA, A2, TT, Val{width}, TapeType}
     @assert AdjT == typeof(nondef[2])
     AugT(primal_ptr), AdjT(adjoint_ptr)
 end
 
+@inline function autodiff_deferred_thunk(
```
vchuravy, line 768: Maybe move TapeType after FA

michel2323: `autodiff_deferred_thunk` is variadic. As long as we keep both methods I think this could lead to some trouble to distinguish between the two in a call. Or maybe not?

src/Enzyme.jl
michel2323: That would repeat the boilerplate code before the `thunk`, plus still return and do the `asserts`. So not sure that helps anything. @vchuravy mentioned that we should maybe get rid of the tape typeless call?

wsmoses:
Yeah tests would be nice

michel2323:
Added a test. That test with CuArray fails, though, but I guess that's expected as it's not a kernel but uses broadcast.

michel2323:
@wsmoses What still needs to be done here?

test/runtests.jl
```diff
@@ -290,6 +290,39 @@ make3() = (1.0, 2.0, 3.0)
 
 end
 
+@testset "Deferred and deferred thunk" begin
+    function square!(A)
+        A .*= A
```
wsmoses, line 295: Can you simplify this test. For example, to have it be

```
function tapetest(A)
   return A[0] * A[1]
end
```

This should have a tape type of two float64's, rather than the complex and julia-version specific tape type from a broadcast

src/Enzyme.jl
wsmoses: What is A in contrast to A2

src/Enzyme.jl
wsmoses: Yeah I'm fine nuking this

test/runtests.jl
michel2323: Done.

wsmoses (commented):
This looks good to me, except the Julia1.6 errors

@vchuravy any thoughts (maybe from the jit changes?)

src/Enzyme.jl
michel2323: > Yeah I'm fine nuking this

Nuked and changed the API. Moving TapeType to second argument.

wsmoses:
@michel2323 bumping this, once 1.6 is fixed this should be good

michel2323:
> @michel2323 bumping this, once 1.6 is fixed this should be good

@wsmoses Was indeed an easy fix. I merged back `main`. I hope I resolved the conflicts correctly. I've just seen now the orcv1 issue was also touched by you. Might be worth not rebasing.

michel2323:
I don't know about the this error after the merge. Can you look into this?
```
Simple Complex tests: Test Failed at /home/runner/work/Enzyme.jl/Enzyme.jl/test/runtests.jl:354
  Expression: ((autodiff(ReverseHolomorphic, Base.inv, Active, Active(3.0 + 4.0im)))[1])[1] ≈ 0.0112 + 0.0384im
   Evaluated: 0.0112 - 0.038400000000000004im ≈ 0.0112 + 0.0384im
```

wsmoses:
LGTM. I can't actually approve this since I'm the one who made the PR. @vchuravy can you give it a once over (in particular the orcv1), approve, and let's merge!

wsmoses:
and yeah @michel2323 that's a "we should jll bump soon, but I haven't gotten to, since at visit weekend, and there's some other complex number stuff that should also land"

michel2323:
> Still needs a test, but pinging @vchuravy for review/thoughts

I added a test for `tape_type` and deferred thunk. What else?

michel2323:
> LGTM. I can't actually approve this since I'm the one who made the PR. @vchuravy can you give it a once over (in particular the orcv1), approve, and let's merge!

He wrote the orc1 stuff. That one change I did was for sure only a typo level mistake.

wsmoses:
Okay I'll just force merge it then.

PR #1130, opened 2023-11-06
Use emit_exception

vchuravy:
Might need some more work, but should be more CUDA compatible.


wsmoses:
`JIT session error: Symbols not found: [ gpu_report_exception, gpu_signal_exception ]`

wsmoses:
Superceeded on main

PR #1139, opened 2023-11-08
Remove unecessary shows

vchuravy:


wsmoses (commented):
So these shows were here originally because at one point we failed to pass in the Julia object so we needed some use of them to validate

vchuravy:
Sure, but right now they are just noise when you run the testsuite?

wsmoses:
Could we print to like dev null or some other use that ensures the object is valid

PR #1142, opened 2023-11-09
Attempt to fix version testing mishap

wsmoses:


wsmoses:
Didn't fix but still an improvement

PR #1185, opened 2023-12-06
Docs: Mention `gradient` et al. in 'getting started'

### LINKED ISSUE #1035: Mention `gradient`  in the _Getting started_ section
Currently the user-friendly function `gradient` and it's friends are only mentioned in the API section.

I think it would be great for new users to have it in the Readme and in the _Getting started_ section. 

This more in line with other AD packages 
```Julia
gradient(Reverse, rosenbrock, [1.0, 2.0])
```
and seems much simpler than
```Julia
autodiff(Reverse, rosenbrock, Active, Active(1.0), Active(2.0))
```

In case  `gradient` has a lower performance than `autodiff`, the _Getting started_ section would be a good place to explain this too.

lassepe:
Resolves #1035

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -34,30 +34,6 @@ rosenbrock_inp (generic function with 1 method)
 
 ### Reverse mode
 
-For scalar-valued functions that take in a single vector, `gradient` and its in-place variant, `gradient!`, provide a simple way to compute only the gradient of `rosenbrock` at a given point:
-
-```jldoctest rosenbrock
-julia> gradient(Reverse, x -> rosenbrock(x[1], x[2]), [1.0, 2.0])
-2-element Vector{Float64}:
- -400.0
-  200.0
-
-julia> dx = [0.0, 0.0];
-
-julia> gradient!(Reverse, dx, x -> rosenbrock(x[1], x[2]), [1.0, 2.0])
-2-element Vector{Float64}:
- -400.0
-  200.0
-
-julia> dx
-2-element Vector{Float64}:
- -400.0
-  200.0
-```
-
-For the more general case with arbitrary arguments and settings in which both the derivative and the primal result are needed, 
-[`autodiff`](@ref) can be used directly.
-
 The return value of reverse mode [`autodiff`](@ref) is a tuple that contains as a first value
 the derivative value of the active inputs and optionally the primal return value.
 
@@ -94,13 +70,6 @@ Both the inplace and "normal" variant return the gradient. The difference is tha
 
 ### Forward mode
 
-Again, for scalar-valued functions that take in a single vector, `gradient` can be used to compute the gradient of `rosenbrock` at a given point:
-    
-```jldoctest rosenbrock
-julia> gradient(Forward, x -> rosenbrock(x[1], x[2]), [1.0, 2.0])
-(-400.0, 200.0)
-```
-For more fine-grained control and arbitrary function arguments, again [`autodiff`](@ref) can be used directly.
 The return value of forward mode with a `Duplicated` return is a tuple containing as the first value
 the primal return value and as the second value the derivative.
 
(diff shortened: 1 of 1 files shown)


docs/src/index.md
```diff
@@ -33,7 +34,31 @@ rosenbrock_inp (generic function with 1 method)
 
 ### Reverse mode
 
-The return value of reverse mode is a tuple that contains as a first value
+For scalar-valued functions that take in a single vector, `gradient` and its in-place variant, `gradient!`, provide a simple way to compute only the gradient of `rosenbrock` at a given point:
```
wsmoses, line 37: I'm hesitant to have this mentioned first as the gradient interface is much more limited than autodiff, and may also come with performance penalties (especially if one makes a closure of a multi arg function instead of calling the multi arg autodiff).

wsmoses: Could this be qualified in some way or otherwise strongly encourage the use of autodiff (but have an explanation of the gradient for the sake of adding clarity).

lassepe: I've addressed your feedback by making a separate subsection on "convenience functions" which opens with a disclaimer about flexibility and performance implications. Let me know if you think the performance note is now confusing or overly specific.

docs/src/index.md
```diff
@@ -127,6 +129,54 @@ julia> autodiff(Forward, rosenbrock_inp, BatchDuplicated, BatchDuplicated(x, (dx
 (400.0, (var"1" = -800.0, var"2" = 400.0))
 ```
 
+### Convenience functions
+
+!!! note
+    While the convenience functions discussed below use [`autodiff`](@ref) internally, they are generally more limited in their functionality. Beyond that, these convenience functions may also come with performance penalties; especially if one makes a closure of a multi-argument function instead of calling the appropriate multi-argument [`autodiff`](@ref) function directly.
+
+Two convenient functions for common derivative computations are [`gradient`](@ref) and [`jacobian`](@ref).
+As for [`autodiff`](@ref), the mode (forward or reverse) is determined by the first argument.
```
wsmoses, line 138: As like autodiff?

docs/src/index.md
```diff
@@ -127,6 +129,54 @@ julia> autodiff(Forward, rosenbrock_inp, BatchDuplicated, BatchDuplicated(x, (dx
...
+
+The function [`gradient`](@ref) computes the gradient of a scalar-valued function that takes in a single vector at a given point.
```
wsmoses, line 140: For clarity can we say a function with a vector input and scalar return.

Also it may be useful here to mention that these functions take an optional Val(width) where width is how many derivatives to compute in a single batch.

wsmoses: However this shouldn't be confused with the Val argument needed for reverse mode jacobian which is the number of returned values

docs/src/index.md
```diff
@@ -127,6 +129,54 @@ julia> autodiff(Forward, rosenbrock_inp, BatchDuplicated, BatchDuplicated(x, (dx
 (400.0, (var"1" = -800.0, var"2" = 400.0))
 ```
 
+### Convenience functions
+
+!!! note
+    While the convenience functions discussed below use [`autodiff`](@ref) internally, they are generally more limited in their functionality. Beyond that, these convenience functions may also come with performance penalties; especially if one makes a closure of a multi-argument function instead of calling the appropriate multi-argument [`autodiff`](@ref) function directly.
+
+Two convenient functions for common derivative computations are [`gradient`](@ref) and [`jacobian`](@ref).
```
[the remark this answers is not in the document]
lassepe: Done

docs/src/index.md
lassepe: I added the clarifications and also discuss the chunk size argument now. Since I believe that, for a complete overview of the interface the user should consult the API docs (this tutorial does not cover all functionality), I also added  a clarifying remark at the beginning of the tutorial that highlights this and links to the API.

docs/src/index.md
```diff
@@ -127,6 +133,60 @@ julia> autodiff(Forward, rosenbrock_inp, BatchDuplicated, BatchDuplicated(x, (dx
...
+2×2 Matrix{Float64}:
+ -400.0  200.0
+    2.0    1.0
+
+julia> jacobian(Forward, foo, [1.0, 2.0])
+2×2 Matrix{Float64}:
+ -400.0  200.0
+    2.0    1.0
+```
+
```
wsmoses, line 189: Sorry, one final thing, then let's merge it.

Can you similarly have one more example here of forward/reverse jacobian in vector mode

lassepe: No need to be sorry; I like complete docs :).

Looking at the api docs, there may also be an opportunity to aggregate some of the doc strings of methods with optional arguments into one. They are currently quite scattered which is why I actually didn't know there's Jacobian vector mode for reverse too. 

But, if you don't mind, I'll leave that cleanup for another PR.

wsmoses (approved):

PR #1193, opened 2023-12-12
Update compiler.jl

ZuseZ4:
Clearly not enough people know about this flag right now (see slack).


vchuravy (commented):
I would default to tblgen on, and then add a way to turn it off to the docs.

ZuseZ4:
I'd love to do that, I can create a PR on/after Friday.

wsmoses:
Already enabled

PR #1195, opened 2023-12-15
Fix gc error

wsmoses:
Essentially if there was an sret like { jlvaluet, jlvaluet }, and only the second one was loaded, LLVM (specifically the attributor) would remove the load to the first arg of the sret. This would then cause gc errors in calling conv fixup/etc.

This is remedied by making a fake use of the sret that the attributor and other passes cannot delete. However, to ensure we can still dce sret, we have a new analysis that removes unused srets if entirely unnecessary


vchuravy:
> (specifically the attributor)

The murder is always the attributor...

PR #1198, opened 2023-12-16
Fix kindattr check

wsmoses:


wsmoses:
Extension to https://github.com/EnzymeAD/Enzyme.jl/pull/1195

PR #1207, opened 2023-12-18
Disable sort on integers

wsmoses:


wsmoses:
@jgreener64 your sort rule was broken on integers, which resulted in issues when used, for example, on the sort! inside of a sortperm.

This PR is over conservative and changes the sort! rule to only apply to floats.

What I'd really like here is to define the sort rule only for types that are not Enzyme.guaranteed_const. Or at minimum for Vector{Int}. @vchuravy any thoughts here how?

Going to merge this regardless for now since otherwise it breaks other things (like oceananigans)

PR #1220, opened 2023-12-19
[WIP] First stab at Cholesky

michel2323:
Just added files for now:

* cholesky.jl: tests
* forward.jl: forward rules
* reverse.jl: reverse rules (WIP)

Forward.jl should be pretty tidy. The reverse mode is a mess. However, the issue of Duplicate only accepting the same types already arises here. I return a Cholesky object, although it is not a Cholesky decomposition.


forward.jl
```diff
@@ -0,0 +1,59 @@
+function EnzymeRules.forward(
+    ::Const{typeof(cholesky)},
```
wsmoses, line 2: Can you put these both in src/internal_rules.jl

forward.jl
```diff
@@ -0,0 +1,59 @@
...
+    if RT <: Const
+        return fact
+    end
+    # TODO: This will be a problem for sparse matrices as invL and dL are dense
+    invL = inv(fact.L)
+    # TODO: dL is dense even when L was sparse
+    dL = Matrix(fact.L * LowerTriangular(invL * A.dval * invL' * 0.5 * I))
+    dfact = Cholesky(dL, 'L', 0)
+    if RT <: DuplicatedNoNeed
+        return fact
```
wsmoses, line 18: this should be dfact

reverse.jl
```diff
@@ -0,0 +1,68 @@
...
+    println("Custom Cholesky augmented forward rule")
+    # dfact would be a dense matrix
+    if needs_primal(config)
+        return AugmentedReturn(fact, dfact, (fact, dfact, dA))
+    else
+        return AugmentedReturn(nothing, (bx, bstats), (fact,))
+    end
+end
+
+function EnzymeRules.augmented_primal(
```
wsmoses, line 19: We already have a rule for \, see here: https://github.com/EnzymeAD/Enzyme.jl/blob/8fb0f1000380187f395258d08b727a15653cf493/src/internal_rules.jl#L311C59-L311C68

cholesky.jl
```diff
@@ -0,0 +1,101 @@
+using Enzyme
+using LinearAlgebra
+using Random
+using SparseArrays
+using Test
+using FiniteDifferences
+
+include("utils.jl")
+include("utils_check.jl")
```
wsmoses, line 9: These files don't exist, moreover can you put this inside of test/internal_rules.jl (and use the testing framework).

src/internal_rules.jl
```diff
@@ -557,3 +557,132 @@ function EnzymeRules.reverse(
...
+    else
+	A.dval
+    end
+
+    fact = func.val(cache_A; kwargs...)
+
+    for (dA, dfact) in zip(dAs, shadows)
+	    println("Custom Cholesky reverse rule")
+	    #  michel whatever is the rule here
+	    # either way it should read from dfact
```
wsmoses, line 680: @michel2323 the rule should go here.

forward.jl
michel2323: I didn't put them there because I wanted to avoid reloading Enzyme and have an MWE if something fails.

reverse.jl
michel2323: That is for a matrix as an input. That's slow if you call it multiple times as you refactorize. This should take the factorization (and work with other factorizations).

cholesky.jl
michel2323: Yes. Will do.

michel2323:
@wsmoses This shows the issue I don't know really how to resolve. `tests/cholesky.jl` will fail with
```
ERROR: AssertionError: issymmetric(dA)
```
https://github.com/EnzymeAD/Enzyme.jl/blob/1a36e63f032b7b7646553a73fb3af0510e4192e1/src/internal_rules.jl#L692

This is because the adjoint of a matrix represented as a Cholesky factorization is an asymmetric matrix. In the case of LU, this works fine.

michel2323:
Passes the tests for the driver and computing `dxdb` in forward an reverse. `dxdA` should also work. I'd need now some code that actually uses this in something. This only supports `\`. Would be nice to see what fails now in https://github.com/TuringLang/Turing.jl/pull/1887 .

forward.jl
michel2323: Fixed.

src/internal_rules.jl
michel2323: I left your commit in here. Should I change the rule somehow based on this commit?

src/internal_rules.jl
```diff
@@ -557,3 +557,127 @@ function EnzymeRules.reverse(
...
+    copyto!(A.dval, dfact.factors)
+    return (nothing,)
+end
+
+function EnzymeRules.reverse(
+    config,
+    ::Const{typeof(\)},
+    dret,
+    cache,
+    fact::Union{Const, Duplicated}, B::Union{Const, Duplicated};
```
wsmoses, line 675: add vector mode!

michel2323:
@wsmoses Vector mode added for both forward and reverse + some clean up.

src/internal_rules.jl
```diff
@@ -558,3 +558,247 @@ function EnzymeRules.reverse(
...
+    end
+    # TODO: This will be a problem for sparse matrices as invL and dL are dense
+    invL = inv(fact.L)
+    # TODO: dL is dense even when L was sparse
+    dL = Matrix(fact.L * LowerTriangular(invL * A.dval * invL' * 0.5 * I))
+    # TODO: Stored as Cholesky, although it isn't a Cholesky factorization
+    dfact = Cholesky(dL, 'L', 0)
+    if RT <: DuplicatedNoNeed
+        return dfact
+    else
```
wsmoses, line 580: Can you merge these two into one function? It'll reduce duplication. See examples above

src/internal_rules.jl
```diff
@@ -558,3 +558,247 @@ function EnzymeRules.reverse(
...
+    if isa(fact, Duplicated) && isa(B, Duplicated)
+        retdval = zeros(length(retval))
+        mul!(retdval, fact.dval.U, retval)
+        mul!(retdval, fact.dval.L, retdval)
+        retdval .= B.dval .- retdval
+        ldiv!(fact.val, retdval)
+    elseif isa(fact, Duplicated) && isa(B, Const)
+        retdval = zeros(length(retval))
+        mul!(retdval, A, B.val)
+        mul!(retdval, -1, retdval)
```
wsmoses, line 626: Same here

src/internal_rules.jl
michel2323: In the examples above they are split (`sort!`). What happens if `RT` is `Duplicated` and input `BatchDuplicated` and vice versa?

wsmoses:
No not all are split. See \ for example above or even the \ I modified in
this PR

src/internal_rules.jl
```diff
@@ -558,3 +558,244 @@ function EnzymeRules.reverse(
...
+
+function EnzymeRules.forward(
+        ::Const{typeof(\)},
+        RT::Type{<:Union{BatchDuplicatedNoNeed, BatchDuplicated}},
+        fact::Union{Const, BatchDuplicated{T1,N}}, B::Union{Const, BatchDuplicated{T2,N}};
+        kwargs...
+) where {T1,T2,N}
+    retval = copy(B.val)
+    ldiv!(fact.val, retval)
+    bretdval = ntuple(Val(N)) do b
```
michel2323, line 653: This should be parallelized somehow. I don't know how.

wsmoses:
And no activity states must all have the same width. So they need to either
be all duplicated or all batchduplicated if the same width

src/internal_rules.jl
wsmoses: For now tuple is fine and standard practice

michel2323:
@sunxd Added `ldiv!` in anticipation...

@wsmoses No idea what Aqua doesn't like.

src/internal_rules.jl
wsmoses: Bumping this comment for merging. As discussed on call, the same width is used for all types/returns so one cannot have an input with a different batch size (including duplicated (aka 1) vs batch).

src/internal_rules.jl
```diff
@@ -558,3 +558,418 @@ function EnzymeRules.reverse(
...
+        return retval
+    elseif (RT <: DuplicatedNoNeed) || (RT <: BatchDuplicatedNoNeed)
+        return retdval
+    else
+        return BatchDuplicated(retval, ntuple(i -> bretdval[i], Val(N)))
+    end
+end
+
+function EnzymeRules.forward(
+        func::Const{typeof(ldiv!)},
```
wsmoses, line 696: This rule seems wrong (and unnecssary) remove it?

src/internal_rules.jl
```diff
@@ -558,3 +558,418 @@ function EnzymeRules.reverse(
...
+        return retval
+    elseif (RT <: DuplicatedNoNeed)
+        return retdval
+    else
+        return Duplicated(retval, retdval)
+    end
+end
+
+function EnzymeRules.forward(
+        func::Const{typeof(ldiv!)},
```
wsmoses, line 748: Merge these rules?

src/internal_rules.jl
```diff
@@ -558,3 +558,418 @@ function EnzymeRules.reverse(
...
+    config,
+    ::Const{typeof(cholesky)},
+    dret,
+    cache,
+    A;
+    kwargs...
+)
+    (_dfact,) = cache
+    for b in 1:EnzymeRules.width(config)
+        dA = if EnzymeRules.width(config) == 1
```
wsmoses, line 826: see elsewhere, but we've found it simpler to do something like

```julia
dAs = if EnzymeRules.width(config) == 1
  (A.dval,)
else
  A.dval
end
```

and then have the for loop be 
```julia
for (dA, dfact) in zip(dAs, dfacts)
  ...
end
```

src/internal_rules.jl
```diff
@@ -558,3 +558,418 @@ function EnzymeRules.reverse(
...
+    dret,
+    cache,
+    _fact::Annotation{C},
+    _B::Annotation;
+    kwargs...
+) where {C <: Cholesky}
+
+    (x, _dx, _buffer) = cache
+    for b in 1:EnzymeRules.width(config)
+        dx = if EnzymeRules.width(config) == 1
```
wsmoses, line 883: Same comment as above re zip for clarity.

src/internal_rules.jl
michel2323: I don't know how to resolve this. I tried this:

michel2323: Is there a function to access N at runtime?

src/internal_rules.jl
michel2323: It's because of ambiguity with the batch and nonbatched methods both can be called when arguments are `Const`.

src/internal_rules.jl
wsmoses: `EnzymeRules.width(config)`

wsmoses: oh, not offhand for forward mode (and we should probably create one), but you can make one like:

```
@inline width(::Duplicated) = 1
@inline width(::BatchDuplicated{T, N}) where {T, N} = N
@inline width(::DuplicatedNoNeed) = 1
@inline width(::BatchDuplicatedNoNeed{T, N}) where {T, N} = N
```

wsmoses: probably borrow this? https://github.com/EnzymeAD/Enzyme.jl/blob/443f798457da9295598b47cf2fcfe216ca53a2c7/src/Enzyme.jl#L111 and just pass in all the input/return types

wsmoses: Though you don't actually need N here, only if RT <: BatchDuplicated/BatchDuplicatedNoNeed, which you can just get the type parameter from

src/internal_rules.jl
michel2323: Fixed.

michel2323:
@wsmoses I tried to address all the issues. I don't understand why Julia 1.7 and 1.8 don't pass.

devmotion:
I just reran the Turing tests with Enzyme#main and it seems this PR contains a bug that causes a dimension mismatch: https://github.com/TuringLang/Turing.jl/actions/runs/7662529077/job/20883907601?pr=1887#step:6:18724

wsmoses:
Open an issue, ideally with a MWE with just a cholesky call and tag Michel
to investigate?

PR #1232, opened 2024-01-01
[EnzymeCore] Move adapt to extension

wsmoses:
CUDA CI is presently broken since CUDA requires Adapt 4, and this wasn't allowed by compat.

Let's make this an extension while we're bumping versions.


vchuravy:
This broke Adapt on any version of Julia that doesn't support extensions. So Julia 1.8

PR #1240, opened 2024-01-13
Fix GitHub action configuration – `continue-on-error`

MilesCranmer:
Second version of #1030.

The appearance of the  ❌  everywhere in this repo makes it look poorly maintained (which it is not!). Right now a new user curious about Enzyme will land on the GitHub repo, and see everything broken:

Screenshot 2024-01-13 at 23 01 40

including all of the pull requests:

Screenshot 2024-01-13 at 23 01 57

and even the commit history (with many more than just 1 failing job) –

Screenshot 2024-01-13 at 23 07 23

The issue is that julia nightly is always a bit sketchy to test on, and so one should configure CI to not fail if the nightly builds fail – they should simply be warnings. GitHub doesn't have a proper way to do this, the only way is the `continue-on-error` parameter.

`continue-on-error` is actually used in the current CI configuration file but it seems to not actually being applied (#72). This PR implements a fix for that and also for the `enzymetestutils` job.

Relevant docs:
- https://docs.github.com/en/actions/learn-github-actions/contexts
- https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -117,6 +117,7 @@ jobs:
           JULIA_PKG_SERVER_REGISTRY_PREFERENCE: eager
       - name: Build libEnzyme
         if: ${{ matrix.libEnzyme == 'local' && matrix.os != 'macOS-latest'}}
+        continue-on-error: ${{ matrix.version == 'nightly' }}
         id: build_libenzyme
         run: |
           julia --project=deps -e 'using Pkg; Pkg.instantiate()'
@@ -124,6 +125,7 @@ jobs:
           cp LocalPreferences.toml test/
       - name: Build libEnzyme MacOS
         if: ${{ matrix.libEnzyme == 'local' && matrix.os == 'macOS-latest'}}
+        continue-on-error: ${{ matrix.version == 'nightly' }}
         id: build_libenzyme_mac
         run: |
           julia --project=deps -e 'using Pkg; Pkg.instantiate()'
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -118,24 +117,36 @@ jobs:
...
       - name: Build libEnzyme MacOS
         if: ${{ matrix.libEnzyme == 'local' && matrix.os == 'macOS-latest'}}
+        continue-on-error: ${{ matrix.version == 'nightly' }}
+        id: build_libenzyme_mac
         run: |
           julia --project=deps -e 'using Pkg; Pkg.instantiate()'
           SDKROOT=`xcrun --show-sdk-path` julia --project=deps deps/build_local.jl
           cp LocalPreferences.toml test/
       - uses: julia-actions/julia-buildpkg@v1
+        if: matrix.version != 'nightly' || steps.build_libenzyme.outcome == 'success' || steps.build_libenzyme_mac.outcome == 'success'
```
MilesCranmer, line 135: FYI The reason for the `matrix.version != 'nightly'` at the front is that I _think_ the `.outcome` is only available when you set `continue-on-error` to true.

MilesCranmer:
The remaining failures look to be real (apart from the GH action download issue), which might be another good reason to filter nightly Julia builds

vchuravy (approved):

vchuravy:
The windows failure is sporadic and sofar we have struggled to isolate it (and neither @wsmoses nor me use Windows as our primary system)

MilesCranmer:
This sounds similar to some issues I had in SymbolicRegression.jl testing on GitHub actions. In the end I fixed it by simply setting a pretty conservative `--heap-size-hint`. I think the GH actions virtual machine on windows is a bit tricky for the Julia GC to manage the memory usage of so it helps if you force it to get more aggressive. But it could be something else too.

PR #1251, opened 2024-01-26
Cleanup and simplify cholesky handler

wsmoses:


src/internal_rules.jl
```diff
@@ -568,298 +573,258 @@ function EnzymeRules.reverse(
...
-    if RT <: BatchDuplicatedNoNeed
-        return dfact
-    else
-        return BatchDuplicated(fact, dfact)
-    end
-end
+        invL = inv(fact.L)
 
-function EnzymeRules.forward(
-        func::Const{typeof(\)},
```
devmotion, line 618: It seems the PR removed rules for `\` - doesn't this break support for StaticArrays and other arrays that cannot be mutated in-place?

wsmoses:
The definitions were in place so I’m not sure they would’ve applied?

Feel free to try an open an issue with.

wsmoses:
Most of the time static arrays don’t require custom rules (I don’t think we
have any for) as they write functions entirely in Julia rather than calling
BLAS.

Whether they would be helpful, as opposed to necessary, however is a
different question that should be answered with benchmarks.

Issue #1254, opened 2024-01-27
Wrong gradient for `cholesky`


simsurace:
There seems to be something wrong still with the rules for `cholesky`. The issue here seems distinct from #1250. The following test fails so this case seems not to be covered by the current test suite:
```julia
g(A, b) = sum(cholesky(A * A') \ b)
A = rand(3, 3)
b = rand(3)
test_reverse(g, Active, (A, Duplicated), (b, Duplicated))
```

wsmoses:
If you're able to can you make a version of this with constant inputs, and ideally as little computation in the differentiation as possible (eg no matmul)?

simsurace:
Hmm, it's not easy to get it working with finite differences. If you want, this should also work to show there is an issue:
```
h(A, b) = sum(cholesky(A) \ b)
A = Symmetric(rand(3, 3)) + Matrix{Float64}(I, 3, 3)
b = rand(3)
V = zeros(3, 3); V[1, 1] = 1
dA = Enzyme.make_zero(A)
Enzyme.autodiff(Reverse, h, Duplicated(A, dA), Const(b))
(h(A + 1e-10*V, b) - h(A, b)) / 1e-10 - dA[1, 1] # should be approx zero
```

wsmoses:
Oh I meant more can you say A = [0.3, 0.4, ...].

simsurace:
Sure, if it means something to you:
```julia

h(A, b) = sum(cholesky(A) \ b)
A = [1.3 0.5; 0.5 1.5]
b = [1., 2.]
V = [1.0 0.0; 0.0 0.0]
dA = Enzyme.make_zero(A)
Enzyme.autodiff(Reverse, h, Duplicated(A, dA), Const(b))
(h(A + 1e-10*V, b) - h(A, b)) / 1e-10 - dA[1, 1] # should be approx zero

```

simsurace:
I'm trying to figure out why the rule implemented in #1220 is so different from the one in ChainRules, i.e. much shorter and not doing any real computation:

https://github.com/JuliaDiff/ChainRules.jl/blob/e569283d59d08a65244bed9c37f49fdefbe541ab/src/rulesets/LinearAlgebra/factorization.jl#L491C1-L528C4

PR #1256, opened 2024-01-28
Fix cholesky part 2

wsmoses:
Fixes https://github.com/EnzymeAD/Enzyme.jl/issues/1254

### DIFF
--- a/src/internal_rules.jl
+++ b/src/internal_rules.jl
@@ -109,6 +109,7 @@ end
 
 @inline EnzymeRules.inactive_type(v::Type{Nothing}) = true
 @inline EnzymeRules.inactive_type(v::Type{Union{}}) = true
+@inline EnzymeRules.inactive_type(v::Type{Char}) = true
 @inline EnzymeRules.inactive_type(v::Type{T}) where {T<:Integer} = true
 @inline EnzymeRules.inactive_type(v::Type{Function}) = true
 @inline EnzymeRules.inactive_type(v::Type{T}) where {T<:DataType} = true
@@ -138,7 +139,7 @@ function EnzymeRules.forward(::Const{typeof(Base.deepcopy)}, ::Type{<:BatchDupli
 end
 
 # Deepcopy preserving the primal if runtime inactive
-@inline function deepcopy_rtact(copied::RT, primal::RT, seen::IdDict, shadow::RT) where {RT <: Integer}
+@inline function deepcopy_rtact(copied::RT, primal::RT, seen::IdDict, shadow::RT) where {RT <: Union{Integer, Char}}
     return Base.deepcopy_internal(shadow, seen)
 end
 @inline function deepcopy_rtact(copied::RT, primal::RT, seen::IdDict, shadow::RT) where {RT <: AbstractFloat}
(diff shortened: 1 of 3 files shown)


src/internal_rules.jl
```diff
@@ -684,7 +685,7 @@ function EnzymeRules.forward(
     end
 end
 
-function EnzymeRules.augmented_primal(config, func::Const{typeof(cholesky)}, RT::Type, A; kwargs...)
+function EnzymeRules.augmented_primal(config, func::Const{typeof(cholesky)}, RT::Type, A::Annotation{AT}; kwargs...) where {AT <: Array}
```
devmotion, line 688: I think this should restricted 
```suggestion
function EnzymeRules.augmented_primal(config, func::Const{typeof(cholesky)}, RT::Type, A::Annotation{<:Matrix}; kwargs...)
```
or the more general
```suggestion
function EnzymeRules.augmented_primal(config, func::Const{typeof(cholesky)}, RT::Type, A::Annotation{<:AbstractMatrix}; kwargs...)
```

src/internal_rules.jl
```diff
@@ -696,11 +697,11 @@ function EnzymeRules.augmented_primal(config, func::Const{typeof(cholesky)}, RT:
         nothing
     else
         if EnzymeRules.width(config) == 1
-            Cholesky(Matrix(fact), 'L', 0)
+            Enzyme.make_zero(fact)
```
devmotion, line 700: AFAICT this will also create a `Cholesky` object with the same `uplo` char as `fact`? One issue I had noticed (but unfortunately had no time to open an issue) is that on main always `L` is used.

src/internal_rules.jl
```diff
@@ -718,8 +719,8 @@ function EnzymeRules.reverse(
     ::Const{typeof(cholesky)},
     RT::Type,
     dfact,
-    A;
-    kwargs...)
+    A::Annotation{AT};
+    kwargs...) where {AT <: Array}
```
devmotion, line 723: Similarly, maybe use
```suggestion
    A::Annotation{<:Matrix};
    kwargs...)
```
or
```suggestion
    A::Annotation{<:AbstractMatrix};
    kwargs...)
```

src/internal_rules.jl
```diff
@@ -813,14 +814,13 @@ function EnzymeRules.reverse(
             #   dA −= z B(out)^T
 
             func.val(cache_A, dB, kwargs...)
-
             if !isa(A, Const)
                 dA = EnzymeRules.width(config) == 1 ? A.dval : A.dval[b]
 
                 if AType <: Array
-                    mul!(dA, dB, transpose(cache_Bout), 1, -1)
+                    mul!(dA, dB, transpose(cache_Bout), -1, 1)
```
devmotion, line 821: Good! Another issue I had noticed 🙂

src/internal_rules.jl
wsmoses: correct this is what it would do

src/internal_rules.jl
wsmoses: Not opposed to this, but this is the style convention of the rest of the custom rules, so will be consistent for now.

PR #1257, opened 2024-01-28
Restrict `ldiv!` rules to `Cholesky`

devmotion:
I noticed that the `ldiv!` rules added in #1220 also support `Array` arguments (or `Const` etc. of `Array`s). However, no such primal methods exist in LinearAlgebra:
```julia
julia> using LinearAlgebra

julia> methods(ldiv!, Tuple{Array, Any})
# 0 methods for generic function "ldiv!" from LinearAlgebra
```
There exist a few methods for `AbstractMatrix` subtypes such as `Diagonal` but I think it might be best to just restrict the rule to `Cholesky` for now (also in line with the docstring of `ldiv!` which tells users that "The argument A should not be a matrix. Rather, instead of matrices it should be a factorization object").

Additionally, the PR fixes a call to the primal in the `reverse` rule which did not forward the `kwargs...` correctly.

### DIFF
--- a/src/internal_rules.jl
+++ b/src/internal_rules.jl
@@ -622,7 +622,7 @@ function EnzymeRules.forward(
         func::Const{typeof(ldiv!)},
         RT::Type,
         fact::Annotation{<:Cholesky},
-        B::Union{Const{<:AbstractVecOrMat}, DuplicatedNoNeed{<:AbstractVecOrMat}, Duplicated{<:AbstractVecOrMat}, BatchDuplicatedNoNeed{<:AbstractVecOrMat}, BatchDuplicated{<:AbstractVecOrMat}};
+        B;
         kwargs...
 )
     if isa(B, Const)
--- a/src/internal_rules.jl
+++ b/src/internal_rules.jl
@@ -627,14 +627,14 @@ function EnzymeRules.forward(
 )
     if isa(B, Const)
         @assert (RT <: Const)
-        return ldiv!(fact.val, B.val; kwargs...)
+        return func.val(fact.val, B.val; kwargs...)
     else
         N = width(B)
 
         @assert !isa(B, Const)
 
         retval = if !isa(fact, Const) || (RT <: Const) || (RT <: Duplicated) || (RT <: BatchDuplicated)
-            ldiv!(fact.val, B.val; kwargs...)
+            func.val(fact.val, B.val; kwargs...)
         else
             nothing
         end
@@ -660,7 +660,7 @@ function EnzymeRules.forward(
                 mul!(dB, dfact.L, tmp, -1, 1)
             end
 
-            ldiv!(fact.val, dB; kwargs...)
+            func.val(fact.val, dB; kwargs...)
         end
 
         if RT <: Const
@@ -746,7 +746,7 @@ function EnzymeRules.augmented_primal(
         B::Union{Const{<:AbstractVecOrMat}, DuplicatedNoNeed{<:AbstractVecOrMat}, Duplicated{<:AbstractVecOrMat}, BatchDuplicatedNoNeed{<:AbstractVecOrMat}, BatchDuplicated{<:AbstractVecOrMat}};
         kwargs...
 )
-    ldiv!(A.val, B.val; kwargs...)
+    func.val(A.val, B.val; kwargs...)
 
     cache_Bout = if !isa(A, Const) && !isa(B, Const)
         if EnzymeRules.overwritten(config)[3]
@@ -803,7 +803,7 @@ function EnzymeRules.reverse(
             #   dB = z, where  z = inv(A^T) dB
             #   dA −= z B(out)^T
 
-            ldiv!(cache_A, dB; kwargs...)
+            func.val(cache_A, dB; kwargs...)
             if !isa(A, Const)
                 dA = EnzymeRules.width(config) == 1 ? A.dval : A.dval[b]
                 mul!(dA.factors, dB, transpose(cache_Bout), -1, 1)
--- a/src/internal_rules.jl
+++ b/src/internal_rules.jl
@@ -620,20 +620,20 @@ end
 function EnzymeRules.forward(
         func::Const{typeof(ldiv!)},
         RT::Type,
-        fact::Annotation{C},
-        B,
+        fact::Annotation{<:Cholesky},
+        B::Union{Const{<:AbstractVecOrMat}, DuplicatedNoNeed{<:AbstractVecOrMat}, Duplicated{<:AbstractVecOrMat}, BatchDuplicatedNoNeed{<:AbstractVecOrMat}, BatchDuplicated{<:AbstractVecOrMat}};
         kwargs...
-) where {C <: Union{Cholesky,Array}}
+)
     if isa(B, Const)
         @assert (RT <: Const)
-        return func.val(fact.val, B.val; kwargs...)
+        return ldiv!(fact.val, B.val; kwargs...)
     else
         N = width(B)
 
         @assert !isa(B, Const)
 
         retval = if !isa(fact, Const) || (RT <: Const) || (RT <: Duplicated) || (RT <: BatchDuplicated)
-            func.val(fact.val, B.val; kwargs...)
+            ldiv!(fact.val, B.val; kwargs...)
         else
             nothing
         end
@@ -655,16 +655,8 @@ function EnzymeRules.forward(
                     fact.dval[b]
                 end
                 
-                if C <: Array
-                    mul!(dB, dfact, retval, -1, 1)
-                else
-                    tmp = dfact.U * retval
-
-                    dB .-= dfact.L * tmp
-
-                    # if mul! was implemented for LU, this would be faster
-                    # mul!(dB, dfact.L, tmp, -1, 1)
-                end
+                tmp = dfact.U * retval
+                mul!(dB, dfact.L, tmp, -1, 1)
             end
 
             ldiv!(fact.val, dB; kwargs...)
@@ -749,11 +741,10 @@ function EnzymeRules.augmented_primal(
         func::Const{typeof(ldiv!)},
         RT::Type{<:Union{Const, DuplicatedNoNeed, Duplicated, BatchDuplicatedNoNeed, BatchDuplicated}},
 
-        A::Annotation{AType},
-        B::Union{Const, DuplicatedNoNeed, Duplicated, BatchDuplicatedNoNeed, BatchDuplicated};
+        A::Annotation{<:Cholesky},
+        B::Union{Const{<:AbstractVecOrMat}, DuplicatedNoNeed{<:AbstractVecOrMat}, Duplicated{<:AbstractVecOrMat}, BatchDuplicatedNoNeed{<:AbstractVecOrMat}, BatchDuplicated{<:AbstractVecOrMat}};
         kwargs...
-) where {AType <: Union{Cholesky, Array}}
-
+)
     ldiv!(A.val, B.val; kwargs...)
 
     cache_Bout = if !isa(A, Const) && !isa(B, Const)
@@ -796,11 +787,10 @@ function EnzymeRules.reverse(
     func::Const{typeof(ldiv!)},
     dret,
     cache,
-    A::Annotation{AType},
-    B;
+    A::Annotation{<:Cholesky},
+    B::Union{Const{<:AbstractVecOrMat}, DuplicatedNoNeed{<:AbstractVecOrMat}, Duplicated{<:AbstractVecOrMat}, BatchDuplicatedNoNeed{<:AbstractVecOrMat}, BatchDuplicated{<:AbstractVecOrMat}};
     kwargs...
-) where {AType <: Union{Cholesky,Array}}
-
+)
     if !isa(B, Const)
 
         (cache_A, cache_Bout) = cache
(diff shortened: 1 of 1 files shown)


src/internal_rules.jl
```diff
@@ -621,20 +621,20 @@ end
 function EnzymeRules.forward(
         func::Const{typeof(ldiv!)},
         RT::Type,
-        fact::Annotation{C},
-        B,
+        fact::Annotation{<:Cholesky},
+        B::Union{Const{<:AbstractVecOrMat}, DuplicatedNoNeed{<:AbstractVecOrMat}, Duplicated{<:AbstractVecOrMat}, BatchDuplicatedNoNeed{<:AbstractVecOrMat}, BatchDuplicated{<:AbstractVecOrMat}};
```
wsmoses, line 625: This shouldn't be necessary (and is limiting if different annotations come out). The only uses of B are in a similar ldiv and B inputs are guaranteed valid arguments to ldiv (since that's the only way this could be called).

src/internal_rules.jl
```diff
@@ -621,20 +621,20 @@ end
...
         kwargs...
-) where {C <: Union{Cholesky,Array}}
+)
     if isa(B, Const)
         @assert (RT <: Const)
-        return func.val(fact.val, B.val; kwargs...)
+        return ldiv!(fact.val, B.val; kwargs...)
```
wsmoses, line 630: I would keep this as func.val just in case ldiv is overridden somewhere and ends up not referring to the expected function.

wsmoses (changes requested):

src/internal_rules.jl
devmotion: Ah, I was fooled by if-else statements in lines 666-676 - I somehow mixed up `RT` with `B` and assumed that we want `B` to be one of these annotations to ensure that all cases are covered correctly. But
```suggestion
        B::Annotation{<:AbstractVecOrMat};
```
should be fine, shouldn't it? `ldiv!(::Cholesky, x)` is not defined for any other types of `x`:
```julia
julia> methods(ldiv!, Tuple{Cholesky, Any})
# 2 methods for generic function "ldiv!" from LinearAlgebra:
 [1] ldiv!(C::Cholesky{T, <:StridedMatrix{T} where T}, B::StridedVecOrMat{T}) where T<:Union{Float32, Float64, ComplexF64, ComplexF32}
     @ ~/.asdf/installs/julia/1.10.0/share/julia/stdlib/v1.10/LinearAlgebra/src/cholesky.jl:577
 [2] ldiv!(C::Cholesky, B::AbstractVecOrMat)
     @ ~/.asdf/installs/julia/1.10.0/share/julia/stdlib/v1.10/LinearAlgebra/src/cholesky.jl:580
```

src/internal_rules.jl
devmotion:
```suggestion
        return func.val(fact.val, B.val; kwargs...)
```

src/internal_rules.jl
```diff
@@ -621,20 +621,20 @@ end
...
     else
         N = width(B)
 
         @assert !isa(B, Const)
 
         retval = if !isa(fact, Const) || (RT <: Const) || (RT <: Duplicated) || (RT <: BatchDuplicated)
-            func.val(fact.val, B.val; kwargs...)
+            ldiv!(fact.val, B.val; kwargs...)
```
devmotion, line 637:
```suggestion
            func.val(fact.val, B.val; kwargs...)
```

src/internal_rules.jl
```diff
@@ -656,16 +656,8 @@ function EnzymeRules.forward(
...
-                    dB .-= dfact.L * tmp
-
-                    # if mul! was implemented for LU, this would be faster
-                    # mul!(dB, dfact.L, tmp, -1, 1)
-                end
+                tmp = dfact.U * retval
+                mul!(dB, dfact.L, tmp, -1, 1)
             end
 
             ldiv!(fact.val, dB; kwargs...)
```
devmotion, line 663:
```suggestion
            func.val!(fact.val, dB; kwargs...)
```

src/internal_rules.jl
```diff
@@ -813,15 +803,10 @@ function EnzymeRules.reverse(
             #   dB = z, where  z = inv(A^T) dB
             #   dA −= z B(out)^T
 
-            func.val(cache_A, dB, kwargs...)
+            ldiv!(cache_A, dB; kwargs...)
```
devmotion, line 806:
```suggestion
            func.val(cache_A, dB; kwargs...)
```

src/internal_rules.jl
```diff
@@ -750,11 +742,10 @@ function EnzymeRules.augmented_primal(
...
 
-        A::Annotation{AType},
-        B::Union{Const, DuplicatedNoNeed, Duplicated, BatchDuplicatedNoNeed, BatchDuplicated};
+        A::Annotation{<:Cholesky},
+        B::Union{Const{<:AbstractVecOrMat}, DuplicatedNoNeed{<:AbstractVecOrMat}, Duplicated{<:AbstractVecOrMat}, BatchDuplicatedNoNeed{<:AbstractVecOrMat}, BatchDuplicated{<:AbstractVecOrMat}};
         kwargs...
-) where {AType <: Union{Cholesky, Array}}
-
+)
     ldiv!(A.val, B.val; kwargs...)
```
devmotion, line 749:
```suggestion
    func.val(A.val, B.val; kwargs...)
```

src/internal_rules.jl
wsmoses: I'd prefer to leave it unspecified in case Julia APIs or other packages overload the method.

src/internal_rules.jl
```diff
@@ -656,19 +656,11 @@ function EnzymeRules.forward(
...
-                    mul!(dB, dfact, retval, -1, 1)
-                else
-                    tmp = dfact.U * retval
-
-                    dB .-= dfact.L * tmp
-
-                    # if mul! was implemented for LU, this would be faster
-                    # mul!(dB, dfact.L, tmp, -1, 1)
-                end
+                tmp = dfact.U * retval
```
wsmoses, line 659: I'd honestly prefer to leave the support for Array here, if anything because it's what I'd like to exist as an api for cholesky -- but also this is much more efficient for arrays and I'd be interested in adding such a method in the future.

devmotion: But this branch can never be hit, so why should it be kept? There's no method `ldiv!(::Matrix, ...)` (more general `Array`s can definitely not work).

wsmoses: At minimum can you leave the comment 

```
# if mul! was implemented for LU, this would be faster
# mul!(dB, dfact.L, tmp, -1, 1)
```

This is the general fast and correct rule, which doesn't exist as a method, unfortunately doesn't exist as a method for diagonal matricies atm (which I hope gets added).

devmotion: I removed the comment because actually this line is already used (just one line below). `mul!` is already defined for triangular and diagonal matrices:
```julia
julia> methods(mul!, Tuple{AbstractVecOrMat, Diagonal, AbstractVecOrMat, Number, Number})
# 6 methods for generic function "mul!" from LinearAlgebra:
 [1] mul!(C::AbstractVector, A::Union{Bidiagonal, Diagonal, SymTridiagonal, Tridiagonal}, B::AbstractVector, alpha::Number, beta::Number)
     @ ~/.asdf/installs/julia/1.10.0/share/julia/stdlib/v1.10/LinearAlgebra/src/bidiag.jl:425
 [2] mul!(C::AbstractMatrix, A::Union{Bidiagonal, Diagonal, SymTridiagonal, Tridiagonal}, B::LinearAlgebra.AbstractTriangular, alpha::Number, beta::Number)
     @ ~/.asdf/installs/julia/1.10.0/share/julia/stdlib/v1.10/LinearAlgebra/src/special.jl:115
 [3] mul!(C::AbstractMatrix, A::Union{Bidiagonal, Diagonal, SymTridiagonal, Tridiagonal}, B::Union{Bidiagonal, Diagonal, SymTridiagonal, Tridiagonal}, alpha::Number, beta::Number)
     @ ~/.asdf/installs/julia/1.10.0/share/julia/stdlib/v1.10/LinearAlgebra/src/bidiag.jl:429
 [4] mul!(C::AbstractMatrix, A::Union{Bidiagonal, Diagonal, SymTridiagonal, Tridiagonal}, B::AbstractMatrix, alpha::Number, beta::Number)
     @ ~/.asdf/installs/julia/1.10.0/share/julia/stdlib/v1.10/LinearAlgebra/src/bidiag.jl:427
 [5] mul!(C::AbstractMatrix, A::Union{Bidiagonal, Diagonal, SymTridiagonal, Tridiagonal}, B::AbstractVector, alpha::Number, beta::Number)
     @ ~/.asdf/installs/julia/1.10.0/share/julia/stdlib/v1.10/LinearAlgebra/src/bidiag.jl:426
 [6] mul!(C::AbstractMatrix, A::AbstractVecOrMat, B::AbstractVecOrMat, α::Number, β::Number)
     @ ~/.asdf/installs/julia/1.10.0/share/julia/stdlib/v1.10/LinearAlgebra/src/matmul.jl:263

julia> methods(mul!, Tuple{AbstractVecOrMat, LinearAlgebra.AbstractTriangular, AbstractVecOrMat, Number, Number})
# 6 methods for generic function "mul!" from LinearAlgebra:
 [1] mul!(C::AbstractVector, A::LinearAlgebra.AbstractTriangular, B::AbstractVector, alpha::Number, beta::Number)
     @ ~/.asdf/installs/julia/1.10.0/share/julia/stdlib/v1.10/LinearAlgebra/src/triangular.jl:724
 [2] mul!(C::AbstractMatrix, A::LinearAlgebra.AbstractTriangular, B::Union{Bidiagonal, Diagonal, SymTridiagonal, Tridiagonal}, alpha::Number, beta::Number)
     @ ~/.asdf/installs/julia/1.10.0/share/julia/stdlib/v1.10/LinearAlgebra/src/special.jl:113
 [3] mul!(C::AbstractMatrix, A::LinearAlgebra.AbstractTriangular, B::LinearAlgebra.AbstractTriangular, alpha::Number, beta::Number)
     @ ~/.asdf/installs/julia/1.10.0/share/julia/stdlib/v1.10/LinearAlgebra/src/triangular.jl:736
 [4] mul!(C::AbstractMatrix, A::LinearAlgebra.AbstractTriangular, B::AbstractMatrix, alpha::Number, beta::Number)
     @ ~/.asdf/installs/julia/1.10.0/share/julia/stdlib/v1.10/LinearAlgebra/src/triangular.jl:736
 [5] mul!(C::AbstractMatrix, A::LinearAlgebra.AbstractTriangular, B::AbstractVector, alpha::Number, beta::Number)
     @ ~/.asdf/installs/julia/1.10.0/share/julia/stdlib/v1.10/LinearAlgebra/src/triangular.jl:724
 [6] mul!(C::AbstractMatrix, A::AbstractVecOrMat, B::AbstractVecOrMat, α::Number, β::Number)
     @ ~/.asdf/installs/julia/1.10.0/share/julia/stdlib/v1.10/LinearAlgebra/src/matmul.jl:263
```

src/internal_rules.jl
devmotion:
```suggestion
        B;
```

src/internal_rules.jl
```diff
@@ -750,12 +742,11 @@ function EnzymeRules.augmented_primal(
         func::Const{typeof(ldiv!)},
         RT::Type{<:Union{Const, DuplicatedNoNeed, Duplicated, BatchDuplicatedNoNeed, BatchDuplicated}},
 
-        A::Annotation{AType},
-        B::Union{Const, DuplicatedNoNeed, Duplicated, BatchDuplicatedNoNeed, BatchDuplicated};
+        A::Annotation{<:Cholesky},
+        B::Union{Const{<:AbstractVecOrMat}, DuplicatedNoNeed{<:AbstractVecOrMat}, Duplicated{<:AbstractVecOrMat}, BatchDuplicatedNoNeed{<:AbstractVecOrMat}, BatchDuplicated{<:AbstractVecOrMat}};
```
wsmoses, line 746: Same here, this should just be B

src/internal_rules.jl
```diff
@@ -797,11 +788,10 @@ function EnzymeRules.reverse(
     func::Const{typeof(ldiv!)},
     dret,
     cache,
-    A::Annotation{AType},
-    B;
+    A::Annotation{<:Cholesky},
+    B::Union{Const{<:AbstractVecOrMat}, DuplicatedNoNeed{<:AbstractVecOrMat}, Duplicated{<:AbstractVecOrMat}, BatchDuplicatedNoNeed{<:AbstractVecOrMat}, BatchDuplicated{<:AbstractVecOrMat}};
```
wsmoses, line 792: and here

src/internal_rules.jl
devmotion: It's
```suggestion
        B::Union{Const, DuplicatedNoNeed, Duplicated, BatchDuplicatedNoNeed, BatchDuplicated};
```
on main, is this incorrect?

src/internal_rules.jl
devmotion:
```suggestion
    B;
```

src/internal_rules.jl
wsmoses: ah I see what main is doing here, it's doing Any annotation, except active. You should leave the main variant.

I wouldn't specialize the abstractvecormat here.

src/internal_rules.jl
wsmoses: Can you put this as the same as the augmented forward, so all activities, excluding active

src/internal_rules.jl
devmotion: 👍 Should the same (without AbstractVecOrMat) be used also for the `forward` rule?

src/internal_rules.jl
devmotion:
```suggestion
    B::Union{Const, DuplicatedNoNeed, Duplicated, BatchDuplicatedNoNeed, BatchDuplicated};
```

src/internal_rules.jl
wsmoses: No, because Active is only applicable to reverse mode, so all activities are valid there.

src/internal_rules.jl
wsmoses: oh that's cool. If it builds/runs that's great. I somehow remember it failing, so didn't use

wsmoses (approved):

PR #1262, opened 2024-01-31
Throw descriptive AsssertionError instead of `@show` + `@assert` inside of `create_abi_wrapper`

devmotion:
I just ran into
```julia
(actualRetType, literal_rt, rettype) = (actualRetType, literal_rt, rettype) = (actualRetType, literal_rt, rettype) = Gradient - enzyme: Error During Test at REPL[5]:1
  task switch not allowed from inside staged nor pure functions
  Stacktrace:
    [1] try_yieldto(undo::typeof(Base.ensure_rescheduled))
      @ Base ./task.jl:910
    [2] wait()
      @ Base ./task.jl:984
    [3] uv_write(s::Base.TTY, p::Ptr{UInt8}, n::UInt64)
      @ Base ./stream.jl:1048
    [4] unsafe_write(s::Base.TTY, p::Ptr{UInt8}, n::UInt64)
      @ Base ./stream.jl:1120
    [5] write
      @ ./strings/io.jl:244 [inlined]
    [6] print
      @ ./strings/io.jl:246 [inlined]
    [7] print(::Base.TTY, ::String, ::String, ::Vararg{String})
      @ Base ./strings/io.jl:46
    [8] println(::Base.TTY, ::String, ::Vararg{String})
      @ Base ./strings/io.jl:75
    [9] println(::String, ::String)
      @ Base ./coreio.jl:4
   [10] create_abi_wrapper(enzymefn::LLVM.Function, TT::Type, rettype::Type, actualRetType::Type, Mode::Enzyme.API.CDerivativeMode, augmented::Nothing, width::Int64, returnPrimal::Bool, shadow_init::Bool, world::UInt64, interp::Enzyme.Compiler.Interpreter.EnzymeInterpreter)
      @ Enzyme.Compiler ~/.julia/packages/Enzyme/KJgKj/src/compiler.jl:3255
...
```
Based on the stacktrace I assume that the problem is the use of `@show` instead of `@safe_show` (as a few lines below) in the `create_abi_wrapper` function.

By the way, is there any typical problem (or error on my side) when hitting this branch?

### DIFF
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -2221,8 +2221,7 @@ function julia_undef_value_for_type(Ty::LLVM.API.LLVMTypeRef, forceZero::UInt8):
         end
         return ConstantStruct(ty, vals).ref
     end
-    @safe_show "Unknown type to val", Ty
-    @assert false
+    throw(AssertionError("Unknown type to val: $(Ty)"))
 end
 
 function shadow_alloc_rewrite(V::LLVM.API.LLVMValueRef, gutils::API.EnzymeGradientUtilsRef)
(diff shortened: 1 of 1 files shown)


src/compiler.jl
```diff
@@ -3252,7 +3252,7 @@ function create_abi_wrapper(enzymefn::LLVM.Function, TT, rettype, actualRetType,
     if is_adjoint && rettype <: Active
         @assert !sret_union
         if allocatedinline(actualRetType) != allocatedinline(literal_rt)
-            @show actualRetType, literal_rt, rettype
+            @safe_show actualRetType, literal_rt, rettype
```
wsmoses, line 3255: I would recommend turning this into a better `throw(AssertionError(string))` and printing these directly into the string, and getting rid o the assert below

devmotion: I turned it into an `AssertionError` and replaced a few other occurrences of the same pattern in src/compiler.jl as well.

wsmoses (approved):

src/compiler.jl
wsmoses: Awesome!

devmotion:
Are the test errors on Ubuntu related to this PR? At least I didn't see them on the main branch.

devmotion:
> By the way, is there any typical problem (or error on my side) when hitting this branch?

@wsmoses just wondering if there's some experience/knowledge you can share regarding these `allocatedinline` checks?

wsmoses:
It basically is an assert that type inference inside enzyme should be
compatible (aka if by refewnice or not) with the type inference result
passed into enzyme

devmotion:
And if the assertion fails, is it usually an issue with my code or in Enzyme?

wsmoses:
It's sufficiently rare/weird that I don't know

PR #1264, opened 2024-02-01
Add rule for triangular solves

simsurace:
So this is a first attempt by a complete beginner at this.
One issue is that the tests using EnzymeTestUtils fail because they seem to compare the `data` field of the cotangents:
```julia
A = UpperTriangular(rand(2, 2))
B = copy(A); B.data[2, 1] = 0
EnzymeTestUtils.test_approx(A, B) # Errors because the `data` field is not the same
```
The remaining tests currently pass, but they are not covering all cases.

### DIFF
--- a/src/internal_rules.jl
+++ b/src/internal_rules.jl
@@ -531,14 +531,14 @@ function EnzymeRules.reverse(
     end
 
     for (dA, db, dy) in zip(dAs, dbs, dys)
-        z = transpose(cache_A) \ dy
+        z = adjoint(cache_A) \ dy
         if !(typeof(A) <: Const)
-            dA.data .-= _zero_unused_elements!(AT(z * transpose(y)))
+            dA.data .-= _zero_unused_elements!(AT(z * adjoint(y)))
         end
         if !(typeof(b) <: Const)
             db .+= z
         end
-        dy .= eltype(dy)(0)
+        dy .= zero(eltype(dy))
     end
 
     return (nothing,nothing)
--- a/src/internal_rules.jl
+++ b/src/internal_rules.jl
@@ -456,92 +456,48 @@ const EnzymeTriangulars = Union{
 
 function EnzymeRules.augmented_primal(
     config,
-    func::Const{typeof(\)},
+    func::Const{typeof(ldiv!)},
     ::Type{RT},
+    Y::Annotation{YT},
     A::Annotation{AT},
-    b::Annotation{BT}
-) where {RT, AT <: EnzymeTriangulars, BT <: Array}
+    B::Annotation{BT}
+) where {RT, YT <: Array, AT <: EnzymeTriangulars, BT <: Array}
+    cache_Y = EnzymeRules.overwritten(config)[1] ? copy(Y.val) : Y.val
     cache_A = EnzymeRules.overwritten(config)[2] ? copy(A.val) : A.val
-    cache_A = compute_lu_cache(cache_A, b.val)
-    res = (cache_A \ b.val)::eltype(RT)
-    dres = if EnzymeRules.width(config) == 1
-        zero(res)
-    else
-        ntuple(Val(EnzymeRules.width(config))) do i
-            Base.@_inline_meta
-            zero(res)
-        end
-    end
-    retres = EnzymeRules.needs_primal(config) ? res : nothing
-    cache_res = EnzymeRules.needs_primal(config) ? copy(res) : res
-    cache_b = EnzymeRules.overwritten(config)[3] ? copy(b.val) : nothing
-    cache = NamedTuple{
-        (Symbol("1"), Symbol("2"), Symbol("3"), Symbol("4")),
-        Tuple{typeof(res), typeof(dres), typeof(cache_A), typeof(cache_b)}
-    }((cache_res, dres, cache_A, cache_b))
-    return EnzymeRules.AugmentedReturn{typeof(retres), typeof(dres), Any}(retres, dres, cache)
+    cache_A = compute_lu_cache(cache_A, B.val)
+    cache_B = EnzymeRules.overwritten(config)[3] ? copy(B.val) : nothing
+    primal = EnzymeRules.needs_primal(config) ? Y.val : nothing
+    shadow = EnzymeRules.needs_shadow(config) ? Y.dval : nothing
+    return EnzymeRules.AugmentedReturn{typeof(primal), typeof(shadow), Any}(
+        primal, shadow, (cache_Y, cache_A, cache_B))
 end
 
 function EnzymeRules.reverse(
     config,
-    func::Const{typeof(\)},
+    func::Const{typeof(ldiv!)},
     ::Type{RT},
     cache,
+    Y::Annotation{YT},
     A::Annotation{AT},
-    b::Annotation{BT}
-) where {RT, AT <: EnzymeTriangulars, BT <: Array}
-    y, dys, cache_A, cache_b = cache
-
-    if !EnzymeRules.overwritten(config)[3]
-        cache_b = b.val
-    end
-
-    if EnzymeRules.width(config) == 1
-        dys = (dys,)
-    end
-
-    dAs = if EnzymeRules.width(config) == 1
-        typeof(A) <: Const ? (nothing,) : (A.dval,)
-    else
-        if typeof(A) <: Const
-            ntuple(Val(EnzymeRules.width(config))) do i
-                Base.@_inline_meta
-                nothing
+    B::Annotation{BT}
+) where {YT <: Array, RT, AT <: EnzymeTriangulars, BT <: Array}
+    if !isa(Y, Const)
+        (cache_Yout, cache_A, cache_B) = cache
+        for b in 1:EnzymeRules.width(config)
+            dY = EnzymeRules.width(config) == 1 ? Y.dval : Y.dval[b]
+            z = adjoint(cache_A) \ dY
+            if !isa(B, Const)
+                dB = EnzymeRules.width(config) == 1 ? B.dval : B.dval[b]
+                dB .+= z
             end
-        else
-            A.dval
-        end
-    end
-
-    dbs = if EnzymeRules.width(config) == 1
-        if typeof(b) <: Const
-            (nothing,)
-        else
-            (b.dval,)
-        end
-    else
-        if typeof(b) <: Const
-            ntuple(Val(EnzymeRules.width(config))) do i
-                Base.@_inline_meta
-                nothing
+            if !isa(A, Const)
+                dA = EnzymeRules.width(config) == 1 ? A.dval : A.dval[b]
+                dA.data .-= _zero_unused_elements!(AT(z * adjoint(cache_Yout)))
             end
-        else
-            b.dval
+            dY .= zero(eltype(dY))
         end
     end
-
-    for (dA, db, dy) in zip(dAs, dbs, dys)
-        z = adjoint(cache_A) \ dy
-        if !(typeof(A) <: Const)
-            dA.data .-= _zero_unused_elements!(AT(z * adjoint(y)))
-        end
-        if !(typeof(b) <: Const)
-            db .+= z
-        end
-        dy .= zero(eltype(dy))
-    end
-
-    return (nothing,nothing)
+    return (nothing, nothing, nothing)
 end
 
 _zero_unused_elements!(A::UpperTriangular) = triu!(A.data)
--- a/src/internal_rules.jl
+++ b/src/internal_rules.jl
@@ -447,6 +447,99 @@ function EnzymeRules.reverse(config, func::Const{typeof(\)}, ::Type{RT}, cache, 
     return (nothing,nothing)
 end
 
+
+function EnzymeRules.augmented_primal(
+    config,
+    func::Const{typeof(\)},
+    ::Type{RT},
+    A::Annotation{AT},
+    b::Annotation{BT}
+) where {RT, AT <: Union{UpperTriangular, LowerTriangular}, BT <: Array}
+    cache_A = EnzymeRules.overwritten(config)[2] ? copy(A.val) : A.val
+    cache_A = compute_lu_cache(cache_A, b.val)
+    res = (cache_A \ b.val)::eltype(RT)
+    dres = if EnzymeRules.width(config) == 1
+        zero(res)
+    else
+        ntuple(Val(EnzymeRules.width(config))) do i
+            Base.@_inline_meta
+            zero(res)
+        end
+    end
+    retres = EnzymeRules.needs_primal(config) ? res : nothing
+    cache_res = EnzymeRules.needs_primal(config) ? copy(res) : res
+    cache_b = EnzymeRules.overwritten(config)[3] ? copy(b.val) : nothing
+    cache = NamedTuple{
+        (Symbol("1"), Symbol("2"), Symbol("3"), Symbol("4")),
+        Tuple{typeof(res), typeof(dres), typeof(cache_A), typeof(cache_b)}
+    }((cache_res, dres, cache_A, cache_b))
+    return EnzymeRules.AugmentedReturn{typeof(retres), typeof(dres), Any}(retres, dres, cache)
+end
+
+function EnzymeRules.reverse(
+    config,
+    func::Const{typeof(\)},
+    ::Type{RT},
+    cache,
+    A::Annotation{AT},
+    b::Annotation{BT}
+) where {RT, AT <: Union{UpperTriangular, LowerTriangular}, BT <: Array}
+    y, dys, cache_A, cache_b = cache
+
+    if !EnzymeRules.overwritten(config)[3]
+        cache_b = b.val
+    end
+
+    if EnzymeRules.width(config) == 1
+        dys = (dys,)
+    end
+
+    dAs = if EnzymeRules.width(config) == 1
+        typeof(A) <: Const ? (nothing,) : (A.dval,)
+    else
+        if typeof(A) <: Const
+            ntuple(Val(EnzymeRules.width(config))) do i
+                Base.@_inline_meta
+                nothing
+            end
+        else
+            A.dval
+        end
+    end
+
+    dbs = if EnzymeRules.width(config) == 1
+        if typeof(b) <: Const
+            (nothing,)
+        else
+            (b.dval,)
+        end
+    else
+        if typeof(b) <: Const
+            ntuple(Val(EnzymeRules.width(config))) do i
+                Base.@_inline_meta
+                nothing
+            end
+        else
+            b.dval
+        end
+    end
+
+    for (dA, db, dy) in zip(dAs, dbs, dys)
+        z = transpose(cache_A) \ dy
+        if !(typeof(A) <: Const)
+            @show dA.data
+            dA.data .-= AT(z * transpose(y))
+            @show dA.data
+        end
+        if !(typeof(b) <: Const)
+            db .+= z
+        end
+        dy .= eltype(dy)(0)
+    end
+
+    return (nothing,nothing)
+end
+
 @static if VERSION >= v"1.7-"
 # Force a rule around hvcat_fill as it is type unstable if the tuple is not of the same type (e.g., int, float, int, float)
 function EnzymeRules.augmented_primal(config, func::Const{typeof(Base.hvcat_fill!)}, ::Type{RT}, out::Annotation{AT}, inp::Annotation{BT}) where {RT, AT <: Array, BT <: Tuple}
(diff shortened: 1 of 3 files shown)


src/internal_rules.jl
```diff
@@ -447,6 +447,99 @@ function EnzymeRules.reverse(config, func::Const{typeof(\)}, ::Type{RT}, cache,
     return (nothing,nothing)
 end
 
+
+function EnzymeRules.augmented_primal(
+    config,
```
wsmoses, line 452: Why make this a separate rule instead of just adding support in the existing rule for `\` by expanding the scope of input types to include these?

simsurace: Yeah good question. I think there are various ways to refactor but I would like to keep this for after we made sure we have the correct rule. Also I find the rules with too many conditionals to be quite hard to read and reason about (it is already kind of hard by nature).

wsmoses: But isn't the implementation identical to the previous rule?

simsurace: No, [this line](https://github.com/simsurace/Enzyme.jl/blob/41a9f36d9792b4b03054af0d79460e576408ff3e/src/internal_rules.jl#L531) is different than the counterpart in the dense rule and I removed [these lines](https://github.com/simsurace/Enzyme.jl/blob/41a9f36d9792b4b03054af0d79460e576408ff3e/src/internal_rules.jl#L365C1-L381C4).

simsurace: In any case, there are still a lot of problems. The EnzymeTestUtils test seems to test the wrong thing, so we maybe have to overload that for triangular matrices and perhaps will have to for other ones. But I may be mistaken here. The test would pass if the irrelevant entries of the matrix were zero. I don't see why the irrelevant entries of `dA` are non-zero (I added two `@show` to debug). The non-zero ness does not come from my rule. Presumably the shadow is initialized with zeros by EnzymeTestUtils, so where does it come from?

sethaxen:
Actually, EnzymeTestUtils is doing the right thing. You don't want to end up pulling back a nonzero cotangent on the unused triangle, which is what you're doing here, since next Enzyme will pull back the cotangent through the constructor of the triangular matrix, and this pullback just extracts the `data`. So call `tril!` or `triu!` (whichever is correct) before constructing the cotangent.

simsurace:
> Actually, EnzymeTestUtils is doing the right thing. You don't want to end up pulling back a nonzero cotangent on the unused triangle, which is what you're doing here, since next Enzyme will pull back the cotangent through the constructor of the triangular matrix, and this pullback just extracts the `data`. So call `tril!` or `triu!` (whichever is correct) before constructing the cotangent.

I think it relates to my question above, but if I'm assigning to the shadow with the `UpperTriangular` or `LowerTriangular` wrapper, I'm should not be changing the unused triangle. 
```julia
A = UpperTriangular([1 2; 3 4])
B = UpperTriangular([0 0; 1 0])
A .+= B
A.data == [1 2; 3 4] # true
```
So I'm not sure where the non-zero entries come from. So your suggestion is to add a call to `triu!` after the assignment?

sethaxen:
Ah, I see now you're wrapping in `AbstractTriangular` before decrementing. No, that shouldn't be the problem, your rule is fine. The issue seems to be caused by https://github.com/JuliaDiff/FiniteDifferences.jl/blob/0766bbc7b81381b134835d31145ad4822fd7e65f/src/to_vec.jl#L98-L104, specifically, when FiniteDifferences.jl vectorizes a triangular matrix, it ignores the lower triangle. This is fine if one only ever interacts with the `data` via the abstract matrix API. If one ever accesses the unused triangle of `data` directly, then this will do the wrong thing.

We can show this by pirating this method. This overload causes tests to pass for me:

```julia
function FiniteDifferences.to_vec(x::T) where {T<:LinearAlgebra.AbstractTriangular}
    x_vec, back = FiniteDifferences.to_vec(parent(x))
    function AbstractTriangular_from_vec(x_vec)
        return T(reshape(back(x_vec), size(x)))
    end
    return x_vec, AbstractTriangular_from_vec
end
```

We could pirate this method here, but it's not the ideal way of handling this. An alternative is to manually create the tangents (i.e. construct the `Duplicated` yourself and pass to `test_reverse`), instantiating the unused triangle of the shadow of `A.data` as zeros.

A long-term solution would be for EnzymeTestUtils to roll its own `to_vec` alternative that always unwraps everything to scalars or `Array`s.

src/internal_rules.jl
```diff
@@ -447,6 +447,99 @@ function EnzymeRules.reverse(config, func::Const{typeof(\)}, ::Type{RT}, cache,
...
+    func::Const{typeof(\)},
+    ::Type{RT},
+    A::Annotation{AT},
+    b::Annotation{BT}
+) where {RT, AT <: Union{UpperTriangular, LowerTriangular}, BT <: Array}
```
sethaxen, line 457: Can the same rule also support `UnitUpperTriangular` and `UnitLowerTriangular`?

sethaxen:
> Sorry, but who is writing to that unused triangle? If I understand your explanation correctly, it is not FiniteDifferences.

No-one's writing to it. I'll explain. We want to test that a given rule only overwrites the shadow of an argument if that argument is itself overwritten by the primal. In all other cases, it should just increment the shadow. The way to test this is to start with a shadow filled with random numbers. There's then some complexity with how we get FiniteDifferences to work with this, since it doesn't support mutating functions. In short, we replace the mutating function with one that copies all arguments first, and then all arguments are also returned as outputs. Then for these returned arguments we pull back as cotangents the exact same random shadows we use with Enzyme. This causes FD to increment these random cotangents in the same way Enzyme should. It's a little complicated and happens here: https://github.com/EnzymeAD/Enzyme.jl/blob/a511e4e6979d6161699f5c9919d49801c0764a09/lib/EnzymeTestUtils/src/finite_difference_calls.jl#L126-L177

But in this case, FD is intentionally ignoring the unused triangle stored by the triangular matrix, which is a problem for us, because a user need not ignore it, and Enzyme won't either. So FD is not pulling back the random values in the unused triangle, and that causes Enzyme and FD to disagree. This is why either modifying FD to pull back these values as I did above or using zeros instead of random values will work just fine.

simsurace:
Ok but it sounds like if we want to test both aspects we need to keep the random vectors. Looks like FD is providing the wrong abstractions for this. Instead of pirating `to_vec`, why don't we introduce interfaces that are better geared towards Enzyme?

simsurace:
I think I got it wrong. You meant to initialize randomly but only on the used triangle?

sethaxen:
> Ok but it sounds like if we want to test both aspects we need to keep the random vectors.

If we initialize the shadow of the used diagonal randomly, then we still test this for the used diagonal at least, just not the unused one. :man_shrugging: 

> Instead of pirating `to_vec`, why don't we introduce interfaces that are better geared towards Enzyme?

Yes, this is what I meant by the final sentence in https://github.com/EnzymeAD/Enzyme.jl/pull/1264#issuecomment-1921654762 . It just takes more time than I have right now to work on.

sethaxen:
> What if we add a test that ADs through the constructor as well?

Yeah I think that's a good way to go, but for some reason Enzyme errored. But sure, if you can get it working, it's probably best.

simsurace:
What is the API for injecting a custom cotangent into `test_reverse`?

sethaxen:
> What is the API for injecting a custom cotangent into `test_reverse`?

Instead of passing a tuple `(A, TA)`, construct an activity (e.g. `Duplicated`) where the shadow contains the custom cotangent and pass that instead. See https://github.com/EnzymeAD/Enzyme.jl/blob/a511e4e6979d6161699f5c9919d49801c0764a09/lib/EnzymeTestUtils/src/test_reverse.jl#L35-L38

src/internal_rules.jl
simsurace: Yes, done!

simsurace:
The tests should now pass, maybe we can trigger the CI.
I would keep extending and refactoring the existing rule for later.
A remaining doubt I had, @sethaxen: it this the right rule to implement, or should we implement one for `ldiv!` with three arguments instead?

simsurace:
I only ran the tests in a local script, so I forgot to include EnzymeTestUtils. Should be fine now.

src/internal_rules.jl
wsmoses: I'd still like this to be merged with the other rule to minimize the maintenance burden and bug surface (since more code will be shared).

simsurace: Yeah I understand this. There are some things (I believe) that should be fixed for the other rule so if you don't mind I can make a follow-up PR where I fix this and refactor both rules to hopefully merge them (if it is possible).

simsurace: Tangentially, do you think this rule is the right one to implement even though it is not the innermost call before hitting LAPACK (3-arg `ldiv!` is)? If it isn't, and we change it for the triangular matrices, we would probably not be able to merge these rules anyway. 
If we keep it as is, people calling ldiv! directly don't get any rule.

src/internal_rules.jl
```diff
@@ -447,6 +447,108 @@ function EnzymeRules.reverse(config, func::Const{typeof(\)}, ::Type{RT}, cache,
...
+                Base.@_inline_meta
+                nothing
+            end
+        else
+            b.dval
+        end
+    end
+
+    for (dA, db, dy) in zip(dAs, dbs, dys)
+        z = transpose(cache_A) \ dy
```
sethaxen, line 534: Pretty sure this needs to be for the complex case:
```suggestion
        z = adjoint(cache_A) \ dy
```

src/internal_rules.jl
```diff
@@ -447,6 +447,108 @@ function EnzymeRules.reverse(config, func::Const{typeof(\)}, ::Type{RT}, cache,
...
+        if !(typeof(A) <: Const)
+            dA.data .-= _zero_unused_elements!(AT(z * transpose(y)))
```
sethaxen, line 536: Same here:
```suggestion
            dA.data .-= _zero_unused_elements!(AT(z * adjoint(y)))
```

sethaxen (changes requested):
If `\` is under the hood just calling `ldiv!`, and all other function calls are differentiable by Enzyme, then I agree it makes more sense to target `ldiv!`.

test/internal_rules.jl
```diff
@@ -359,5 +360,50 @@ end
...
+    end
+    @testset "test against EnzymeTestUtils through constructor" begin
+        hUT(A, B) = h(UpperTriangular(A), B)
+        hLT(A, B) = h(LowerTriangular(A), B)
+        hUUT(A, B) = h(UnitUpperTriangular(A), B)
+        hULT(A, B) = h(UnitLowerTriangular(A), B)
+        for f in (hUT, hLT, hUUT, hULT),
+            Tret in (Const, Active), TA in (Const, Duplicated), TB in (Const, Duplicated)
+            test_reverse(f, Tret, (M, TA), (B, TB); rtol = 1e-2, atol = 1e-2)
+        end
```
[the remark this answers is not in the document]
simsurace: The loop did not work when I tried it. I had to declare each function explicitly. Maybe it has to do with the type instability that occurs when closing over types.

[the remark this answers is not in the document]
simsurace: As for complex elements, I'll add those.

test/internal_rules.jl
```diff
@@ -359,5 +360,50 @@ end
         @test isapprox(dA, dA_sym)
     end
 end
+
+@testset "Linear solve for triangular matrices" begin
+    h(A, B) = sum(A \ B)
```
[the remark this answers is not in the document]
simsurace: When I switch to `ldiv!` I think this will be resolved.

src/internal_rules.jl
simsurace: Thanks for the hint!

test/internal_rules.jl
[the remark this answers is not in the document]
simsurace: Solved the type instability issue.

test/internal_rules.jl
```diff
@@ -359,5 +360,50 @@ end
...
+            @test A * Y ≈ B
+            dA = Enzyme.make_zero(A)
+            dB = Enzyme.make_zero(B)
+            V = _rand_tangent(A)
+            W = rand(3, 3)
+            Enzyme.autodiff(Reverse, h, Duplicated(A, dA), Duplicated(B, dB))
+            @test (h(A + ϵ*V, B) - h(A, B)) / ϵ ≈ tr(V' * dA) rtol = 1e-5
+            @test (h(A, B + ϵ*W) - h(A, B)) / ϵ ≈ tr(W' * dB) rtol = 1e-5
+        end
+        @testset "test against EnzymeTestUtils" begin
```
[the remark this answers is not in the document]
simsurace: Removed!

simsurace:
I will try to re-write the rule in terms of `ldiv!` next.

src/internal_rules.jl
simsurace: I think this is now obsolete because there is no existing 3-arg `ldiv!` rule implementation.

wsmoses (approved):

sethaxen (approved):

simsurace:
Maybe I should add some tests for batched inputs? Otherwise feel free to merge.

src/internal_rules.jl
```diff
@@ -447,6 +447,64 @@ function EnzymeRules.reverse(config, func::Const{typeof(\)}, ::Type{RT}, cache,
...
+    ::Type{RT},
+    Y::Annotation{YT},
+    A::Annotation{AT},
+    B::Annotation{BT}
+) where {RT, YT <: Array, AT <: EnzymeTriangulars, BT <: Array}
+    cache_Y = EnzymeRules.overwritten(config)[1] ? copy(Y.val) : Y.val
+    cache_A = EnzymeRules.overwritten(config)[2] ? copy(A.val) : A.val
+    cache_A = compute_lu_cache(cache_A, B.val)
+    cache_B = EnzymeRules.overwritten(config)[3] ? copy(B.val) : nothing
+    primal = EnzymeRules.needs_primal(config) ? Y.val : nothing
```
wsmoses, line 469: Ah I see a problem here, you're never running the original function `ldiv!(A, B)`. Can you ensure this happens?

As well as make sure the tests confirm this?

test/internal_rules.jl
```diff
@@ -359,5 +360,27 @@ end
...
     end
 end
+
+@testset "Linear solve for triangular matrices" begin
+    @testset for T in (UpperTriangular, LowerTriangular, UnitUpperTriangular, UnitLowerTriangular),
+        TE in (Float64, ComplexF64), sizeB in ((3,), (3, 3))
+        M = rand(TE, 3, 3)
+        B = rand(TE, sizeB...)
+        Y = zeros(TE, sizeB...)
+        A = T(M)
```
wsmoses, line 370: yeah batch tests as well would be nice

src/internal_rules.jl
simsurace: OMG. That is worrying! So does EnzymeTestUtils not check that the primal is the same as if the function is called?

wsmoses: it appears not @sethaxen ?

sethaxen: That's really strange. This line does check that the output of the primal and of the forward pass are the same:
https://github.com/EnzymeAD/Enzyme.jl/blob/cf0b4674a671c9c8f427e0a04a1c90a247b0adb2/lib/EnzymeTestUtils/src/test_reverse.jl#L129-L131

simsurace: I think it should be doing that:
https://github.com/simsurace/Enzyme.jl/blob/7ad702936e3907b5c582d5c35481e2fe4222a49c/lib/EnzymeTestUtils/src/test_reverse.jl#L129-L131

sethaxen: > So the issue I think is that we need to check all of the in place arguments, not just the actual return. For example, here the issue is that the 1st arg is not mutated in place to have the return.

Aaaah, that's it. Yes we do test that an incorrect primal is detected in https://github.com/EnzymeAD/Enzyme.jl/blob/cf0b4674a671c9c8f427e0a04a1c90a247b0adb2/lib/EnzymeTestUtils/test/test_reverse.jl#L151-L162, but true, we need to do it for all arguments.

simsurace: Let me add a manual test of this so this PR doesn't get blocked by this refactor.

test/internal_rules.jl
simsurace: I added some tests, but they don't pass yet. Maybe we can trigger CI to get a full picture. Will fix them later.

simsurace:
Let's trigger CI to get a full picture of the additional tests.

src/internal_rules.jl
```diff
@@ -447,6 +447,65 @@ function EnzymeRules.reverse(config, func::Const{typeof(\)}, ::Type{RT}, cache,
...
+
+function EnzymeRules.augmented_primal(
+    config,
+    func::Const{typeof(ldiv!)},
+    ::Type{RT},
+    Y::Annotation{YT},
+    A::Annotation{AT},
+    B::Annotation{BT}
+) where {RT, YT <: Array, AT <: EnzymeTriangulars, BT <: Array}
+    func.val(Y.val, A.val, B.val)
```
wsmoses, line 465: Tests presently fail, presumably because you need to cache before running `func.val`

simsurace: Hmm, I thought that the point of caching was to preserve the value before it might be overwritten. So if during the forward pass you do the linear solve and write to Y and that gets overwritten by the subsequent call, in the backward pass you need to retrieve the value from a copy that was not touched by the next function. So this would imply that we need to call the function before making the copy. Is my reasoning off?

wsmoses: No you are correct, the code when I commented ran the in place linear solve, then cached -- meaning you would cache the version overwritten by the linear solve. After making the comment, I subsequently fixed the code.

simsurace: But then my logic _is_ backwards. So the cache in the `reverse` call will be retrieved from the next function? Because the rule needs the `Y` after the linear solve, but before any overwrite has happened.
Would this not hold for the `ldiv!` rule for Cholesky factorizations as well?
In any case we still have failures.

simsurace:
I've stared at the errors for a while but I cannot make sense of them. They are mostly inside the EnzymeTestUtils logic for calling FiniteDifferences for `BatchDuplicated`. Is everything fine in there? @sethaxen could you please have another look?

test/internal_rules.jl
```diff
@@ -386,5 +387,39 @@ end
...
+        @testset "test against EnzymeTestUtils through constructor" begin
+            _A = T(A)
+            function f!(Y, A, B, ::T) where T
+                ldiv!(Y, T(A), B)
+                return nothing
+            end
+            for TY in (Const, Duplicated, BatchDuplicated),
+                TM in (Const, Duplicated, BatchDuplicated),
+                TB in (Const, Duplicated, BatchDuplicated)
+                test_reverse(f!, Const, (Y, TY), (M, TM), (B, TB), (_A, Const))
```
sethaxen, line 419: This should allow the tests to pass. Only certain combinations of activities are valid (e.g. can't mix `Duplicated` and `BatchDuplicated`), and the invalid combinations are the ones causing the current failures:
```suggestion
                are_activities_compatible(Const, TY, TA, TB) || continue
                test_reverse(f!, Const, (Y, TY), (M, TM), (B, TB), (_A, Const))
```

test/internal_rules.jl
```diff
@@ -386,5 +387,39 @@ end
         @test isapprox(dA, dA_sym)
     end
 end
+
+@testset "Linear solve for triangular matrices" begin
+    @testset for T in (UpperTriangular, LowerTriangular, UnitUpperTriangular, UnitLowerTriangular),
+        TE in (Float64, ComplexF64), sizeB in ((3,), (3, 3))
+        M = rand(TE, 3, 3)
```
sethaxen, line 394:
```suggestion
        n = sizeB[1]
        M = rand(TE, n, n)
```

sethaxen (changes requested):

test/internal_rules.jl
```diff
@@ -386,5 +387,39 @@ end
...
+        end
+        @testset "test against EnzymeTestUtils through constructor" begin
+            _A = T(A)
+            function f!(Y, A, B, ::T) where T
+                return ldiv!(Y, T(A), B)
+            end
+            for Tret in (Const, Active),
+                TY in (Const, Duplicated, BatchDuplicated),
+                TA in (Const, Duplicated, BatchDuplicated),
+                TB in (Const, Duplicated, BatchDuplicated)
```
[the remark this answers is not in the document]
sethaxen: Ah, looking at the CI logs, it seems it doesn't succeed.

simsurace:
Despite the check for compatibility of activities, I still get some failures.

sethaxen (approved):
LGTM, thanks!

PR #1265, opened 2024-02-03
Describe activity stable

wsmoses:


docs/src/index.md
```diff
@@ -237,6 +240,44 @@ Enzyme.autodiff(Reverse, f, Active(1.2), Duplicated(Vector{Float64}(undef, 1), V
...
+  if cond
+    return active_var
+  else
+    return constant_var
+end
+
+Enzyme.autodiff(Forward, g, Const(condition), Duplicated(x, dx), Const(y))
+```
+
+The returned value here could either by constant or duplicated, depending on the runtime-defined value of `cond`. If `cond` is true, Enzyme simply returns the shadow of `active_var` as the derivative. However, if `cond` is false, there is no derivative shadow for `constant_var` and Enzyme will throw a "Mismatched activity" error. For some simple types, e.g. a float Enzyme can circumvent this issue, for example by returning the float 0. Similarly, for some types like the Symbol type, which are never differentiable, such a shadow value will never be used, and Enzyme can return the original `primal` value as its derivative.  However, for arbitrary data structures, Enzyme presently has no generic mechanism to resolve this.
```
wsmoses, line 258:
```suggestion
The returned value here could either by constant or duplicated, depending on the runtime-defined value of `cond`. If `cond` is true, Enzyme simply returns the shadow of `active_var` as the derivative. However, if `cond` is false, there is no derivative shadow for `constant_var` and Enzyme will throw a "Mismatched activity" error. For some simple types, e.g. a float Enzyme can circumvent this issue, for example by returning the float 0. Similarly, for some types like the Symbol type, which are never differentiable, such a shadow value will never be used, and Enzyme can return the original "primal" value as its derivative.  However, for arbitrary data structures, Enzyme presently has no generic mechanism to resolve this.
```

gdalle:
Just wanted to signal that this emerged following a Slack conversation around the following issue:
https://github.com/EnzymeAD/Enzyme.jl/issues/1266

PR #1274, opened 2024-02-08
Isolate 1.6 x86 failure

wsmoses:


Project.toml
```diff
@@ -8,6 +8,7 @@ CEnum = "fa961155-64e5-5f13-b03f-caf6b980ea82"
 EnzymeCore = "f151be2c-9106-41f4-ab19-57ee4f262869"
 Enzyme_jll = "7cc45869-7501-5eee-bdea-0790c847d4ef"
 GPUCompiler = "61eb1bfa-7361-4325-ad38-22787b887f55"
+InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
```
vchuravy, line 11: You probably want this in `test/Project.toml`?

wsmoses:
@vchuravy interestingly the correctness issue here [well 2.715 vs 2.718] only occurs when using the call inttoptr jit, and doing the "inline abi" [aka splatting the LLVM into enzyme_call] does not err.

I don't quite understand why. We use Ptr{Cvoid} as the way to pass around things rather than UInt, could that cause it?

```
     %92 = call [1 x double] inttoptr (i32 -591568512 to [1 x double] (double, double)*)(double %.sroa.076.0.copyload, double %.sroa.2.0.copyload)
     %.fca.0.extract = extractvalue [1 x double] %92, 0
```

vs

```
     %75 = call fast double @julia_exp_fast_3192(double %.sroa.074.0.copyload)
     %76 = fmul fast double %75, %.sroa.2.0.copyload
; └└└
;  @ /home/runner/work/Enzyme.jl/Enzyme.jl/src/Enzyme.jl:224 within `autodiff'
  %77 = getelementptr inbounds [1 x double], [1 x double]* %0, i32 0, i32 0
  store double %76, double* %77, align 4
  ret void
```

PR #1276, opened 2024-02-08
fix rendering

ZuseZ4:
Renders now, check if that's what you wanted to write.


examples/autodiff.jl
```diff
@@ -116,9 +116,9 @@ dbx[2] == 1.0
 # (stored in y), we can mark it DuplicatedNoNeed. Specifically, this will perform the following:
 # ```math
 # \begin{aligned}
```
wsmoses, line 118: This should be begin{align} and \end{align}

PR #1300, opened 2024-02-20
TA pass julia type info via param attr

wsmoses:
Requires https://github.com/EnzymeAD/Enzyme/pull/1746


vchuravy:
On @swilliamson7 this seems to yield another 50s improvement from 187->132

We still spent most of our time in type-tree.

image

image

Across those three towers we spend: 83s out of 132s in constructing the type-tree. The closest other thing is ~10s in julia_activity_rule

wsmoses:
Specifically I want to understand whether these functions are called too
many times or if they are too slow

PR #1302, opened 2024-02-21
Memoize typetree calls

vchuravy:
- memoize typetree
- fixup! memoize typetree
- fixup! memoize typetree
- fixup! memoize typetree


src/typetree.jl
```diff
@@ -63,70 +63,86 @@ function to_md(tt::TypeTree, ctx)
     return LLVM.Metadata(LLVM.MetadataAsValue(ccall((:EnzymeTypeTreeToMD, API.libEnzyme), LLVM.API.LLVMValueRef, (API.CTypeTreeRef,LLVM.API.LLVMContextRef), tt, ctx)))
 end
 
-function typetree(::Type{T}, ctx, dl, seen=nothing) where T <: Integer
+const TypeTreeTable = IdDict{Any, Union{Nothing, TypeTree}}
```
wsmoses, line 66: Minor but could we have this be Type as the key?

src/typetree.jl
```diff
@@ -153,21 +169,11 @@ else
...
 
-function typetree(@nospecialize(T), ctx, dl, seen=nothing)
+function typetree_inner(@nospecialize(T), ctx, dl, seen::TypeTreeTable)
     if T isa UnionAll || T isa Union || T == Union{} || Base.isabstracttype(T)
         return TypeTree()
     end
 
-    if seen !== nothing && T ∈ seen
-        return TypeTree()
-    end
```
wsmoses, line 163: I'm a bit concerned about the siblings case here (and not sure if we have a test).

But basically the problem here was if you have
```

struct B
   ...
end

struct A
  lhs::B
  rhs::B
``

The issue was as follows:
we hit this function for A, creating a new map.
We first add B to the seen map via the lhs, computing things.
Next we call this for B again, via the rhs. However if it's the same seen map, we will thing we have already seen B from recursion, when in reality it is because we already computed it from the lhs call.

File: src/typetree.jl (hunk 1 above)
Comment (vchuravy) [reply src/typetree.jl:66]: Doesn't really matter since we use objectid

File: test/typetree.jl [hunk 3]
```diff
@@ -38,4 +48,7 @@ end
     @test at2.y == 0.0
     @test at2.z == 0.0
     @test at2.type == 4
+
+    @test tt(LList2{Float64}) == "{[8]:Float@double}"
+    @test tt(Sibling) == "{[-1]:Pointer, [-1,8]:Float@double}"
```
vchuravy, line 53: @wsmoses this looks wrong to me, but is the same answer that Enzyme currently gives.

wsmoses: The reason it appears right is that unions should intersect the inner types, which here would be nothing I presume. This is because TT's don't represent unions but represent a guarantee that this offset is always type x

src/compiler.jl
```diff
@@ -3060,7 +3062,7 @@ function enzyme!(job, mod, primalf, TT, mode, width, parallel, actualRetType, wr
     logic = Logic()
     TA = TypeAnalysis(logic, rules)
 
-    retTT = typetree((!isa(actualRetType, Union) && GPUCompiler.deserves_retbox(actualRetType)) ? Ptr{actualRetType} : actualRetType, ctx, dl)
+    retTT = typetree((!isa(actualRetType, Union) && GPUCompiler.deserves_retbox(actualRetType)) ? Ptr{actualRetType} : actualRetType, ctx, dl, seen)
```
vchuravy, line 3065:
```suggestion
    retT = (!isa(actualRetType, Union) && GPUCompiler.deserves_retbox(actualRetType)) ? Ptr{actualRetType} : actualRetType
    retTT = typetree(retT, ctx, dl, seen)
```

wsmoses (approved):

src/rules/typerules.jl
```diff
@@ -116,3 +116,110 @@ function alloc_rule(direction::Cint, ret::API.CTypeTreeRef, args::Ptr{API.CTypeT
     end
     return UInt8(false)
 end
+
+function julia_type_rule(direction::Cint, ret::API.CTypeTreeRef, args::Ptr{API.CTypeTreeRef}, known_values::Ptr{API.IntList}, numArgs::Csize_t, val::LLVM.API.LLVMValueRef)::UInt8
```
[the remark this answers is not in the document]
wsmoses: Ah yes we did (it is now passed by fn attributes) which is also more stable as args get eliminated

Issue #776, opened 2023-04-24
Something invalidating `Enzyme.__init__` causes a heavy package load time penalty


KristofferC:
When loading OmniPackage.jl(https://github.com/JuliaComputing/OmniPackage.jl/) a big part of its load time (around 10s) is spent compiling and running the `__init__` method in Enzyme, profile available at https://kristofferc.github.io/tracy-traces/omnipackage/.

Ideally, Enzyme would somehow protect itself against this, by either making the code in `__init__` less vulnerable to invalidations or some other way (maybe using a frozen world age).

cc @vchuravy

wsmoses:
This should now be resolved by https://github.com/EnzymeAD/Enzyme.jl/pull/881

vchuravy:
I am not sure that fixed all instances. Need to confirm with an actual profile

KristofferC:
I just want to point out that this is still happening and causes ~18s load time increase to any package that recursively ends up getting Enzyme as a dependency.

wsmoses:
hm @KristofferC do you have any information regarding what could be causing this?

KristofferC:
- Someone adds a method that invalidates some function called in `Enzyme.__init__`.
- `Enzyme.__init__` in invalidated and has to be recompiled at load time.
- Compiling `Enzyme.__init__` takes about 18 seconds.

wsmoses:
is there a tool or something to profile what methods/invalidations are the source?

KristofferC:
One would usually use SnoopCompile. I can try (later) to see if I can find what actually ends up invalidating it.

KristofferC:
```
julia> @time using Tracker
1.450969 seconds (1.83 M allocations: 109.970 MiB, 7.33% gc time, 31.56% compilation time: 26% of which was recompilation)

julia> @time using Enzyme
 21.592203 seconds (43.97 M allocations: 2.517 GiB, 1.84% gc time, 98.84% compilation time: 100% of which was recompilation)
```

is some sort of MWE. Will narrow it down more.

KristofferC:
Okej, Tracker committed the mortal sin of defining `show` on a `Type`. With https://github.com/FluxML/Tracker.jl/pull/156 this is improved to

```
julia> @time using Tracker
  1.463012 seconds (1.90 M allocations: 113.004 MiB, 4.50% gc time, 25.19% compilation time: 16% of which was recompilation)

julia> @time using Enzyme
[...]

PR #1303, opened 2024-02-21
Explicitly precompile register_llvm_rules

vchuravy:
Reduces loading time on my system significantly from ~30s to 10s

### DIFF
--- a/src/rules/llvmrules.jl
+++ b/src/rules/llvmrules.jl
@@ -1155,7 +1155,7 @@ macro fwdfunc(f)
     ))
 end
 
-@inline function register_llvm_rules()
+@noinline function register_llvm_rules()
     register_handler!(
         ("julia.call",),
         @augfunc(jlcall_augfwd),
@@ -1361,3 +1361,5 @@ end
         @fwdfunc(jl_unhandled_fwd),
     )
 end
+
+precompile(register_llvm_rules, ())


vchuravy:
Fixes #776

KristofferC:
#776 is mostly about the time it takes if anything in here is invalidated. When that issue was opened the package was just as quick to load on its own as shown here, it was in combination with other packages it got slow.

vchuravy:
Is that still the case? I could additionally break the call edge by using invokelatest.

KristofferC:
```
julia> @time using Enzyme
  0.231514 seconds (259.79 k allocations: 20.705 MiB, 3.36% gc time, 0.88% compilation time)

######

julia> struct T end

julia> Base.show(io::IO, t::Type{T}) = print(io, "heh")

julia> @time using Enzyme
 10.714881 seconds (26.53 M allocations: 1.836 GiB, 5.36% gc time, 97.91% compilation time: 26% of which was recompilation)
```

KristofferC:
To me, something like https://github.com/KristofferC/Enzyme.jl/commit/4246ddb42b18ef43958774b13621f6160f7992a8#diff-21145f7299f4174cfb41f7c0c845a0ee798fc065bae0a352171552dc4b810508R1033 feels correct it just has to be ensured that the actual callbacks themselves that end up getting run during runtime gets properly optimized (but not the code that generates and inserts them).

PR #1305, opened 2024-02-23
run `register_llvm_rules` in a module where optimizations are disabled

### LINKED ISSUE #776: Something invalidating `Enzyme.__init__` causes a heavy package load time penalty

vchuravy:
fixes #776

### DIFF
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -2450,7 +2450,7 @@ function __init__()
     API.EnzymeSetUndefinedValueForType(@cfunction(
                                             julia_undef_value_for_type, LLVM.API.LLVMValueRef, (LLVM.API.LLVMTypeRef,UInt8)))
     register_alloc_rules()
-    register_llvm_rules()
+    @time LLVMRules.register_llvm_rules()
 end
 
 # Define EnzymeTarget
--- a/src/rules/llvmrules.jl
+++ b/src/rules/llvmrules.jl
@@ -1018,6 +1018,7 @@ end
 
 function register_handler!(variants, augfwd_handler, rev_handler, fwd_handler=nothing)
     for variant in variants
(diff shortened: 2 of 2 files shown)


wsmoses (commented):
I am worried about this as these runtime functions make up a non negligible part of the adjoint generation runtime

vchuravy:
I still need to confirm, but iirc this should only apply to the ccallable thunk.

vchuravy:
Superseded by #3531.

Measured this on today's tree: as written this PR now recovers 5s of the 22s available. The mechanism moved out from under it. In 2023 `@augfunc` built an anonymous closure *inside* the low-optimization module, so `@compiler_options` covered the actual `@cfunction` target. Today `@register_aug`/`_rev`/`_fwd`/`_diffuse` define named `<rule>_cfunc` shims in `Compiler`, which stay fully optimized and `@inline` the rule bodies, so moving only the `@cfunction` call site leaves all 115 rule bodies in codegen.

#3531 moves the shim definitions as well: precompile goes 47.9s -> 37.8s as shipped, or 33.3s -> 11.1s with the `@compile_workload` disabled.

On the concern about adjoint generation runtime -- it is real but small. A deliberately generic-dispatch-heavy differentiation goes 5.18s -> 5.42s; scalar and array gradients do not move, and neither does running an already-compiled gradient.

PR #1306, opened 2024-02-24
Fix triangular solve rule for `Adjoint`s

simsurace:
When wrapping an `Adjoint`, the previous rule errored.

### DIFF
--- a/test/internal_rules.jl
+++ b/test/internal_rules.jl
@@ -409,7 +409,9 @@ end
                 test_reverse(f!, Const, (Y, TY), (M, TM), (B, TB), (_A, Const))
             end
         end
-        @testset "regression test for #1306" begin
+        @testset "test through `Adjoint` wrapper (regression test for #1306)" begin
+            # Test that we get the same derivative for `M` as for the adjoint of its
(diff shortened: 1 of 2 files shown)


simsurace:
I reverted the changes in the `src` to disentangle the failures. There are some problems on `main` still.

simsurace:
Dang, I just made the same mistake, we can't use EnzymeTestUtils this way.

simsurace:
[67e0260](https://github.com/EnzymeAD/Enzyme.jl/pull/1306/commits/67e0260faf0d166771a5252f7f9d7d96805117ce) documents what is broken on main.

test/internal_rules.jl
```diff
@@ -409,6 +409,24 @@ end
                 test_reverse(f!, Const, (Y, TY), (M, TM), (B, TB), (_A, Const))
             end
         end
+        @testset "regression test for #1306" begin
```
wsmoses, line 412: Can you give this a nicer name?

simsurace: Done. I also added two lines of explanation. Let me know if you need more.

wsmoses (approved):

PR #1307, opened 2024-02-25
Fix rules for `cholesky` and `ldiv!` on `Cholesky`

simsurace:
I had a suspicion that the existing rule was somehow off. I managed to break it with something like this:
```julia
function f(A)
    C = cholesky(A * A')
    return sum(abs2, C.L * C.U)
end
A = rand(3, 3)
test_reverse(f, Active, (A, Duplicated))
```
Working on a fix...


simsurace:
I went back to ChainRules.jl and copied the rule from there. As noted previously, the rules looked completely different before the change, with the Enzyme rule just accumulating the `factors` field of the cotangent, while the ChainRules rule did a sequence of linear algebra operations. Now with the change the new test passes, but some of the old tests are broken. @michel2323 could you double check whether the tests are correct?

simsurace:
@michel2323 I think the rules are now fixed. I hope you did not mind me reformatting your tests.
@wsmoses if @michel2323 does not respond, we might need another reviewer.
Maybe @sethaxen could also help with this rule?

src/internal_rules.jl
```diff
@@ -764,7 +764,7 @@ function EnzymeRules.augmented_primal(
     cache = if isa(A, Const)
         nothing
     else
-        dfact
+        (fact, dfact)
```
wsmoses, line 767: fact is not always computed (e.g. if `EnzymeRules.needs_primal(config)` is false). Can you change fact to also be computed if `!isa(A, Const)` and also then have a `fact_returned = EnzymeRules.needs_primal(config) ? fact : nothing` which is the returned factors

wsmoses: Also fact may be overwritten from forward to reverse so you should probably copy fact for the cache

test/internal_rules.jl
```diff
@@ -384,8 +317,43 @@ end
         Enzyme.autodiff(Reverse, h, Active, Duplicated(A, dA), Const(b))
 
         dA_sym = - (transpose(A) \ [1.0, 0.0]) * transpose(A \ b)
+        dA_sym = (dA_sym + dA_sym') / 2
         @test isapprox(dA, dA_sym)
     end
+    @testset "Regression test for #1307" begin
```
wsmoses, line 323: Can you give this a nicer name?

src/internal_rules.jl
simsurace: Thanks, need to look into this. This is not covered by any test utilities?

michel2323:
> @michel2323 I think the rules are now fixed. I hope you did not mind me reformatting your tests. @wsmoses if @michel2323 does not respond, we might need another reviewer. Maybe @sethaxen could also help with this rule?

Go ahead. I went over it now. Looks good to me. I never could fully match the ChainRules implementation to this [paper](https://arxiv.org/abs/1602.07527). But that may just be me. Thanks for taking a look at this!

simsurace:
I did not look at the paper in detail. Maybe their method is faster than the current implementation, which matches ChainRules.jl. We can always improve things later, in my opinion it's more important to merge a correct rule now. 
Maybe some of these rules will even become obsolete when Enzyme capabilities improve to also be able to autodiff through LAPACK routines. Thanks for taking a look!

test/internal_rules.jl
simsurace: Done

simsurace:
@wsmoses, I tried to follow your guidance on the `fact`, let me know if I got it right.
I just noticed that I did not check the forward rule in the regression test, and in fact it also needs to be changed.
Please let me fix that before merging so we get everything right this time.

src/internal_rules.jl
simsurace: Let me know if I got it right.

simsurace:
Ok, the additional forward unit tests for `cholesky` all fail, documenting there's still something wrong with the forward rule.

simsurace:
I noticed there are still errors. Please don't merge just yet.

simsurace:
Now we should probably refactor to target `cholesky!` instead of `cholesky`.

sethaxen:
Agreed, also I believe this rule assumes `NoPivot`, so you probably want to just target this method: https://github.com/JuliaLang/julia/blob/688ae0a58361889feedcff9eeb5ff1e1b710906b/stdlib/LinearAlgebra/src/cholesky.jl#L267-L271. You perhaps can expand the wrapped type from `Matrix` to `StridedMatrix`, and you could expand to `RealHermSymComplexHerm`.

simsurace:
The issue is that currently [this](#1320) fails, so the easiest would be to write two rules, one for `cholesky` and one for `cholesky!`, but I'd rather first merge this in a correct state before further expanding it. Without some better tooling the effort of writing these rules is not sustainable for me.

wsmoses:
The in place reverse mode cholesky rule is going to be harder because the out of place one from Michel makes an assumption about the aliasing usage of the shadow (in essence it stuffs the derivative somewhere it shouldn't be, then uses it to set the correct result in a later stage).

This should be fixed to just do the proper thing, but FYI.

@sethaxen would you have cycles to work with @simsurace to set up the tooling they (and others) need?

wsmoses:
And yeah @simsurace you boldly (and successfully) started with some of the most difficult rules to do -- which have been left as todos for this reason!

simsurace:
> The in place reverse mode cholesky rule is going to be harder because the out of place one from Michel makes an assumption about the aliasing usage of the shadow (in essence it stuffs the derivative somewhere it shouldn't be, then uses it to set the correct result in a later stage).

Is this still true of this PR though? I pretty much changed every line by now. Could you indicate the questionable lines?

sethaxen:
> Without some better tooling the effort of writing these rules is not sustainable for me.

> @sethaxen would you have cycles to work with @simsurace to set up the tooling they (and others) need?

Spare cycles are a bit rare right now, but maybe I can do something. Which tooling is needed?

 > The issue is that currently [this](#1320) fails, so the easiest would be to write two rules, one for `cholesky` and one for `cholesky!`,

I didn't quite follow, could you explain why two rules are needed? I think `cholesky` for the base types effectively just copies and then calls `cholesky!`.

simsurace:
See #1320, i.e. writing a rule for `cholesky!` won't be sufficient because the stuff in `cholcopy` fails to differentiate.

simsurace:
I would suggest merging this now, keeping the focus to the correctness, and making the `cholesky!` rule in another PR.

As for tooling, two things would be helpful in my view: 1) making it so EnzymeTestUtils can be used to completely check the correctness of any rules, including those with wrapper types etc. (this means fixing some of the issues which are inherited from FiniteDifferences such as the treatment of `to_vec`). 2) developing some tooling to reduce the boilerplate in rules, i.e. dealing with caches, which seem to be pretty independent of the actual mathematical content, such that rule writers can focus more on the latter. As a side benefit, I think the rules will also get more readable and more uniformly styled.

sethaxen:
> See #1320, i.e. writing a rule for `cholesky!` won't be sufficient because the stuff in `cholcopy` fails to differentiate.

Seems like that's an Enzyme issue that would need to be fixed either way.

> As for tooling, two things would be helpful in my view: 1) making it so EnzymeTestUtils can be used to completely check the correctness of any rules, including those with wrapper types etc. (this means fixing some of the issues which are inherited from FiniteDifferences such as the treatment of `to_vec`).

I'll take a pass at this. It might be easier than I'm thinking.

> 2) developing some tooling to reduce the boilerplate in rules, i.e. dealing with caches, which seem to be pretty independent of the actual mathematical content, such that rule writers can focus more on the latter. As a side benefit, I think the rules will also get more readable and more uniformly styled.

I agree this would be great, but I unfortunately cannot help with this. I suspect doing this correctly would require being able to identify what intermediates need to be cached and the best way to do this just by code analysis, which actually seems like a pretty complex problem to solve.

simsurace:
> > See #1320, i.e. writing a rule for `cholesky!` won't be sufficient because the stuff in `cholcopy` fails to differentiate.
> 
> Seems like that's an Enzyme issue that would need to be fixed either way.
> 
> > As for tooling, two things would be helpful in my view: 1) making it so EnzymeTestUtils can be used to completely check the correctness of any rules, including those with wrapper types etc. (this means fixing some of the issues which are inherited from FiniteDifferences such as the treatment of `to_vec`).
> 
> I'll take a pass at this. It might be easier than I'm thinking.
> 
> > 2. developing some tooling to reduce the boilerplate in rules, i.e. dealing with caches, which seem to be pretty independent of the actual mathematical content, such that rule writers can focus more on the latter. As a side benefit, I think the rules will also get more readable and more uniformly styled.
> 
> I agree this would be great, but I unfortunately cannot help with this. I suspect doing this correctly would require being able to identify what intermediates need to be cached and the best way to do this just by code analysis, which actually seems like a pretty complex problem to solve.

Ok, I do not expect anyone in particular to do this, but it would definitely make life easier.

simsurace:
> FWIW, I think the safest way to go is to only define the rules only for `LAPACK.potrf!`, which are necessary until Enzyme can differentiate through LAPACK code. I don't much like covering up https://github.com/JuliaLang/julia/blob/892c49162a48e9646dc3f0f2b94ebb33ae2a76be/stdlib/LinearAlgebra/src/cholesky.jl#L196-L251 for the following reasons:
> 
>     1. It shouldn't be necessary. Enzyme should be able to AD through those methods.
> 
>     2. The "right" way to implement a Cholesky rule is to by hand differentiate the same path the primal takes. The primal function makes no assumption of the primal matrix being Hermitian; failure is an allowed option (and does not error if `check=false`). The rules that use linear algebra assume no failure and so will be wrong if the Cholesky factorization failed. For LAPACK we have no choice, but if we had the option to differentiate the LAPACK code directly, I would always recommend that.

I agree, but I think there is a tradeoff and I'd rather merge this rule if it is correct, because it unlocks many applications and can boost the adoption of Enzyme. 

I would rather prefer not having to write more rules for LAPACK routines, but I have no idea how to make it so Enzyme can differentiate them natively, nor how difficult that is.

sethaxen:
I agree that since this fixes a correctness issue, it should be merged soon. But I also think that after the `cholmod` issue is fixed, it should be replaced with a mutating version, ideally one that acts just on `potrf!`. I'll try to review this evening.

I have a functioning `to_vec` implementation in #1327

src/internal_rules.jl
```diff
@@ -761,47 +775,73 @@ function EnzymeRules.augmented_primal(
...
         end
     end
-    cache = if isa(A, Const)
-        nothing
-    else
-        dfact
-    end
 
+    cache = isa(A, Const) ? nothing : (fact_returned, dfact)
     return EnzymeRules.AugmentedReturn(fact, dfact, cache)
```
wsmoses, line 780: this should now be fact_returned (aka only returned if EnzymeRules.needs_primal)

src/internal_rules.jl
```diff
@@ -742,11 +758,9 @@ function EnzymeRules.augmented_primal(
     RT::Type,
     A::Annotation{<:Union{Matrix,LinearAlgebra.RealHermSym{<:Real,<:Matrix}}};
     kwargs...)
-    fact = if EnzymeRules.needs_primal(config)
-        cholesky(A.val; kwargs...)
-    else
-        nothing
-    end
+
+    fact = cholesky(A.val; kwargs...)
```
wsmoses, line 762: This should be `(EnzymeRules.needs_primal(config) || !(RT <: Const)) ? cholesky(...) : nothing` (aka only creating it if we would use it

src/internal_rules.jl
```diff
@@ -761,47 +775,73 @@ function EnzymeRules.augmented_primal(
             end
         end
     end
-    cache = if isa(A, Const)
-        nothing
-    else
-        dfact
-    end
 
+    cache = isa(A, Const) ? nothing : (fact_returned, dfact)
```
wsmoses, line 779: This one should just be fact (aka this is not the returned one, which is a better name for the one below).

wsmoses (approved):
If I'm totally honest, I'm not 100% followng the math (some comments saying what's happening there/why may be nice), but I've added some comments about conventions

simsurace:
Please add your questions as comments to specific lines and I'll try to answer to the best of my knowledge.

src/internal_rules.jl
```diff
@@ -761,47 +775,73 @@ function EnzymeRules.augmented_primal(
...
         dfacts = EnzymeRules.width(config) == 1 ? (dfact,) : dfact
 
         for (dA, dfact) in zip(dAs, dfacts)
             _dA = dA isa LinearAlgebra.RealHermSym ? dA.data : dA
             if _dA !== dfact.factors
-                _dA .+= dfact.factors
+                Ā = _cholesky_pullback_shared_code(fact, dfact)
+                idx = diagind(Ā)
+                @views Ā[idx] .= real.(Ā[idx]) ./ 2
+                _dA .+= UpperTriangular(Ā)
```
sethaxen, line 801: I don't think any of this should be here when `A` is a `Matrix`, but this doesn't look right to me for `RealHermSym`. I would expect something like
```suggestion
                _dA .+= LowerTriangular(Ā) + UpperTriangular(Ā)' - real(Diagonal(Ā))
```
(which in ChainRules would be the pullback of the `Symmetric` constructor), but I would move that to `_cholesky_pullback_shared_code`.

test/internal_rules.jl
```diff
@@ -379,12 +312,91 @@ end
...
         dA_sym = - (transpose(A) \ [1.0, 0.0]) * transpose(A \ b)
-        @test isapprox(dA, dA_sym)
+        @test isapprox((dA + dA') / 2, (dA_sym + dA_sym') / 2)
+    end
+    @testset "Unit test for `cholesky` (regression test for #1307)" begin
+        # This test checks the `cholesky` rules without involving `ldiv!`
+        function f(A)
+            C = cholesky(A * adjoint(A))
+            return sum(abs2, C.L * C.U)
+        end
```
sethaxen, line 326: This would be a more thorough test.
```suggestion
        function f!(Y, A)
            C = cholesky(A * adjoint(A))
            copyto!(Y, C.factors)
            return nothing
        end
```
Do you have a test somewhere that any differentials for the unused triangle of `factors` are correctly pulled back?

test/internal_rules.jl
```diff
@@ -379,12 +312,91 @@ end
...
+        @testset for TE in (Float64, ComplexF64)
+            A = rand(TE, 3, 3)
```
sethaxen, line 328: Let's make this a little harder.
```suggestion
            A = rand(TE, 5, 5)
```

test/internal_rules.jl
```diff
@@ -379,12 +312,91 @@ end
...
+                    end
+
+                    @testset "Reverse mode" begin
+                        dA1 = zero(A)
+                        dA2 = zero(A)
+                        autodiff(Reverse, f, Active, Duplicated(A, dA1))
+                        autodiff(Reverse, g, Active, Duplicated(A, dA2))
+                        @test dA1 ≈ dA2
+                    end
+                end
```
sethaxen, line 351: IMO this is unnecessary. The EnzymeTestUtils checks are more than sufficient, and, with the function proposed above, are more thorough.
```suggestion
```

test/internal_rules.jl
```diff
@@ -379,12 +312,91 @@ end
...
+                if TE == Float64
```
sethaxen, line 352: Same, I would simplify these to just call `test_forward` and `test_reverse` using the combinations `[(Symmetric, Float64), (Hermitian, Float64), (Hermitian, ComplexF64)]`.

src/internal_rules.jl
```diff
@@ -672,6 +664,30 @@ function EnzymeRules.forward(::Const{typeof(cholesky)}, RT::Type, A; kwargs...)
     end
 end
 
+function _cholesky_forward(C::Cholesky, Σdot)
+    # Computes the cholesky forward mode update rule
+    # C.f. eq. 8 in https://arxiv.org/pdf/1602.07527.pdf
+    if C.uplo == 'U'
```
sethaxen, line 670:
```suggestion
    if C.uplo === 'U'
```

src/internal_rules.jl
```diff
@@ -672,6 +664,30 @@ function EnzymeRules.forward(::Const{typeof(cholesky)}, RT::Type, A; kwargs...)
     end
 end
 
+function _cholesky_forward(C::Cholesky, Σdot)
```
sethaxen, line 667: I recommend keeping notation the same as elsewhere
```suggestion
function _cholesky_forward(C::Cholesky, Ȧ)
```

src/internal_rules.jl
```diff
@@ -672,6 +664,30 @@ function EnzymeRules.forward(::Const{typeof(cholesky)}, RT::Type, A; kwargs...)
...
+    # C.f. eq. 8 in https://arxiv.org/pdf/1602.07527.pdf
+    if C.uplo == 'U'
+        U = C.U
+        Udot = Σdot / U
+        ldiv!(U', Udot)
+        idx = diagind(Udot)
+        Udot[idx] ./= 2
+        triu!(Udot)
+        rmul!(Udot, U)
+        return Cholesky(Udot, 'U', 0)
```
sethaxen, line 678: Notational consistency. Also, we should probably keep the same `.info` as the primal.
```suggestion
        U̇ = Ȧ / U
        ldiv!(U', U̇)
        idx = diagind(U̇)
        U̇[idx] ./= 2
        triu!(U̇)
        rmul!(U̇, U)
        return Cholesky(U̇, 'U', C.info)
```

src/internal_rules.jl
```diff
@@ -672,6 +664,30 @@ function EnzymeRules.forward(::Const{typeof(cholesky)}, RT::Type, A; kwargs...)
...
+    else
+        L = C.L
+        Ldot = L \ Σdot
+        rdiv!(Ldot, L')
+        idx = diagind(Ldot)
+        Ldot[idx] ./= 2
+        tril!(Ldot)
+        lmul!(L, Ldot)
+        return Cholesky(Ldot, 'L', 0)
```
sethaxen, line 687:
```suggestion
        L̇ = L \ Ȧ
        rdiv!(L̇, L')
        idx = diagind(L̇)
        L̇[idx] ./= 2
        tril!(L̇)
        lmul!(L, L̇)
        return Cholesky(L̇, 'L', C.info)
```

src/internal_rules.jl
sethaxen: I don't think this method is correct for real `Symmetric` or `Hermitian` `A`, since it assumes that `Ȧ` represents the Hermitian tangent of the Hermitian matrix, but in such cases, `Ȧ` is just a container for the tangent of `A.data`. More generally, this rule can be wrong for many structured types that represent Hermitian matrices. I recommend restricting the type of `A` in `forward` and making sure this is thoroughly tested.

sethaxen (changes requested):
I focused on the linear algebra and tests and did not look at the boilerplate too closely.

src/internal_rules.jl
simsurace: Do you have a test for this? I'm not sure I understand why the constructor should enter (see also the recent Slack discussion for `sum` [here](https://julialang.slack.com/archives/C01J3R4FHB3/p1709555032435419)).

test/internal_rules.jl
simsurace: I don't think EnzymeTestUtils can cover all the cases with wrappers. Also, this was a regression test to protect against the bug that prompted me to open this PR. It looks dangerous to me to remove this.

test/internal_rules.jl
simsurace: That doesn't work IIRC because of issues with FiniteDiff.

src/internal_rules.jl
simsurace: You may be right, I need to revisit that after https://julialang.slack.com/archives/C01J3R4FHB3/p1709555032435419.

src/internal_rules.jl
sethaxen: When working out AD rules, we generally treat tangents as dense arrays (i.e. we use the same semantics as ChainRules), and that's what the function `_cholesky_forward` is doing. But after doing that, you need to convert to Enzyme's semantics. That's effectively the ChainRules-style pullback of the constructor (which is what this code does).

src/internal_rules.jl
```diff
@@ -761,47 +775,73 @@ function EnzymeRules.augmented_primal(
...
+                @views Ā[idx] .= real.(Ā[idx]) ./ 2
+                _dA .+= UpperTriangular(Ā)
                 dfact.factors .= 0
             end
         end
     end
     return (nothing,)
 end
 
+# Taken from ChainRules.jl
```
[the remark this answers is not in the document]
simsurace: Done

src/internal_rules.jl
simsurace: Making the suggested change breaks a bunch of tests (the original test that was here before this PR, as well as its modified version [here](https://github.com/simsurace/Enzyme.jl/blob/33e48afdd7ea345f69218ba6630d4a36f337e7c7/test/internal_rules.jl#L319), and [this testset](https://github.com/simsurace/Enzyme.jl/blob/33e48afdd7ea345f69218ba6630d4a36f337e7c7/test/internal_rules.jl#L331)). While I'm not completely sure of the former, I'm close to certain that the latter test is correct because it compares differentiating two functions that are mathematically equivalent (both representing `A -> sum(abs2, A * A')`), one of which passes through `cholesky`, and therefore shouldn't break under any circumstance.

src/internal_rules.jl
simsurace: Do you have a test that shows that this is wrong?

simsurace:
@sethaxen I wrote some tests from scratch using EnzymeTestUtils. They have the known limitations. 
I'm not sure that is better than the previous situation. We still have some disagreements about the forward rules, and no test to decide whether the implementation is correct. The current implementation however seems to give the right results in all applications I've tried. I hope we can find a way to move forward with this.

sethaxen:
@simsurace EnzymeTestUtils v0.1.7 removes most of the limitations of the testing functions, so that should allow some of the tests here to be simplified (i.e. no problem with Hermitian arguments and Cholesky outputs). This week I'll look closely at the points we disagreed on.

simsurace:
That's great, looking forward to it.

sethaxen:
I just tested this PR locally with the following, and both forward- and reverse-mode tests fail:

```julia
julia> @testset "cholesky" begin
           @testset for Te in (Float64,), TS in (Symmetric, Hermitian)
               @testset for TA in (Const, Duplicated), Tret in (Const, Duplicated)
       
                   A = exp(TS(rand(Te, 4, 4)))
                   are_activities_compatible(Tret, TA) || continue
                   test_forward(cholesky, Tret, (A, TA))
                   test_reverse(cholesky, Tret, (A, TA))
               end
           end
       end
```

simsurace:
Ok, I was able to fix the tests suggested above. I removed all previous tests and made new tests entirely with EnzymeTestUtils. Now, the remaining failures are from the forward rule of `ldiv!` on `Cholesky`, which I hadn't touched yet. I will address this later.

test/internal_rules.jl
```diff
@@ -136,327 +157,101 @@ end
...
+            M = rand(TE, n, n)
+            B = rand(TE, sizeB...)
+            Y = zeros(TE, sizeB...)
+            A = T(M)
+            @testset "test through constructor" begin
+                _A = T(A)
+                function f!(Y, A, B, ::T) where {T}
+                    ldiv!(Y, T(A), B)
+                    return nothing
+                end
```
sethaxen, line 219: EnzymeTestUtils can now handle this, and it's slightly better because it also checks that the rule and primal return the same quantity.

```suggestion
                f!(Y, A, B, ::T) where {T} = ldiv!(Y, T(A), B)
```

test/internal_rules.jl
```diff
@@ -136,327 +157,101 @@ end
...
+                for TY in (Const, Duplicated, BatchDuplicated),
+                    TM in (Const, Duplicated, BatchDuplicated),
+                    TB in (Const, Duplicated, BatchDuplicated)
+
+                    are_activities_compatible(Const, TY, TM, TB) || continue
+                    test_reverse(f!, Const, (Y, TY), (M, TM), (B, TB), (_A, Const))
```
sethaxen, line 225:
```suggestion
                    test_reverse(f!, TY, (Y, TY), (M, TM), (B, TB), (_A, Const))
```

sethaxen (changes requested):

test/internal_rules.jl
simsurace: Those are good points, but they are unrelated to this PR. The only reason they end up as changed lines is because I ran the formatter on these files. I'm going to open another PR for this https://github.com/EnzymeAD/Enzyme.jl/pull/1407

test/internal_rules.jl
simsurace: https://github.com/EnzymeAD/Enzyme.jl/pull/1407

sethaxen:
FWIW, the forward-rule for `ldiv!(F::Cholesky, B::AbstractVecOrMat)` is:

```julia
function ldiv_Cholesky_forward!(L, U, B, dL, dU, dB)
    ldiv!(L, B)
    mul!(dB, dL, B, -1, 1)
    ldiv!(L, dB)
    ldiv!(U, B)
    mul!(dB, dU, B, -1, 1)
    ldiv!(U, dB)
    return B, dB
end
```
LAPACK xpotrs does nothing more than call two triangular solves with the triangular factors (see e.g. https://netlib.org/lapack/explore-html-3.6.1/d1/d7a/group__double_p_ocomputational_ga167aa0166c4ce726385f65e4ab05e7c1.html), which is what this code ultimately does. But having access to the result of the first triangular solve before overwriting it allows to reduce some computation.

simsurace:
Thanks! I'll try it out ASAP.

simsurace:
I haven't yet gotten it right. Apparently for some reason the rule is not called when one of the input activities is `Const` (the "Hi from forward..." message does not appear), and when all activities are `Duplicated`, the computed derivative is different from the finite differences one.

simsurace:
This is puzzling to me. When I make the call
```julia
Te = Float64
A = exp(Symmetric(rand(Te, 4, 4)))
B = rand(Te, 4)
C = cholesky(A)

dB = ones(Te, 4)
dC = make_zero(C)

f(x, y) = ldiv!(x, y)
autodiff(Forward, f, Const, Const(C), Duplicated(B, dB))
```
the forward rule is used and everything seems to work just fine.
However, when doing
```julia
test_forward(f, Const, (C, Const), (B, Duplicated))
```
this ends up doing an autodiff call that does not use the rule, or at least an error is thrown before entering the rule, as can be confirmed by putting a log message at the beginning of the rule. @sethaxen do you have any idea what may be going on?
The rule seems to be correct when all activities are `Duplicated`, it's just the tests with some arguments being `Const` that behave weirdly.

simsurace:
Interesting. I always thought that this is allowed. The following also works:
```julia
julia> function f!(x, y)
           y .= x .* y
           return y
       end
f! (generic function with 1 method)

julia> x = rand(4); y = rand(4); dy = ones(4);

julia> autodiff(Forward, f!, Const, Const(x), Duplicated(y, dy))
()

julia> dy ≈ x
true
```

simsurace:
I would propose to mark the questionable activity combinations as broken and open a separate issue about this, as it seems to be a more general question, in order to finally get this merged.

sethaxen:
> This is puzzling to me. When I make the call
> 
> ```julia
> Te = Float64
> A = exp(Symmetric(rand(Te, 4, 4)))
> B = rand(Te, 4)
> C = cholesky(A)
> 
> dB = ones(Te, 4)
> dC = make_zero(C)
> 
> f(x, y) = ldiv!(x, y)
> autodiff(Forward, f, Const, Const(C), Duplicated(B, dB))
> ```
> 
> the forward rule is used and everything seems to work just fine. However, when doing
> 
> ```julia
> test_forward(f, Const, (C, Const), (B, Duplicated))
> ```
> 
> this ends up doing an autodiff call that does not use the rule, or at least an error is thrown before entering the rule, as can be confirmed by putting a log message at the beginning of the rule. @sethaxen do you have any idea what may be going on? The rule seems to be correct when all activities are `Duplicated`, it's just the tests with some arguments being `Const` that behave weirdly.

To capture keywords correctly, EnzymeTestUtils actually does the following:

```julia
fkwargs = (;)
call_with_kwargs(f, xs...) = f(xs...; fkwargs...);
autodiff(Forward, call_with_kwargs, Const, Const(f), Const(C), Duplicated(B, dB))
```
This raises the activity error you're seeing. In fact, dropping `fkwargs` also does. It seems the issue is making `f` an argument. You could open an issue about this. Here the error message is informative:
```
test_forward: f with return activity Const on Const{Cholesky{Float64, Matrix{Float64}}}, (::Vector{Float64}, Duplicated): Error During Test at /home/sethaxen/.julia/packages/EnzymeTestUtils/5JRuz/src/test_forward.jl:70
  Enzyme execution failed.
  Mismatched activity for:   store {} addrspace(10)* %.fca.0.0.extract, {} addrspace(10)** %.fca.0.0.gep, align 8, !tbaa !19, !alias.scope !23, !noalias !26 const val:   %.fca.0.0.extract = extractvalue { {} addrspace(10)*, i32, i64 } %0, 0
  Type tree: {[-1]:Pointer, [-1,0]:Pointer, [-1,0,-1]:Float@double, [-1,8]:Integer, [-1,9]:Integer, [-1,10]:Integer, [-1,11]:Integer, [-1,12]:Integer, [-1,13]:Integer, [-1,14]:Integer, [-1,15]:Integer, [-1,16]:Integer, [-1,17]:Integer, [-1,18]:Integer, [-1,19]:Integer, [-1,20]:Integer, [-1,21]:Integer, [-1,22]:Integer, [-1,23]:Integer, [-1,24]:Integer, [-1,25]:Integer, [-1,26]:Integer, [-1,27]:Integer, [-1,28]:Integer, [-1,29]:Integer, [-1,30]:Integer, [-1,31]:Integer, [-1,32]:Integer, [-1,33]:Integer, [-1,34]:Integer, [-1,35]:Integer, [-1,36]:Integer, [-1,37]:Integer, [-1,38]:Integer, [-1,39]:Integer}
  You may be using a constant variable as temporary storage for active memory (https://enzyme.mit.edu/julia/stable/faq/#Activity-of-temporary-storage). If not, please open an issue, and either rewrite this variable to not be conditionally active or use Enzyme.API.runtimeActivity!(true) as a workaround for now
```

So do
```julia
Enzyme.API.runtimeActivity!(true)
... # tests
Enzyme.API.runtimeActivity!(false)
```
There are some examples of this in `test/runtests.jl`. You'll need to do this before calling `test_forward` for this function. Otherwise you'll have to restart Enzyme.

```quote
I would propose to mark the questionable activity combinations as broken and open a separate issue about this, as it seems to be a more general question, in order to finally get this merged.
```

This example seems to fail with incorrect derivative:
```julia
test_forward(f, Duplicated, (C, Const), (B, Duplicated))
```

simsurace:
> This example seems to fail with incorrect derivative:
>```julia
>test_forward(f, Duplicated, (C, Const), (B, Duplicated))
>```
Hmm, this does still not make sense to me. When I evaluate the autodiff call (outside of EnzymeTestUtils), the result is exactly the same whether `C` is `Const` or not, and the test passes when it is `Const`. The derivative of what exactly is wrong?

simsurace:
So I put an `@info "Strange"` in the branch that should only be visited when `!isa(C, Const)`. But then
```julia
julia> test_forward(f, Duplicated, (C, Const), (B, Duplicated))
[ Info: Strange
test_forward: f with return activity Duplicated on (::Cholesky{Float64, Matrix{Float64}}, Const), (::Vector{Float64}, Duplicated): Test Failed at /Users/simone/.julia/packages/EnzymeTestUtils/5JRuz/src/test_approx.jl:7
  Expression: isapprox(x, y; kwargs...)
  Problem: derivative should agree with finite differences
   Evaluated: isapprox([6.140528054645173, 5.595696177579358, -9.846857611815441, -5.902596380301342], [5.842892339729809, 5.519492606061886, -8.640493849066678, -5.669506374510098]; atol = 1.0e-9, rtol = 1.0e-9)
...
```
So somehow the activity of `C` seems to be changed in EnzymeTestUtils. Could it be an off-by-one error in indexing the activities due to the fact that the argument `f` is inserted for the call with keywords?

sethaxen:
> > This example seems to fail with incorrect derivative:
> >```julia
> >test_forward(f, Duplicated, (C, Const), (B, Duplicated))
> >```
> Hmm, this does still not make sense to me. When I evaluate the autodiff call (outside of EnzymeTestUtils), the result is exactly the same whether `C` is `Const` or not, and the test passes when it is `Const`. The derivative of what exactly is wrong?

Can you share an MWE for this?

simsurace:
> > > This example seems to fail with incorrect derivative:
> > > ```julia
> > > test_forward(f, Duplicated, (C, Const), (B, Duplicated))
> > > ```
> > 
> > 
> > Hmm, this does still not make sense to me. When I evaluate the autodiff call (outside of EnzymeTestUtils), the result is exactly the same whether `C` is `Const` or not, and the test passes when it is `Const`. The derivative of what exactly is wrong?
> 
> Can you share an MWE for this?

In this branch,
```julia
using TestEnv; TestEnv.activate()
using Enzyme, EnzymeTestUtils, LinearAlgebra, Test

Te = Float64
A = exp(Symmetric(rand(Te, 4, 4)))
B = rand(Te, 4)
C = cholesky(A)

f(x, y) = ldiv!(x, y)

res1 = let dB = ones(Te, 4), dC = make_zero(C)
    ret, dret = autodiff(Forward, f, Duplicated, Duplicated(C, dC), Duplicated(B, dB))
    dB, dret
end

res2 = let dB = ones(Te, 4), dC = make_zero(C)
    ret, dret = autodiff(Forward, f, Duplicated, Const(C), Duplicated(B, dB))
    dB, dret
end

res1 .≈ res2 # (true, true)

test_forward(f, Duplicated, (C, Const), (B, Duplicated)) # Error
test_forward(f, Duplicated, (C, Duplicated), (B, Duplicated)) # All tests pass
```

sethaxen:
The problem seems to be that if you pass `Const(C)` but the other activities are `Duplicated`, then Enzyme turns the `Const(C)` into a `Duplicated(C, C)` (can see this by adding a `@show typeof(fact), fact.val === fact.dval` in the rule and then calling `autodiff(Forward, f, Duplicated, Const(C), Duplicated(B, dB))`. This results in the absolute wrong derivative being computed, since instead of being a zero, the shadow is the primal. @wsmoses, this looks like an Enzyme bug, but I'm not certain how to raise it in an issue, since it can only be caught within a rule.

simsurace:
Ah, the conversion from `Const` to `Duplicated` seems to happen in the `llvmcall` from what I can tell. Up to and including `enzyme_call`, the activitiy is still `Const`, but then the next item in the stacktrace is the rule, where the activity is suddenly `Duplicated`.

simsurace:
In any case, it seems that these problems are unrelated to this specific PR. I disabled the activities that are problematic. The ones that aren't now all seem to be correct. I would therefore suggest to merge this in order to have the incorrect rules on main fixed. I will now re-run the formatter.

sethaxen:
Please revert the formatting commit. I'm not quite done reviewing the code yet. (better, only lines modified by this PR should be formatted. Otherwise it is extremely difficult for anyone to identify what the changes of this PR were).

sethaxen (changes requested):

test/internal_rules.jl
```diff
@@ -136,282 +136,59 @@ end
...
-            ldiv!(C, b2)
-            @inbounds b2[1]
+            B = rand(Te, 4, 4)
+            b = rand(Te, 4)
+            @testset for TC in (Const, Duplicated), TB in (Const, Duplicated),
+                         Tret in (Const, Duplicated)
+
+                @testset "$(size(_B))" for _B in (B, b)
+                    are_activities_compatible(Tret, TC, TB) || continue
+                    # Non-uniform activities are disabled due to unresolved questions
```
sethaxen, line 175: It would be best if you could open an issue and include the issue number here.

src/internal_rules.jl
```diff
@@ -865,124 +874,130 @@ function EnzymeRules.augmented_primal(
...
-            #   dA −= z B(out)^T
-
+            dZ = U' \ dB
             func.val(cache_A, dB; kwargs...)
             if !isa(A, Const)
+                ∂B = U \ dZ
+                Ā = -dZ * Y' - Z * ∂B'
                 dA = EnzymeRules.width(config) == 1 ? A.dval : A.dval[b]
-                mul!(dA.factors, dB, transpose(cache_Bout), -1, 1)
+                dA.factors .+= UpperTriangular(Ā)
```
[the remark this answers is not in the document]
simsurace: Thanks, I had missed that,

src/internal_rules.jl
```diff
@@ -865,124 +874,130 @@ function EnzymeRules.augmented_primal(
...
-function EnzymeRules.reverse(
-    config,
-    func::Const{typeof(ldiv!)},
-    dret,
-    cache,
-    A::Annotation{<:Cholesky},
-    B::Union{Const, DuplicatedNoNeed, Duplicated, BatchDuplicatedNoNeed, BatchDuplicated};
-    kwargs...
-)
+function EnzymeRules.reverse(config,
```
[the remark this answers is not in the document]
simsurace: Yeah there is no claim to optimality here. I just want to get the rules correct in this PR.
Unless you have something worked out to plug in directly that doesn't need to be translated first, I would prefer to postpone this optimization to another PR, as I don't have time to dive into another debugging marathon :).

test/internal_rules.jl
```diff
@@ -136,282 +136,59 @@ end
...
-        if op == :\
-            @test isapprox(fwdJ, fdJ)
-        end
-        @test isapprox(fwdJ, revJ)
-
-        function h(A, b)
+    @testset "Linear solve for `Cholesky`" begin
+        @testset for Te in (Float64,)
+            A = exp(Symmetric(rand(Te, 4, 4)))
             C = cholesky(A)
```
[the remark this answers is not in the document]
simsurace: It seems to work, but not at the default tolerance for the reverse rule, maybe due to numerical stability. Can we increase the tolerance?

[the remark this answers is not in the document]
sethaxen: How much of an increase is necessary? It should probably pass with an `atol` of 1e-5. If not, something may be wrong.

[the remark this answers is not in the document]
simsurace: I could make it pass with `I + rand` with the default tolerance. For your `randn` version, we would probably need to go to `1e-6` absolute.

src/internal_rules.jl
```diff
@@ -844,18 +840,31 @@ function EnzymeRules.forward(
...
-    fact = if EnzymeRules.needs_primal(config)
+function _ldiv_Cholesky_forward!(L, U, B, dL, dU, dB)
+    ldiv!(L, B)
+    mul!(dB, dL, B, -1, 1)
+    ldiv!(L, dB)
+    ldiv!(U, B)
+    mul!(dB, dU, B, -1, 1)
+    ldiv!(U, dB)
+    return B, dB
+end
```
sethaxen, line 851: This is necessary to support batch activities:
```suggestion
function _ldiv_Cholesky_forward!(L, U, B, dL, dU, dBs)
    ldiv!(L, B)
    for dB in dBs
        mul!(dB, dL, B, -1, 1)
        ldiv!(L, dB)
    end
    ldiv!(U, B)
    for dB in dBs
        mul!(dB, dU, B, -1, 1)
        ldiv!(U, dB)
    end
    return B, dBs
end
```

src/internal_rules.jl
```diff
@@ -780,54 +772,58 @@ function EnzymeRules.forward(::Const{typeof(cholesky)}, RT::Type, A; kwargs...)
...
-                end
-                
-                tmp = dfact.U * retval
-                mul!(dB, dfact.L, tmp, -1, 1)
+                dfact = N == 1 ? fact.dval : fact.dval[b]
+                L = fact.val.L
+                U = fact.val.U
+                dL = dfact.L
+                dU = dfact.U
+                _ldiv_Cholesky_forward!(L, U, B.val, dL, dU, dB)
```
sethaxen, line 824: Now refactor this to call `_ldiv_Cholesky_forward!(L, U, B.val, dL, dU, B.dval)` instead of executing this in the `ntuple` call (see below)

sethaxen (changes requested):
This is getting really close. Looks like tests for batch activities for the rules that support them need to be added. This would have caught a bug in the forward rule for `ldiv!`.

test/internal_rules.jl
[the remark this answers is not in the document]
sethaxen: 1e-6 should be fine.

src/internal_rules.jl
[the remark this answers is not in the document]
sethaxen: Sure, let's just make sure we're testing all the activities/eltypes, and if tests pass, this shouldn't block.

src/internal_rules.jl
simsurace: What about batches in `dL` and `dU`?

test/internal_rules.jl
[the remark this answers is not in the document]
simsurace: I went with `I + rand` for now.

src/internal_rules.jl
simsurace: Done I believe.

sethaxen:
@simsurace, sorry for the delay in reviewing. In the middle of a few weeks of travel and will get to it as soon as I can.

test/internal_rules.jl
```diff
@@ -136,282 +136,66 @@ end
...
-        J = zeros(length(b), length(b))
-        for i in 1:length(b)
-            db[i] = 1.0
-            dx = A\db
-            db[i] = 0.0
-            J[i, :] = dx
+    @testset "cholesky" begin
+        @testset "with wrapper arguments" begin
+            @testset for Te in (Float64,), TS in (Symmetric, Hermitian), uplo in (:U, :L)
+                @testset for TA in (Const, Duplicated), Tret in (Const, Duplicated)
```
sethaxen, line 142: This is still missing some cases. From what I can tell, the rules allow batch activities, and the forward-mode rules allow complex `Hermitian` matrices. These need to also be tested.

test/internal_rules.jl
```diff
@@ -136,282 +136,66 @@ end
...
-        J = zeros(length(b), length(b))
-        for i in 1:length(b)
-            db[i] = 1.0
-            dx = A\db
-            db[i] = 0.0
-            J[i, :] = dx
+        @testset "without wrapper arguments" begin
+            _square(A) = A * A'
+            @testset for Te in (Float64,)
+                @testset for TA in (Const, Duplicated), Tret in (Const, Duplicated)
```
sethaxen, line 154: I think this is also missing the batch activities and complex eltypes (at least for `Matrix`)

test/internal_rules.jl
```diff
@@ -136,282 +136,66 @@ end
...
-        function h(A, b)
-            C = cholesky(A)
-            b2 = copy(b)
-            ldiv!(C, b2)
-            @inbounds b2[1]
+    @testset "Linear solve for `Cholesky`" begin
+        activities = (Const, Duplicated, DuplicatedNoNeed, BatchDuplicated,
+                      BatchDuplicatedNoNeed)
+        @testset for Te in (Float64, ComplexF64), uplo in ('L', 'U')
+            C = Cholesky(I + rand(Te, 4, 4), uplo, 0)
```
sethaxen, line 168: Could you add a brief not why the addition of `I` is necessary?

sethaxen (changes requested):

sethaxen:
Seems CI is broken on master, but before the most recent commit there were still a number of failures in the tests for these rules on v1.8 and v1.9. Do you know why these are happening?

simsurace:
I could make the other activities work, but I'm not able to make the complex element type work in all cases. Will get back to it later.

simsurace:
I think I may have come across a bug or at least a major annoyance: 
```julia
square(A) = A * adjoin(A)
A = rand(ComplexF64, 5, 5)
ishermitian(square(A)) # true

dA = rand(ComplexF64, 5, 5)
S, dS = autodiff(Forward, square, Duplicated, Duplicated(A, dA))
S ≈ _square(A) # true
ishermitian(S) # false

test_forward(square, Duplicated, (A, Duplicated)) # passes
```
So somehow the forward mode does not produce the same result as the function for `square`, but this is not caught by EnzymeTestUtils.
This makes the tests fail for complex element types.
I will still commit the test because I believe it is otherwise correct, and open a separate issue for this.

test/internal_rules.jl
simsurace: I think this is now addressed, but #1456 leads to test failures for complex element type.

test/internal_rules.jl
simsurace: I think this is now addressed, but https://github.com/EnzymeAD/Enzyme.jl/issues/1456 leads to test failures for complex element type.

test/internal_rules.jl
simsurace: Done!

simsurace:
Seems there are still random failures that I did not see locally. Will try to fix those.

wsmoses:
hey folks just wanted to check back in here how things were going, and if I can be of any help.

admittedly the cholesky math here is a bit out of my depth

simsurace:
Hi, thanks for checking. I was gonna try to fix those failures but won't have time now for a few weeks.

wsmoses:
@simsurace the tests seem to work on everything except 1.8.

Do you have an estimate of when you can look at this?

Some of the pieces are blocking for folks so I may cherry-pick out the forward mode fix, in the interim

simsurace:
Hi, thanks for taking care. From next week on I should be able to get back to this.

wsmoses:
This PR has been partially subsumed by additional support of lapack functions.

In particular we now have lapack support for cholesky solve in forward/reverse, and cholesky ldiv in reverse mode [but not forward]. All of this is limited to real numbers.

This PR still contains useful functionality for complex numbers, as well as testing, but as an FYI @simsurace

simsurace:
Hi @wsmoses, thanks for the heads up. This is cool. I wonder what the best way is to get this PR merged.
I can look at the 1.8 failures. Assuming we can get all tests to pass, do we first merge this PR and then open another one to disable the rules for the cases now covered by the new features?

wsmoses:
At this point with the new stuff already in place in main, I might actually recommend the other way -- trying to add all these tests and seeing what fails, and then adding back the missing stuff when needed

wsmoses:
Closing as stale

Issue #1308, opened 2024-02-25
Incorrect sign for imaginary part of derivative of functions returning complex numbers


giordano:
```julia
julia> using Enzyme

julia> ∂(f, z) = first(first(autodiff(Reverse, f, Active, Active(z))))
∂ (generic function with 1 method)

julia> v = rand(ComplexF64, 3);

julia> ∂.(exp, v)
3-element Vector{ComplexF64}:
 2.4742094004406776 - 0.6831994346902432im
 1.1529144463383987 - 0.31931177102730973im
 1.0208330780674268 - 1.295298037192365im

julia> exp.(v)
3-element Vector{ComplexF64}:
 2.4742094004406776 + 0.6831994346902432im
 1.1529144463383987 + 0.31931177102730973im
 1.0208330780674268 + 1.295298037192365im

julia> conj.(∂.(exp, v)) ≈ exp.(v)
true
```
Note that the sign of the imaginary part is always wrong, Enzyme is basically always returning the conjugate of the correct result.  The same happens with more complicate functions:
```julia
julia> f(z) = z ^ z
f (generic function with 1 method)

julia> f′(z) = f(z) * (log(z) + one(z))
f′ (generic function with 1 method)

julia> v = rand(ComplexF64, 1000);

julia> ∂.(f, v) ≈ f′.(v)
false

julia> conj.(∂.(f, v)) ≈ f′.(v)
true
```
BTW, I'm not sure #547 is still an issue, I could differentiate power of complex numbers (except the result is wrong because of the conjugation).

For the record:
```
julia> versioninfo()
Julia Version 1.10.0
Commit 3120989f39b (2023-12-25 18:01 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: macOS (arm64-apple-darwin22.4.0)
  CPU: 8 × Apple M1
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, apple-m1)
  Threads: 1 on 4 virtual cores

(jl_30Oveo) pkg> st -m Enzyme Enzyme_jll
Status `/private/var/folders/v2/hmy3kzgj4tb3xsy8qkltxd0r0000gn/T/jl_30Oveo/Manifest.toml`
  [7da242da] Enzyme v0.11.17
  [7cc45869] Enzyme_jll v0.0.102+0
```

---
[...]

PR #1309, opened 2024-02-26
Err and correctly handle complex returns in reverse mode

wsmoses:
@giordano @MasonProtter

resolves https://github.com/EnzymeAD/Enzyme.jl/issues/1308

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -276,7 +276,7 @@ That is why Enzyme provides a helper function `Enzyme.make_zero` that does this 
 
 Differentiation of a function which returns a complex number is ambiguous, because there are several different gradients which may be desired. Rather than assume a specific of these conventions and potentially result in user error when the resulting derivative is not the desired one, Enzyme forces users to specify the desired convention by returning a real number instead.
 
-Consider the function `f(z) = z*z`. If we were to differentiate this and have real inputs and outputs, the derivative `f'(z)` would be unambiguously 2. However, consider breaking down a complex number down into real and imaginary parts. Suppose now we were to call `f` with the explicit real and imaginary components, `z = x + i y`. This means that `f` is a function that takes an input of two values and returns two values `f(x, y) = u(x, y) + i v(x, y)`. In the case of `z^2` this means that `u(x,y) = x*x-y*y` and `v(x,y) = 2*x*y`.
+Consider the function `f(z) = z*z`. If we were to differentiate this and have real inputs and outputs, the derivative `f'(z)` would be unambiguously `2*z`. However, consider breaking down a complex number down into real and imaginary parts. Suppose now we were to call `f` with the explicit real and imaginary components, `z = x + i y`. This means that `f` is a function that takes an input of two values and returns two values `f(x, y) = u(x, y) + i v(x, y)`. In the case of `z*z` this means that `u(x,y) = x*x-y*y` and `v(x,y) = 2*x*y`.
 
 
 If we were to look at all first-order derivatives in total, we would end up with a 2x2 matrix (i.e. Jacobian), the derivative of each output wrt each input. Let's try to compute this, first by hand, then with Enzyme.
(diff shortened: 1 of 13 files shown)


docs/src/index.md
```diff
@@ -271,3 +271,132 @@ da2
 
 Sometimes, determining how to perform this zeroing can be complicated.
 That is why Enzyme provides a helper function `Enzyme.make_zero` that does this automatically.
+
+### Complex Numbers
+
+Differentiation of a function which returns a complex number is ambiguous, because there are several different gradients which may be desired. Rather than assume a specific of these conventions and potentially result in user error when the resulting derivative is not the desired one, Enzyme forces users to specify the desired convention by returning a real number instead.
+
+Consider the function `f(z) = z*z`. If we were to differentiate this and have real inputs and outputs, the derivative `f'(z)` would be unambiguously 2. However, consider breaking down a complex number down into real and imaginary parts. Suppose now we were to call `f` with the explicit real and imaginary components, `z = x + i y`. This means that `f` is a function that takes an input of two values and returns two values `f(x, y) = u(x, y) + i v(x, y)`. In the case of `z^2` this means that `u(x,y) = x*x-y*y` and `v(x,y) = 2*x*y`.
```
MasonProtter, line 279:
```suggestion
Consider the function `f(z) = z*z`. If we were to differentiate this and have real inputs and outputs, the derivative `f'(z)` would be unambiguously `2z`. However, consider breaking down a complex number down into real and imaginary parts. Suppose now we were to call `f` with the explicit real and imaginary components, `z = x + i y`. This means that `f` is a function that takes an input of two values and returns two values `f(x, y) = u(x, y) + i v(x, y)`. In the case of `z^2` this means that `u(x,y) = x*x-y*y` and `v(x,y) = 2*x*y`.
```

docs/src/index.md
```diff
@@ -271,3 +271,132 @@ da2
...
+
+# output
+(6.2 + -5.4im, 5.4 + 6.2im)
+```
+
+In contrast, Forward mode differentiation computes the derivative of all outputs with respect to a single input by providing a differential input. Thus we need to seed the shadow input with either 1.0 or 1.0im, respectively. This will compute the transpose of the matrix we found earlier.
+
+```
+d/dx f(x, y) = d/dx [u(x,y), v(x,y)] = d/dx [x*x-y*y, 2*x*y] = [2*x, 2*y];
+d/dy f(x, y) = d/dy [u(x,y), v(x,y)] = d/dy [-2*y, 2*x];
```
MasonProtter, line 337:
```suggestion
d/dy f(x, y) = d/dy [u(x,y), v(x,y)] = d/dy [x*x-y*y, 2*x*y] = [-2*y, 2*x];
```

docs/src/index.md
```diff
@@ -271,3 +271,132 @@ da2
...
+```jldoctest complex
+Enzyme.autodiff(Forward, f, BatchDuplicated(z, (1.0+0.0im, 0.0+1.0im))
+
+# output
+(6.2 + -5.4im, 5.4 + 6.2im)
+```
+
+Taking Jacobians with respect to the real and imaginary results is fine, but for a complex scalar function it would be really nice to have a single complex derivative. More concretely, in this case when differentiating `z*z`, it would be nice to simply return `2*z`. However, there are four independent variables in the 2x2 jacobian, but only two in a complex number. 
+
+Most often, complex differentiation is viewed in the lens of directional derivatives. For example, what is the derivative of the function as the real input increases, or as the imaginary input increases. Consider the derivative along the real axis, $\texttt{lim}_{\Delta x \rightarrow 0} \frac{f(x+\Delta x, y)-f(x, y)}{\Delta x}$. This simplifies to $\texttt{lim}_{\Delta x \rightarrow 0} \frac{u(x+\Delta x, y)-u(x, y) + i \left[ v(x+\Delta x, y)-v(x, y)\right]}{\Delta x} = \frac{\partial}{\partial x} u(x,y) + i\frac{\partial}{\partial x} v(x,y)$. This is exactly what we computed by seeding forward mode with a shadow of `1.0 + 0.0im`.
```
MasonProtter, line 360: I quibble a bit with the wording "most often" here. In my experience talking to people about this, many aren't even aware there's any derivative other than the holomorphic `d/dz` one, though yes there's a whole population of people who only think about it from an optimization point of view (typically people who work with non-holomorphic `C -> R` functions.

MasonProtter (commented):
Great stuff! One area that *might* be good to discuss, but we could also just skip is discussing that for general non-holomorphic functions, you can compute the pair of derivatives `(d/dz, d/dz̄)` instead of `(d/dx, d/dy)`. For a holomorphic function you'll have `df/dz̄ = 0` and for an anti-holomorphic function you'll get `df/dz = 0`

In general these things can be computed as
```
df/dz = (pullback(1)' + pullback(im)' * im)/2
df/dz̄ = (pullback(1)' + pullback(im)' * im)/2
```

People sometimes call this pair of derivatives 'Wirtinger' derivatives.

If you want I can try to write a quick paragraph on it for the docs.

PR #1311, opened 2024-02-27
Add Attributor flag in optimize.jl

dchang10:
Running the attributor sometime causes a segmentation fault during optimization. Added a flag that toggles whether the attributor is run or not. The flag should be set before any enzyme call is made.


wsmoses (approved):
Better code than I'd write

PR #1326, opened 2024-03-05
Add Julia formatter

simsurace:
Maybe this will help a bit with readability.

### DIFF
--- a/.github/workflows/Format.yml
+++ b/.github/workflows/Format.yml
@@ -1,0 +1,26 @@
+name: Format suggestions
+
+on:
+  pull_request:
+  
+concurrency:
+  # Skip intermediate builds: always.
+  # Cancel intermediate builds: only if it is a pull request build.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
+
+jobs:
+  format:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: 1
+      - run: |
+          julia  -e 'using Pkg; Pkg.add("JuliaFormatter")'
+          julia  -e 'using JuliaFormatter; format("."; verbose=true)'
+      - uses: reviewdog/action-suggester@v1
+        with:
+          tool_name: JuliaFormatter
+          fail_on_error: true
--- a/.JuliaFormatter.toml
+++ b/.JuliaFormatter.toml
@@ -1,0 +1,1 @@
+style = "blue"


simsurace:
There is an issue with permissions, probably requires admin permissions to fix.

.github/workflows/Format.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: 1
+      - run: |
+          julia  -e 'using Pkg; Pkg.add("JuliaFormatter")'
+          julia  -e 'using JuliaFormatter; format("."; verbose=true)'
+      - uses: reviewdog/action-suggester@v1
+        with:
```
vchuravy, line 29:
```suggestion
        with:
          github_token: ${{ secrets.GITHUB_TOKEN }}
```

examples/box.jl
```diff
@@ -202,38 +200,50 @@ end
...
     ## first computing the time derivatives of the various temperatures and salinities
-    if u > 0 
-
-        dstate_now_dt[1] = u * (state_now[3] - state_now[1]) / params.boxvol[1] + params.gamma * (params.Tstar[1] - state_now[1]) 
-        dstate_now_dt[2] = u * (state_now[1] - state_now[2]) / params.boxvol[2] + params.gamma * (params.Tstar[2] - state_now[2])
-        dstate_now_dt[3] = u * (state_now[2] - state_now[3]) / params.boxvol[3] 
-
-        dstate_now_dt[4] = u * (state_now[6] - state_now[4]) / params.boxvol[1] + params.FW[1] / params.boxvol[1]
-        dstate_now_dt[5] = u * (state_now[4] - state_now[5]) / params.boxvol[2] + params.FW[2] / params.boxvol[2]
+    if u > 0
```
vchuravy, line 207: I find this a lot less readable,

Generally I am not the biggest fan of the Blue style.

simsurace: We can choose another one, I don't have strong preferences. SciML is better?

vchuravy: YAS or Default.

simsurace: Ok, but all of these styles will break the lines as well. I think this is more of a problem with applying the formatting style post-hoc, because adhering to a style would push one to write things differently in the first place. Here, I would have abbreviated the variable names according to the math equations in the comments, i.e. use `T` instead of `state` etc.

vchuravy:
Can we merge the Formatter on its own and slowly apply the suggested changes? Right now this will cause large conflicts with other PRs. I think `clang-format` can do this and for Python there is https://dev.to/akaihola/improving-python-code-incrementally-3f7a

x-ref: https://github.com/domluna/JuliaFormatter.jl/issues/191

simsurace:
Sure thing.

vchuravy:
Thanks! Lets see how this does and if it is to annoying we may disable it.

vchuravy (approved):

.github/workflows/Format.yml
vchuravy: @simsurace did you intend to remove this when you force pushed?

simsurace:
How is it gonna work to get the codebase formatted now?

vchuravy:
As an example see https://github.com/EnzymeAD/Enzyme.jl/pull/1302 where the changes themselves will be formatted. So slowly over time the code base will match the format rules, but without the cost of having large merge conflicts with existing PRs

simsurace:
Ok. Might take a very long time to traverse the entire code base, but we could still format the other files separately. But e.g. in #1307 the action still fails due to permissions.

PR #1282, opened 2024-02-12
Mark blas as on by default (part 2)

wsmoses:

PR #1327, opened 2024-03-05
[EnzymeTestUtils] Vectorize function for FiniteDifferencesCalls

sethaxen:
This PR adds our own version of `FiniteDifferences.to_vec`, which allows EnzymeTestUtils.jl to correctly test rules with structured array inputs/outputs.

As noted in a few issues, e.g. https://github.com/EnzymeAD/Enzyme.jl/pull/1264#issuecomment-1921654762, FiniteDifferences (like ChainRules) represents tangents of structured arrays with arrays whose elements are semantically interpreted as tangent to the original array in its subspace of the dense arrays. Enzyme, on the other hand, treats arrays no differently from other structs. This difference is highlighted in how the two packages treat tangents of `Symmetric`. Both use a `Symmetric` container, but FD interprets the densified version of the tangent as tangent to the densified version of the original array, while Enzyme interprets the underlying `.data` of the tangent as tangent to the underlying `.data` of the original array.

The `to_vec` implementation included here allows us to convert all functions to the signature `f(::Vector{<:AbstractFloat})::Vector{<:AbstractFloat}` before calling FiniteDifferences, so that we control how the tangents are constructed.


lib/EnzymeTestUtils/src/to_vec.jl
```diff
@@ -0,0 +1,66 @@
...
+
+# types: they and their fields aren't differentiable, so we bypass them
+to_vec(x::Type) = (Float32[], _ -> x)
+
+# below code is adapted from https://github.com/JuliaDiff/FiniteDifferences.jl/blob/99ad77f05bdf6c023b249025dbb8edc746d52b4f/src/to_vec.jl
+# MIT Expat License
+# Copyright (c) 2018 Invenia Technical Computing
+
+# structs: recursively call to_vec on each field
+function to_vec(x::T) where {T}
```
wsmoses, line 18: Does this need to be made with a map of seen variables (e.g. if you have tuple{Vector, Vector} and x[0] === x[1])?

wsmoses: I suppose since we only take type's here and not variables, that doesn't apply here

sethaxen:
This still needs tests for `to_vec`. It seems that on main the reverse-mode mutated callable test in EnzymeTestUtils has been erroring since #1282. The error changed in #1292.

lib/EnzymeTestUtils/src/to_vec.jl
sethaxen: No, I don't think it should matter whether e.g. a tuple contains the same array twice. But we should test for that. If that is a problem, it will put a major restriction on testable functions. We could at least detect that and throw an informative error pretty easily I think.

sethaxen:
@wsmoses [e6fa10b](https://github.com/EnzymeAD/Enzyme.jl/pull/1282/commits/e6fa10b62ff0c408e346082dca1b543733fe354c) marked some formerly broken EnzymeTestUtils tests as passing, but they were still failing when that PR was merged. Should that commit be reverted?

lib/EnzymeTestUtils/src/to_vec.jl
sethaxen: (a limitation we cannot let go of is that the primal and the shadow share the same memory pattern. e.g. if `x` is an array, Enzyme is fine with a user providing `Duplicated((x, x), (dx, copy(dx)))` to `autodiff`, but for testing we would need to require `Duplicated((x, x), (dx, dx))`.)

sethaxen:
@wsmoses this PR is stalled pending answers to the questions in https://github.com/EnzymeAD/Enzyme.jl/pull/1327#discussion_r1516000282 and https://github.com/EnzymeAD/Enzyme.jl/pull/1327#issuecomment-1979737864

lib/EnzymeTestUtils/src/to_vec.jl
wsmoses: Is there a reason why zero_tangent can't just call make_zero? The make_zero logic is intentionally general (and if needed user extensible like deepcopy), to deal with all the sorts of crazy struct/etc things.

wsmoses: I see what you mean by the to_vec unifying simplicity, but we need make_zero to be potentially "easily" user-extensible for custom types and I don't want to make that burden harder unless necessary (even though it is compelling here to simplify code).

wsmoses:
> @wsmoses [e6fa10b](https://github.com/EnzymeAD/Enzyme.jl/pull/1282/commits/e6fa10b62ff0c408e346082dca1b543733fe354c) marked some formerly broken EnzymeTestUtils tests as passing, but they were still failing when that PR was merged. Should that commit be reverted?

Yeah this should be fixed here (that was itself intending to fix things since the default Enzyme BLAS internals had changed).

lib/EnzymeTestUtils/src/to_vec.jl
sethaxen: > Is there a reason why zero_tangent can't just call make_zero?

The main reason is we need aliased arrays to in the primal to also have aliased shadows for our code to do the right thing. `zero_tangent` doesn't currently handle that but could. `make_zero` doesn't either (because it relies on `deepcopy_internal`, which also doesn't).

I guess it depends on whether we think those user-extensions of `make_zero` would ever be for structs one might want to pass to EnzymeTestUtils. If so, the user would also need to similarly extend `zero_tangent`.

sethaxen:
> > @wsmoses [e6fa10b](https://github.com/EnzymeAD/Enzyme.jl/pull/1282/commits/e6fa10b62ff0c408e346082dca1b543733fe354c) marked some formerly broken EnzymeTestUtils tests as passing, but they were still failing when that PR was merged. Should that commit be reverted?
> 
> Yeah this should be fixed here (that was itself intending to fix things since the default Enzyme BLAS internals had changed).

I disagree this should be fixed here, as it's unrelated to this PR and was already broken when #1282 was merged. If you have a fix you can add, that would be great, but I don't intend to work on that.

wsmoses:
Sorry by fixed here I don't mean in this PR but in EnzymeTestUtils

sethaxen:
Got it. Since it currently prevents us from testing the changes here, I'll pick this up again when the EnzymeTestUtils CI passes on master.

wsmoses:
@sethaxen If you don't have cycles, then mind making an issue then with the needs-help tag? It should just be setting the EnzymeTestUtils failures/skips to match what happens in CI.

wsmoses:
@sethaxen enzymetestutils passes main now, if you have cycles to get this over the finish line

lib/EnzymeTestUtils/src/to_vec.jl
```diff
@@ -0,0 +1,151 @@
...
+            for i in 1:nf
+                if isdefined(x, i)
+                    xi = flds_new[i]
+                    ccall(:jl_set_nth_field, Cvoid, (Any, Csize_t, Any), x_new, i - 1, xi)
+                end
+            end
+        else
+            x_new = ccall(:jl_new_structv, Any, (Any, Ptr{Any}, UInt32), RT, flds_new, nf)
+        end
+        seen_xs[x] = x_new
```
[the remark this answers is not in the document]
wsmoses: Yeah I think we probably don't want to make immutable values equivalent to each other here (complex/that struct as an example).

lib/EnzymeTestUtils/src/generate_tangent.jl
```diff
@@ -17,14 +17,20 @@ map_fields_recursive(f, x::T...) where {T<:AbstractFloat} = f(x...)
...
+    v_new = rand(rng, -9:T(0.01):9, length(v))
+    return from_vec(v_new)
+end
 
-zero_tangent(x) = map_fields_recursive(zero_tangent, x)
-zero_tangent(::T) where {T<:AbstractFloat} = zero(T)
-zero_tangent(x::T) where {T<:Array{<:Number}} = zero_tangent.(x)
+# differs from Enzyme.make_zero primarily in that reshaped Arrays in the argument will share
+# the same memory in the output.
+function zero_tangent(x)
```
wsmoses, line 30: So technically we could still implement this with make zero.

Specificially if we passed in say a tuple (output, input) if out and in aliased make_zero would make the results alias as well.

Alternatively we could prepopulate the seed dict.

sethaxen: > Specificially if we passed in say a tuple (output, input) if out and in aliased make_zero would make the results alias as well.

This doesn't work in the case of a reshaped array of floats, which share the same memory but are treated by Julia as non-equivalent. The difference between `make_zero` and `zero_tangent` is this:

```julia
julia> x1 = randn(3, 1);

julia> x2 = vec(x1);

julia> x = (x1, x2);

julia> z1 = Enzyme.make_zero(x)
([0.0; 0.0; 0.0;;], [0.0, 0.0, 0.0])

julia> z2 = EnzymeTestUtils.zero_tangent(x)
([0.0; 0.0; 0.0;;], [0.0, 0.0, 0.0])

julia> z1[1][1] = 1
1

julia> z1
([1.0; 0.0; 0.0;;], [0.0, 0.0, 0.0])

julia> z2[1][1] = 1
1

julia> z2
([1.0; 0.0; 0.0;;], [1.0, 0.0, 0.0])
```

> Alternatively we could prepopulate the seed dict.

Wouldn't this add more code complexity than what we currently do? i.e. we already have a `to_vec` that knows how to handle reshaped float arrays, and we use that for all tangent generation, both for FiniteDifferences, but also for constructing zero and random tangents.

wsmoses: Yeah fair, was a suggestion if it worked and simplifies but otherwise LGTM

wsmoses: Though wait would testutils work on a reshape call?

wsmoses (approved):

sethaxen:
Don't merge yet, need to resolve a bug on v1.7 first.

wsmoses:
Sg will let you merge when things are set

sethaxen:
> Don't merge yet, need to resolve a bug on v1.7 first.

Found it, it's the example in #1394

wsmoses:
Awesome, though unfortunately the solution to that is to use a later Julia as it is a bug in Julia's GC that was fixed. We should just version gate the test.

sethaxen:
@wsmoses I added some more tests and tweaked some docstrings. I also needed to fix the random seed to get tests to pass. This is now ready to merge and could be merged before or after #1398.

wsmoses:
LGTM lmk if you want me to merge (but I think you should have perms to)

PR #1337, opened 2024-03-08
Add JuMP extension

michel2323:
This extends JuMP and allows a user in JuMP to differentiate an external function using Enzyme. I still should reduce the allocations. Feedback is very welcome.

Use case:
```julia
using Ipopt
using JuMP
using Enzyme

# Rosenbrock
rosenbrock(x...) = (1 - x[1])^2 + 100 * (x[2] - x[1]^2)^2

model = Model(Ipopt.Optimizer)
op_rosenbrock =  model[:op_rosenbrock] = add_nonlinear_operator(model, 2, rosenbrock; name=:op_rosenbrock) 
@variable(model, x[1:2])

@objective(model, Min, op_rosenbrock(x[1],x[2]))

optimize!(model)
```


michel2323:
Since we're using the memory, I added the batched mode to get the Hessian in one go.

wsmoses (commented):
I feel like this would be better suited as an extension to Jump

odow:
This should be closed. We're discussing it over in https://github.com/jump-dev/JuMP.jl/pull/3712.

Issue #1189, opened 2023-12-08
How to build documentation locally?


gdalle:
While writing #1187 I tried to build the docs locally by following this process (from the `CI.yml` file):

```julia
pkg> dev ./lib/EnzymeCore
pkg> dev ./lib/EnzymeTestUtils
pkg> dev .
pkg> instantiate
julia> include("docs/make.jl")
```

but Julia still complained about missing dependencies and asked me to instantiate. What am I missing?

lassepe:
Did you run the commands above in a REPL that has the docs env activated (i.e., `julia --project=docs/`)? If so, that's exactly what I do offline and it works well.

PR #1343, opened 2024-03-15
Add FAQ section to docs

### LINKED ISSUE #1329: Why does `Duplicated(x, dx)` assume `x` and `dx` have the same type?
A typical example where this is painful is writing a JVP/VJP into a row/column of a Jacobian. Then `dx` or `x` can be reshaped or views.

cc @adrhill

gdalle:
EDIT: see full extend of the PR below https://github.com/EnzymeAD/Enzyme.jl/pull/1343#issuecomment-2003307409

Fix #1329 by adding a note to the `Duplicated` docstring


lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -58,6 +58,17 @@ Mark a function argument `x` of [`autodiff`](@ref Enzyme.autodiff) as duplicated
...
 `∂f_∂x`.
+
+# Note
+
+Enzyme forces `x` and `∂f_∂x` to have the same types when constructing `Duplicated`.
+This is not a mathematical requirement, but rather a guardrail to prevent user error.
+The memory locations of `x` and `∂f_∂x` are accessed in the same way by the differentiation code, so they should have the same data layout.
+Equality of types is an approximation of this condition.
+
+In some ways, type equality is too strict, but in other ways it is not strict enough.
```
wsmoses, line 69: Could you add an explanation here as to why, in addition to the example? I can imagine people getting confused otherwise.

Also would you be able to add the example from the issue with the pointer offsets?

lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -58,6 +58,17 @@ Mark a function argument `x` of [`autodiff`](@ref Enzyme.autodiff) as duplicated
...
+For instance, if `x` and `∂f_∂x` are sparse arrays, their sparsity pattern should be identical, but this is not encoded in the type.
+To make sure that `∂f_∂x` has the right data layout, create it with `∂f_∂x = Enzyme.make_zero(x)`.
```
wsmoses, line 71: Is the duplicated type constructor the right place for this. This equally applies to DuplicatedNoNeed, BatchDuplicated, etc

cc @vchuravy

lib/EnzymeCore/src/EnzymeCore.jl
gdalle: I added a small explanation on the too strict / too lax thing.

As for the pointer offset example, I cannot add it because I do not understand it. I gave you the beginner version of this explanation, which is the one suited for people like me ;)

lib/EnzymeCore/src/EnzymeCore.jl
gdalle: I put it in a brand new FAQ section, where I also added the complicated stuff from the home page and the "implementing pullbacks" blurb

gdalle:
I used the opportunity to restructure a bit, as mentioned in #1344 

The `Duplicated` explanation is now part of a broader "FAQ" page, which also contains the former end of the home page and the pullback stuff.

I switched the order of the tutorials to put the basic one first, and renamed it accordingly.

gdalle:
Hopefully the build works, it failed locally due to dependency shenanigans (similar to #1189)

<details>

```bash
guillaume Enzyme.jl ±|gd/duplicated_same ✗|→ julia --project=docs -e '
    using Pkg
    Pkg.develop(path="lib/EnzymeCore")
    Pkg.develop(path="lib/EnzymeTestUtils")
    Pkg.develop(PackageSpec(path=pwd()))
    Pkg.instantiate()
    include("docs/make.jl")
'
   Resolving package versions...
    Updating `~/Work/GitHub/Julia/Enzyme.jl/docs/Project.toml`
⌅ [e30172f5] + Documenter v0.27.25
  [f151be2c] + EnzymeCore v0.7.0 `../lib/EnzymeCore`
  [98b081ad] + Literate v2.16.1
    Updating `~/Work/GitHub/Julia/Enzyme.jl/docs/Manifest.toml`
  [a4c015fc] + ANSIColoredPrinters v0.0.1
  [ffbed154] + DocStringExtensions v0.9.3
⌅ [e30172f5] + Documenter v0.27.25
  [f151be2c] + EnzymeCore v0.7.0 `../lib/EnzymeCore`
  [b5f81e59] + IOCapture v0.2.4
  [682c06a0] + JSON v0.21.4
  [98b081ad] + Literate v2.16.1
  [69de0a69] + Parsers v2.8.1
  [aea7be01] + PrecompileTools v1.2.1
  [21216c6a] + Preferences v1.4.3
  [56f22d72] + Artifacts
  [2a0f44e3] + Base64
  [ade2ca70] + Dates
  [b77e0a4c] + InteractiveUtils
  [76f85450] + LibGit2
  [8f399da3] + Libdl
  [56ddb016] + Logging
  [d6f4376e] + Markdown
  [a63ad114] + Mmap
  [ca575930] + NetworkOptions v1.2.0
  [de0858da] + Printf
  [3fa0cd96] + REPL
  [9a3f8284] + Random
  [ea8e919c] + SHA v0.7.0
  [9e88b42a] + Serialization
  [6462fe0b] + Sockets
  [fa267f1f] + TOML v1.0.3
  [8dfed614] + Test
  [cf7118a7] + UUIDs
  [4ec0a83e] + Unicode
  [e37daf67] + LibGit2_jll v1.6.4+0
  [29816b5a] + LibSSH2_jll v1.11.0+1
  [c8ffd9c3] + MbedTLS_jll v2.28.2+1
        Info Packages marked with ⌅ have new versions available but compatibility constraints restrict them from upgrading. To see why use `status --outdated -m`
   Resolving package versions...
ERROR: Unsatisfiable requirements detected for package Enzyme [7da242da]:
 Enzyme [7da242da] log:
 ├─possible versions are: 0.1.0-0.11.20 or uninstalled
 ├─restricted to versions 0.11-0.12 by EnzymeTestUtils [12d8515a], leaving only versions: 0.11.0-0.11.20
 │ └─EnzymeTestUtils [12d8515a] log:
 │   ├─possible versions are: 0.1.5 or uninstalled
 │   └─EnzymeTestUtils [12d8515a] is fixed to version 0.1.5
 └─restricted by compatibility requirements with EnzymeCore [f151be2c] to versions: 0.1.0-0.10.9 or uninstalled — no versions left
   └─EnzymeCore [f151be2c] log:
     ├─possible versions are: 0.7.0 or uninstalled
     └─EnzymeCore [f151be2c] is fixed to version 0.7.0
Stacktrace:
  [1] propagate_constraints!(graph::Pkg.Resolve.Graph, sources::Set{Int64}; log_events::Bool)
    @ Pkg.Resolve ~/.julia/juliaup/julia-1.10.2+0.x64.linux.gnu/share/julia/stdlib/v1.10/Pkg/src/Resolve/graphtype.jl:1072
  [2] propagate_constraints! (repeats 2 times)
    @ ~/.julia/juliaup/julia-1.10.2+0.x64.linux.gnu/share/julia/stdlib/v1.10/Pkg/src/Resolve/graphtype.jl:1008 [inlined]
  [3] simplify_graph!(graph::Pkg.Resolve.Graph, sources::Set{Int64}; clean_graph::Bool)
    @ Pkg.Resolve ~/.julia/juliaup/julia-1.10.2+0.x64.linux.gnu/share/julia/stdlib/v1.10/Pkg/src/Resolve/graphtype.jl:1533
  [4] simplify_graph! (repeats 2 times)
    @ ~/.julia/juliaup/julia-1.10.2+0.x64.linux.gnu/share/julia/stdlib/v1.10/Pkg/src/Resolve/graphtype.jl:1532 [inlined]
  [5] resolve_versions!(env::Pkg.Types.EnvCache, registries::Vector{Pkg.Registry.RegistryInstance}, pkgs::Vector{Pkg.Types.PackageSpec}, julia_version::VersionNumber, installed_only::Bool)
    @ Pkg.Operations ~/.julia/juliaup/julia-1.10.2+0.x64.linux.gnu/share/julia/stdlib/v1.10/Pkg/src/Operations.jl:407
  [6] targeted_resolve(env::Pkg.Types.EnvCache, registries::Vector{Pkg.Registry.RegistryInstance}, pkgs::Vector{Pkg.Types.PackageSpec}, preserve::Pkg.Types.PreserveLevel, julia_version::VersionNumber)
    @ Pkg.Operations ~/.julia/juliaup/julia-1.10.2+0.x64.linux.gnu/share/julia/stdlib/v1.10/Pkg/src/Operations.jl:1360
  [7] tiered_resolve(env::Pkg.Types.EnvCache, registries::Vector{Pkg.Registry.RegistryInstance}, pkgs::Vector{Pkg.Types.PackageSpec}, julia_version::VersionNumber, try_all_installed::Bool)
    @ Pkg.Operations ~/.julia/juliaup/julia-1.10.2+0.x64.linux.gnu/share/julia/stdlib/v1.10/Pkg/src/Operations.jl:1349
  [8] _resolve(io::Base.TTY, env::Pkg.Types.EnvCache, registries::Vector{Pkg.Registry.RegistryInstance}, pkgs::Vector{Pkg.Types.PackageSpec}, preserve::Pkg.Types.PreserveLevel, julia_version::VersionNumber)
    @ Pkg.Operations ~/.julia/juliaup/julia-1.10.2+0.x64.linux.gnu/share/julia/stdlib/v1.10/Pkg/src/Operations.jl:1370
  [9] develop(ctx::Pkg.Types.Context, pkgs::Vector{Pkg.Types.PackageSpec}, new_git::Set{Base.UUID}; preserve::Pkg.Types.PreserveLevel, platform::Base.BinaryPlatforms.Platform)
    @ Pkg.Operations ~/.julia/juliaup/julia-1.10.2+0.x64.linux.gnu/share/julia/stdlib/v1.10/Pkg/src/Operations.jl:1410
 [10] develop
    @ ~/.julia/juliaup/julia-1.10.2+0.x64.linux.gnu/share/julia/stdlib/v1.10/Pkg/src/Operations.jl:1402 [inlined]
 [11] develop(ctx::Pkg.Types.Context, pkgs::Vector{Pkg.Types.PackageSpec}; shared::Bool, preserve::Pkg.Types.PreserveLevel, platform::Base.BinaryPlatforms.Platform, kwargs::@Kwargs{io::Base.TTY})
    @ Pkg.API ~/.julia/juliaup/julia-1.10.2+0.x64.linux.gnu/share/julia/stdlib/v1.10/Pkg/src/API.jl:225
 [12] develop(pkgs::Vector{Pkg.Types.PackageSpec}; io::Base.TTY, kwargs::@Kwargs{})
    @ Pkg.API ~/.julia/juliaup/julia-1.10.2+0.x64.linux.gnu/share/julia/stdlib/v1.10/Pkg/src/API.jl:159
 [13] develop(pkgs::Vector{Pkg.Types.PackageSpec})
    @ Pkg.API ~/.julia/juliaup/julia-1.10.2+0.x64.linux.gnu/share/julia/stdlib/v1.10/Pkg/src/API.jl:148
 [14] develop
    @ ~/.julia/juliaup/julia-1.10.2+0.x64.linux.gnu/share/julia/stdlib/v1.10/Pkg/src/API.jl:146 [inlined]
 [15] develop(; name::Nothing, uuid::Nothing, version::Nothing, url::Nothing, rev::Nothing, path::String, mode::Pkg.Types.PackageMode, subdir::Nothing, kwargs::@Kwargs{})
    @ Pkg.API ~/.julia/juliaup/julia-1.10.2+0.x64.linux.gnu/share/julia/stdlib/v1.10/Pkg/src/API.jl:176
 [16] top-level scope
    @ none:4
```

gdalle:
Documentation passes, the rest is not on me

vchuravy (approved):

docs/src/faq.md
```diff
@@ -0,0 +1,401 @@
...
+DocTestSetup = quote
+    using Enzyme
+end
+```
+
+# Frequently asked questions
+
+## Implementing pullbacks
+
+Enzyme's [`autodiff`](@ref) function can only handle functions with scalar output. To implement pullbacks (back-propagation of gradients/tangents) for array-valued functions, use a mutating function that returns `nothing` and stores it's result in one of the arguments, which must be passed wrapped in a [`Duplicated`](@ref).
```
wsmoses, line 12: This is only the case in (combined) reverse mode. Split reverse mode (aka autodiff_thunk) and forward mode do handle non-scalar output

docs/src/faq.md
```diff
@@ -0,0 +1,401 @@
...
+1-element SparseVector{Float64, Int64} with 1 stored entry:
+  [1]  =  1.0
+```
+
+Sometimes, determining how to perform this zeroing can be complicated.
+That is why Enzyme provides a helper function `Enzyme.make_zero` that does this automatically.
+
+## Activity analysis
+
+If you pass in any temporary storage which may be involved in an active computation to a function you want to differentiate, you must also pass in a duplicated temporary storage for use in computing the derivatives. For example, consider the following function which uses a temporary buffer to compute the result.
```
wsmoses, line 105: We have internal error messages that point to the docs location, can these be updated as well

docs/src/faq.md
```diff
@@ -0,0 +1,401 @@
...
+
+Enabling runtime activity does therefore, come with a sharp edge, which is that if the computed derivative of a function is mutable, one must also check to see if the primal and shadow represent the same pointer, and if so the true derivative of the function is actually zero.
+
+Generally, the preferred solution to these type of activity unstable codes should be to make your variables all activity-stable (e.g. always containing differentiable memory or always containing non-differentiable memory). However, with care, Enzyme does support "Runtime Activity" as a way to differentiate these programs without having to modify your code.
+
+## Mixed activity
+
```
wsmoses, line 179: same link from errror message comment

docs/src/faq.md
```diff
@@ -0,0 +1,401 @@
...
+    tup = (x, vec)
+    tup_ref = Ref(tup)
+    Base.inferencebarrier(myfirst_ref)(tup_ref)::Float64
+end
+
+Enzyme.autodiff(Reverse, f2, Active, Active(3.1))
+```
+
+## Complex numbers
+
```
wsmoses, line 224: same link from error message comment

docs/src/faq.md
gdalle: I'm not the one who wrote this part, I only copy-pasted it from where it was in a specific page to a section of the FAQ

Maybe we save modifications there for another PR?

docs/src/faq.md
gdalle: Is it really wise to point to a link in the docs with an error message? Seems rather brittle, as this PR demonstrates

wsmoses: Yeah I think so. Since otherwise folks will open issues with questions (which are directly answered in the docs)

gdalle: ok I'll fix the link

docs/src/faq.md
gdalle: added it

docs/src/faq.md
gdalle: I only found one error message with that link, were there more?

gdalle:
I updated the link to the section of the docs

wsmoses:
I think a couple more are missing like https://github.com/EnzymeAD/Enzyme.jl/blob/778deecdc9c13f07c3415f0b89de19a33d8372e1/src/Enzyme.jl#L267

and

https://github.com/EnzymeAD/Enzyme.jl/blob/778deecdc9c13f07c3415f0b89de19a33d8372e1/src/compiler.jl#L516

docs/src/faq.md
```diff
@@ -0,0 +1,402 @@
...
+CurrentModule = Enzyme
+DocTestSetup = quote
+    using Enzyme
+end
+```
+
+# Frequently asked questions
+
+## Implementing pullbacks
+
```
wsmoses, line 11: Maybe also say or forwardmode?

docs/src/faq.md
```diff
@@ -0,0 +1,402 @@
...
+In combined reverse mode, Enzyme's [`autodiff`](@ref) function can only handle functions with scalar output (this is not true for split reverse mode, aka `autodiff_thunk`).
+To implement pullbacks (back-propagation of gradients/tangents) for array-valued functions, use a mutating function that returns `nothing` and stores it's result in one of the arguments, which must be passed wrapped in a [`Duplicated`](@ref).
```
wsmoses, line 13: Maybe add However, regardless of AD mode, it often much more efficient to use mutation, both for your original code, as well as the derivative. To implement pullbacks ...., consider using a mutating...

docs/src/faq.md
```diff
@@ -0,0 +1,402 @@
...
+R ≈ A * B            &&
+∂z_∂A ≈ ∂z_∂R0 * B'   &&  # equivalent to Zygote.pullback(*, A, B)[2](∂z_∂R)[1]
+∂z_∂B ≈ A' * ∂z_∂R0       # equivalent to Zygote.pullback(*, A, B)[2](∂z_∂R)[2]
+```
+
+Note that the result of the backpropagation is *added to* `∂z_∂A` and `∂z_∂B`, they act as accumulators for gradient information.
+
+## Identical types in `Duplicated`
+
+Enzyme forces `x` and `∂f_∂x` to have the same types when constructing objects of type `Duplicated`, `DuplicatedNoNeed`, `BatchDuplicated`, etc.
```
wsmoses, line 59: Maybe rather than forces, the word checks would be more precise.

docs/src/faq.md
gdalle: done

docs/src/faq.md
gdalle: did the others

docs/src/faq.md
gdalle: wdym?

docs/src/faq.md
gdalle: done

docs/src/faq.md
```diff
@@ -0,0 +1,402 @@
...
+This is not a mathematical requirement, but rather a guardrail to prevent user error.
```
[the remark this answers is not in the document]
gdalle: replaced with practical because it is internal after all

docs/src/faq.md
gdalle: done

docs/src/faq.md
```diff
@@ -0,0 +1,402 @@
...
+The memory locations of `x` and `∂f_∂x` are accessed in the same way by the differentiation code, so they should have the same data layout.
```
wsmoses, line 62: More precisely, "the memory locations of the shadow memory `df_dx` may only be accessed in the derivative function df, if the original function `f` accessed the corresponding memory location in `x`. Ensuring the original variable x and shadow dx have the same type ensures that they have the same data layout and can help reduce errors where a user provides a shadow variable that cannot be accessed at the corresponding memory location."

docs/src/faq.md
gdalle: I think it would be confusing because we speak about combined reverse mode just below. Pushforwards are readily available with Enzyme.Forward

docs/src/faq.md
```diff
@@ -0,0 +1,402 @@
...
+Equality of types is an approximation of this condition.
+
+In some ways, type equality is too strict: two different types can have the same data layout.
```
wsmoses, line 65: perhaps add, for example Tuple{Float64, Float64} or `struct Pair
  x::Float64
  y::Float64
` have the same datalayout Enzyme would produce the expected results when given a primal of a Pair and a shadow of the tuple of floats.

gdalle: My other example idea was a vector vs a matrix column view

gdalle: That's the one which really is a pain in the butt for me atm

gdalle: My code is riddled with stuff like
```julia
function DI.value_and_pushforward(backend::AutoForwardEnzyme, f, x, dx, extras::Nothing)
    dx_sametype = convert(typeof(x), dx)
    y, dy = autodiff(backend.mode, f, Duplicated, Duplicated(x, dx_sametype))
    return y, dy
end
```

docs/src/faq.md
```diff
@@ -0,0 +1,402 @@
...
+But in other ways it is not strict enough.
+For instance, if `x` and `∂f_∂x` are sparse arrays, their sparsity pattern should be identical, but this is not encoded in the type.
```
wsmoses, line 67: Maybe we should use an even easier example here. Suppose you have a function `f(x) = x[7]`. If you called Enzyme.autodiff(Reverse, f, Duplicated(ones(10), ones(1))`, the type check alone would not be sufficient. Since the original code loaded from x[7], the derivative code will set `dx[7]`. Since the length wasn't encoded in the type, Julia wasn't able to provide a high-level error before running autodiff, and the user may end up with a segfault (or other memory error) when running the generated derivative code.

gdalle: done

docs/src/faq.md
gdalle: done

docs/src/faq.md
gdalle: added the vector example

docs/src/faq.md
```diff
@@ -0,0 +1,402 @@
...
+To make sure that `∂f_∂x` has the right data layout, create it with `∂f_∂x = Enzyme.make_zero(x)`.
+
```
wsmoses, line 76: It would be nice to include the extra stuff I mentioned in the issue here.

"""
Advanced users may leverage Enzyme's memory semantics (only touching locations in the shadow that were touched in the primal) for additional performance/memory savings, at the obvious cost of potential safety if used incorrectly.

Consider the following function that loads from offset 47 of a Ptr

```julia
function f(ptr)
	x = unsafe_load(ptr, 47)
	x * x
end

ptr = Base.reinterpret(Ptr{Float64}, Libc.malloc(100*sizeof(Float64)))
unsafe_store!(ptr, 3.14, 47)

@show f(ptr)
# 3.14 * 3.14
```

The recommended (and guaranteed sound) way to differentiate this is to pass in a shadow pointer that is congruent with the primal. That is to say, its length (and recursively for any sub types) are equivalent to the primal.

```julia

ptr = Base.reinterpret(Ptr{Float64}, Libc.malloc(100*sizeof(Float64)))
unsafe_store!(ptr, 3.14, 47)
dptr = Base.reinterpret(Ptr{Float64}, Libc.calloc(100*sizeof(Float64), 1))

autodiff(Reverse, f, Duplicated(ptr, dptr))

@show unsafe_load(dptr, 47)
# 2 * 3.14
```

However, since we know the original function only reads from one float64, we could choose to only allocate a single float64 for the shadow, as long as we ensure that loading from offset 47 (the only location accessed) is in bounds.

```julia
ptr = Base.reinterpret(Ptr{Float64}, Libc.malloc(100*sizeof(Float64)))
dptr = Base.reinterpret(Ptr{Float64}, Libc.calloc(sizeof(Float64), 1))

# offset the pointer to have unsafe_load(dptr, 47) access the 0th byte of dptr
# since julia one indexes we subtract 46 * sizeof(Float64) here
autodiff(Reverse, f, Duplicated(ptr, dptr - 46 * sizeof(Float64)))

# represents the derivative of the 47'th elem of ptr, 
@show unsafe_load(dptr, 1)
# 2 * 3.14
```

However, this style of optimization is not specific to Enzyme, or AD, as one could have done the same thing on the primal code where it only passed in one float. The difference, here however, is that performing these memory-layout tricks safely in Enzyme requires understanding the access patterns of the generated derivative code -- like discussed here.

```

ptr = Base.reinterpret(Ptr{Float64}, Libc.calloc(sizeof(Float64), 1))
unsafe_store!(ptr, 3.14)
# offset the pointer to have unsafe_load(ptr, 47) access the 0th byte of dptr
# again since julia one indexes we subtract 46 * sizeof(Float64) here
@show f(ptr - 46 * sizeof(Float64)))
# 3.14 * 3.14
```
"""

gdalle: As I told you, I can only contribute stuff I understand :) If you want to add it, be my guest, but perhaps in a separate section of the FAQ? This part about `Duplicated` we just worked on feels reasonably clear to me, and adding this blurb would instantly make it super scary to me and a lot of other people

gdalle: To put you in my shoes as a rather experienced Julia and AD user, I have never heard of:
- `unsafe_load`, `unsafe_store!`
- `Ptr`
- `Base.reinterpret`
- `Libc.malloc`

so this instantly becomes way out of my depth

docs/src/faq.md
```diff
@@ -0,0 +1,402 @@
...
+## CUDA support
+
+[CUDA.jl](https://github.com/JuliaGPU/CUDA.jl) is only supported on Julia v1.7.0 and onwards. On v1.6, attempting to differentiate CUDA kernel functions will not use device overloads
+correctly and thus returns fundamentally wrong results.
```
wsmoses, line 80: Maybe also add. "Specifically, differentiating within device kernels is supported. Differentiating through a heterogeneous code presently requires defining a custom derivative that tells Enzyme that differentiating an @cuda call is done by performing @cuda of its generated derivative. Automating this for a better experience for CUDA.jl requires an update to CUDA.jl (https://github.com/JuliaGPU/CUDA.jl/pull/1869/files), and is now available for Kernel Abstractions.

Differentiating host-side code when accesses device memory (e.g. `sum(CuArray)` is not yet supported, but in progress

docs/src/faq.md
```diff
@@ -0,0 +1,402 @@
...
+# output
+
+1-element SparseVector{Float64, Int64} with 1 stored entry:
+  [1]  =  1.0
+```
+
+Sometimes, determining how to perform this zeroing can be complicated.
+That is why Enzyme provides a helper function `Enzyme.make_zero` that does this automatically.
+
+## Activity analysis
```
wsmoses, line 112: This is much more specific than just activity analysis, so probably keeping the old name is more apt, or alternatively adding much more to describe what activity analysis is.

docs/src/faq.md
wsmoses: Yeah, but does prefacing it with the "Advanced users may..." not suffice?

docs/src/faq.md
gdalle: I was trying to find a name that fits nicely in the TOC on the left side, that's why I shortened it. And it is also the first time in the docs that activity analysis is mentioned at all

docs/src/faq.md
wsmoses: But yeah maybe copy this into a separate FAQ item that comes right after on "Advanced Memory Layout" and change the title of the first one to "Identical types in `Duplicated` / Memory Layout"

docs/src/faq.md
gdalle: what is heterogeneous code?

docs/src/faq.md
gdalle: But I can roll it back if necessary

docs/src/faq.md
wsmoses: Ah I forgot the title of this was "implementing pullbacks", yeah I just meant that duplicated returns are valid in split mode and also forward mode (but not in combined reverse mode atm)

docs/src/faq.md
gdalle: > Yeah, but does prefacing it with the "Advanced users may..." not suffice?

Your call, but I find it more reassuring when, as a beginner, I can understand the whole section. It also helps identify which parts of the page are relevant to me and which ones can be skipped at first glance

gdalle:
going to bed now, it's 11 in Switzerland ;) will pick it up tomorrow with your remarks! thanks for helping me out on this

docs/src/faq.md
wsmoses: Yeah so the fix for your matrix view stuff is to actually leverage the "advanced memory layout" stuff discussed below. Aka if an array just stores values 0...N, and a Matrix stores row 0 elems in 0...N, etc. You can technically force Enzyme to take a Duplicated(Array, Matrix) and it will store the derivative in row 0 (assuming the layout of julia arrays is what I expect). You can then force it to store in row 1, by passing in Duplicated(Array, unsafe_array(unsafe_pointer(Matrix) + N * sizeof(eltype))). [since then it will store the derivative data at offset 0 to the shadow, which is equivalent to starting at element N of your array, aka at Row 1, col 0].

wsmoses:
> going to bed now, it's 11 in Switzerland ;) will pick it up tomorrow with your remarks! thanks for helping me out on this

And thank you so much for working on this! Hopefully together we can make this easier for folks to use/understand!

docs/src/faq.md
wsmoses: cc @vchuravy for thoughts here (also see my linked comment https://github.com/EnzymeAD/Enzyme.jl/pull/1343#discussion_r1532946312)

wsmoses:
Perhaps in another follow up PR, but @gdalle can I distract you further into helping digest https://discourse.julialang.org/t/how-do-you-speed-up-the-linear-sparse-solver-in-zygote/111801/36?u=wsmoses into additional and accessible docs?

docs/src/faq.md
gdalle: renamed to "Activity of temporary storage" and fixed the link

gdalle:
I put the advanced details @wsmoses suggested to add in a separate issue #1355 

Since this PR is first a foremost moving stuff around, I suggest we merge it and come back to the content at a later time. What do you think?

docs/src/faq.md
gdalle: see https://github.com/EnzymeAD/Enzyme.jl/issues/1355

docs/src/faq.md
gdalle: see https://github.com/EnzymeAD/Enzyme.jl/issues/1355

wsmoses (approved):

andrinr:
Would it not be possible to check weather on or two of the arrays are views and then not raise the error? I'm fairly new to Julia, but I assume using a view is oftentimes a best practice. Especially when using an iterative solver, where the entire history of forward results of x need to be stored.

Issue #1334, opened 2024-03-07
Supporting covariant derivatives


MasonProtter:
## The problem:

Currently, when you compute the reverse mode derivative of some function of a `struct MyStruct` in enzyme, i.e.  `autodiff(Reverse, f, Duplicated(A::MyStruct, dA::MyStruct))`, the object `dA` must be an identical type to `A`, but it should be interpreted in a very different way from `A`. What `dA` actually is, is the object you get by treating `MyStruct` as a Cartesian vector whose elements are the struct's fields. 

The `i`th field of `dA` will be given by  $\mathrm{d}A_{i} = {\partial f / \partial A_{i}}$

The problem arises though, that often what people want is not this component-wise Cartesian derivative, but a [**covariant** derivative](https://en.wikipedia.org/wiki/Covariant_derivative), i.e. the mathematically meaningful derivative on the manifold. 

Here's a concrete example posted by @simsurace in Slack:

```julia
julia> using Enzyme, LinearAlgebra

julia> A = Symmetric(rand(3, 3)); dA = make_zero(A);

julia> Enzyme.autodiff(Reverse, sum, Duplicated(A, dA))
((nothing,),)

julia> dA
3×3 Symmetric{Float64, Matrix{Float64}}:
 1.0  2.0  2.0
 2.0  1.0  2.0
 2.0  2.0  1.0
```
compared to the answer given by Zygote:
```julia
julia> using Zygote

julia> Zygote.gradient(sum, A) |> only
3×3 Symmetric{Float64, FillArrays.Fill{Float64, 2, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}}}:
 1.0  1.0  1.0
 1.0  1.0  1.0
 1.0  1.0  1.0
```

What's going on here? Enzyme is treating `A::Symmetric` as a vector with only one differentiable field (`A.data`), and so it computes

$$
dA_{\mathrm{data}}[a, b] = {\partial \over \partial A_{\mathrm{data}}[a, b]}\sum_{i,j} A[i,j] \\
= {\partial \over \partial A_{\mathrm{data}}[a, b]} \sum_{i,j} 
\begin{cases}
    A_{\mathrm{data}}[i, j] & j \geq i \\
    A_{\mathrm{data}}[j, i],              & \text{otherwise}
\end{cases}
$$

Now, since `A.data[i,j]` is never accessed for `j < i`, this evaluates to

$$
[...]

PR #1362, opened 2024-03-26
RFC: strip most types from `gradient` output

mcabbott:
This is a draft of a way to start addressing #1334, for comment.

It implements what I called level 2 here: https://github.com/EnzymeAD/Enzyme.jl/issues/1334#issuecomment-2016573247

On arrays like these, no change. Natural and structural representations agree:
```julia
julia> Enzyme.gradient(Reverse, first, Diagonal([1,2.]))
2×2 Diagonal{Float64, Vector{Float64}}:
 1.0   ⋅ 
  ⋅   0.0

julia> using SparseArrays, StaticArrays

julia> Enzyme.gradient(Reverse, sum, sparse([5 0 6.]))
1×3 SparseMatrixCSC{Float64, Int64} with 2 stored entries:
 1.0   ⋅   1.0

julia> Enzyme.gradient(Reverse, sum, PermutedDimsArray(sparse([1 2; 3 0.]), (2,1)))
2×2 PermutedDimsArray(::SparseMatrixCSC{Float64, Int64}, (2, 1)) with eltype Float64:
 1.0  1.0
 1.0  0.0

julia> Enzyme.gradient(Reverse, first, reshape(SA[1,2,3,4.]',2,2))
2×2 reshape(adjoint(::SVector{4, Float64}), 2, 2) with eltype Float64:
 1.0  0.0
 0.0  0.0
```

On arrays like these, it does not know how to construct the natural representation, so doesn't try:
(I know how, but the fields of the result will not line up with the existing ones.)
```julia
julia> Enzyme.gradient(Reverse, sum, Symmetric(rand(3,3)))
(data = [1.0 2.0 2.0; 0.0 1.0 2.0; 0.0 0.0 1.0], uplo = nothing)

julia> Enzyme.gradient(Reverse, first, reshape(LinRange(1,2,4)',2,2))
(parent = (parent = (start = 1.0, stop = 0.0, len = nothing, lendiv = nothing),), dims = (nothing, nothing), mi = ())
```

Arrays of non-diff objects cannot be wrapped up in array structs:
```julia
julia> Enzyme.gradient(Reverse, float∘first, Diagonal([1,2,3]))
(diag = nothing,)

julia> Enzyme.gradient(Reverse, float∘first, SA[1,2,3]')
(parent = (data = (nothing, nothing, nothing),),)
```

`make_zeros` uses an IdDict cache to preserve identity between different branches of the struct.
At present this does not... 
```julia
julia> mutable struct TwoThings{A,B}; a::A; b::B; end

julia> nt = (x=TwoThings(3.0, 4.0), y=TwoThings(3.0, 4.0));

julia> nt.x === nt.y
false

julia> grad = Enzyme.gradient(Reverse, nt -> nt.x.a + nt.y.a + 20nt.x.b + 20nt.y.b, nt)
(x = (a = [11.0, 11.0], b = 22.0), y = (a = [11.0, 11.0], b = 22.0))

julia> grad.x === grad.y  # new identity created
true

# example 2

julia> arrs = [[1,2.], [3,4.]];

julia> grad = Enzyme.gradient(Reverse, nt -> sum(sum(sum, x)::Float64 for x in nt), (a = arrs, b = arrs))
(a = [[2.0, 2.0], [2.0, 2.0]], b = [[2.0, 2.0], [2.0, 2.0]])

julia> grad.a === grad.b  # container array identity is not preserved
false

julia> grad.a[1] === grad.b[1]  # array of numbers
true
```

A simple Flux model, no functional change, just looks different to the model:
```julia
julia> using Flux

julia> model = Chain(Embedding(reshape(1:6, 2,3) .+ 0.0), softmax);

julia> Enzyme.gradient(Reverse, m -> sum(abs2, m(1)), model)
(layers = ((weight = [-0.18171549534589682 0.0 0.0; 0.18171549534589682 0.0 0.0],), nothing),)
```

Comments:

* I'm not sure what I think about the failure to preserve `===` relations between some mutable objects in the original gradient. Some of this could be solved by adding an IdDict cache like `make_zeros` does. 

* The function called `strip_types` for now probably needs to be public, so that you can call it yourself after constructing `dx = make_zero(x)`, and so that you can overload it for your array wrappers.

* Projecting things like Symmetric to their covariant representation probably needs to be opt-in, by somehow telling `gradient` that you want this. (That's level 4 here: https://github.com/EnzymeAD/Enzyme.jl/issues/1334#issuecomment-2016573247 .) Could be implemented as additional methods of this function, something like `strip_types(x, dx, Val(true), cache)`?

* Surely all the code is in the wrong place, and needs tests.

### DIFF
--- a/src/Enzyme.jl
+++ b/src/Enzyme.jl
@@ -904,7 +904,8 @@ This will allocate and return new array `make_zero(x)` with the gradient result.
 Besides arrays, for struct `x` it returns another instance of the same type,
 whose fields contain the components of the gradient.
 In the result, `grad.a` contains `∂f/∂x.a` for any differential `x.a`,
-while `grad.c == x.c` for other types.
+while `grad.c == nothing` for other types.
+The result is post-processed to remove types from `make_zero(x)`.
 
 Examples:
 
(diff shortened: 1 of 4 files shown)


src/Enzyme.jl
```diff
@@ -925,18 +926,112 @@ grad = gradient(Reverse, only ∘ f, (a = 2.0, b = [3.0], c = "str"))
...
+# strip_types(x::Base.TwicePrecision{<:GoodNum}) = x
+
+# Non-differentiable types
+strip_types(x::Integer) = nothing # zero(x)
+strip_types(x::Union{Symbol, Char, AbstractString, Nothing}) = nothing # x
+strip_types(x::Array{<:Integer}) = nothing
+strip_types(x::Array{<:Union{Symbol, Char, AbstractString, Nothing}}) = nothing
+
+# Containers to recurse into
+strip_types(x::Union{Tuple, NamedTuple, Array}) = map(strip_types, x)  # need to worry about undef?
```
wsmoses, line 996: This won't work, for example, for a recursive type (and will instead infinite loop presumably?)

PR #1369, opened 2024-03-28
Load Adapt unconditionally for Julia 1.8

vchuravy:


vchuravy:
Needs backport to EnzymeCore 0.6

wsmoses:
Some folks in the past were hesitant about adding EnzymeCore as a dep due to Adapt as a sub-dependency, so I'm a bit hesitant here.

vchuravy:
So the question is what is their concern? Is it load time? If so package extensions solves that but it can only solve it on 1.9+

I strongly think that we shouldn't have broken functionality on 1.8 and lower

lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -257,4 +257,9 @@ function compiler_job_from_backend end
 
 include("rules.jl")
 
+if !isdefined(Base, :get_extension)
```
vchuravy, line 260: @wsmoses the original PR by @devmotion had the backwards compatibility https://github.com/EnzymeAD/Enzyme.jl/pull/954/files#diff-94319485f88a6e9c6ce4669ce81ce9d3a46dab48d8c9e579b22272fc798a39a9R172

lib/EnzymeCore/Project.toml
```diff
@@ -7,11 +7,14 @@ version = "0.7.1"
 Adapt = "3, 4"
 julia = "1.6"
 
+[deps]
+Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
```
wsmoses, line 11: To confirm, this doesn't add the dep on 1.9+?

wsmoses:
Okay. I'm a bit uneasy on this still, but I defer to your judgement.

wsmoses (approved):

lib/EnzymeCore/ext/AdaptExt.jl
```diff
@@ -1,11 +1,11 @@
 module AdaptExt
-    using Adapt
-    using EnzymeCore
+isdefined(Base, :get_extension) ? (using Adapt) : (using ..Adapt)
```
devmotion, line 2: If the extension is loaded unconditionally on older Julia versions, you can just use
```suggestion
using Adapt
```
(and not load Adapt in EnzymeCore itself). See #954.

lib/EnzymeCore/ext/AdaptExt.jl
```diff
@@ -1,11 +1,11 @@
...
+isdefined(Base, :get_extension) ? (using EnzymeCore) : (using ..EnzymeCore)
```
devmotion, line 3:
```suggestion
using EnzymeCore
```

lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -257,4 +257,9 @@ function compiler_job_from_backend end
...
+    using Adapt
```
devmotion, line 261:
```suggestion
```

lib/EnzymeCore/Project.toml
devmotion: No, on Julia >= 1.9 it will still be only a weak dependency.

PR #1371, opened 2024-03-31
Allow custom rule for constant arg/ret in rev mode

wsmoses:


michel2323:
This works with arrays. I've checked now with KA using `CPU()` and `CUDABackend()`. What is weird is that with `CUDABackend()` the rule gets triggered, however with `CPU()` not. 

Also, in the case of `CUDABackend()` I get lots of these, with `CPU()` not:
I get lots of:
```
┌ Warning: Type does not have a definite number of fields
│   T = Tuple{Vararg{Union{UInt64, String}}}
└ @ Enzyme /disk/mschanen/julia_depot/packages/GPUCompiler/U36Ed/src/utils.jl:59
```

MWE is [here](https://github.com/michel2323/ka_enzyme/blob/main/mwe.jl) (local Enzyme_jll with Enzyme build is needed)

michel2323:
Could it be that Enzyme takes decisions based on the code `synchronize(backend::T) where T`? I think in the case of `CUDABackend()` it goes through the `KA.synchronize(::CUDABackend) = CUDA.synchronize()`.

michel2323:
Narrowed it down.

```julia
using Enzyme
using EnzymeCore
using EnzymeCore.EnzymeRules

struct MyConst end
struct MyConst2
    v::Vector{Float64}
end
MyConst2() = MyConst2(zeros(5))

bar(x::MyConst)::Nothing = nothing
function bar(x::MyConst2)
    x.v .*= 2.0
    nothing
end

function foo(myconst)
    bar(myconst)
    return nothing
end

function EnzymeRules.augmented_primal(
    config::Config,
    func::Const{typeof(bar)},
    ::Type{Const{Nothing}},
    myconst::T
) where T <: EnzymeCore.Annotation
    println("bar aug_fwd rule $(typeof(myconst))")
    return AugmentedReturn{Nothing, Nothing, Any}(
        nothing, nothing, (nothing)
    )
end

function EnzymeRules.reverse(
    config::Config,
    func::Const{typeof(bar)},
    ::Type{Const{Nothing}},
    tape,
    myconst
)
    println("bar rev rule $(typeof(myconst))")
    return (nothing,)
end

function driver(myconst)
    println("Running $(typeof(myconst()))")
    Enzyme.autodiff(
        ReverseWithPrimal, foo, Const(myconst())
    )
end

# Doesn't trigger rules above
driver(MyConst)
# Triggers rules above
driver(MyConst2)
```

This outputs
```
Running MyConst
Running MyConst2
bar aug_fwd rule Const{MyConst2}
bar rev rule Const{MyConst2}
```
when the rule should also be applied in the case of MyConst.
```

michel2323:
As per our chat, this should be all resolved and ready to be merge.

Issue #1387, opened 2024-04-15
Overly broad inactive marker for Random.rand


gaurav-arya:
It is quite common for other packages (e.g. `Distributions.jl`) to overload `Random.rand` with parameter-dependent distributions, e.g.
```julia
using Enzyme
import Random

# N(x, 1) 
struct MyDistribution
    x::Float64
end

Random.rand(rng::Random.AbstractRNG, d::MyDistribution) = d.x + randn()
Random.rand(d::MyDistribution) = rand(Random.default_rng(), d)

@show autodiff(Enzyme.Reverse, x -> rand(MyDistribution(x)), Active, Active(1.0)) # ((0.0,),)
```
or even
```julia
struct DiracDelta 
    x::Float64
end

Random.rand(rng::Random.AbstractRNG, d::DiracDelta) = d.x
Random.rand(d::DiracDelta) = rand(Random.default_rng(), d)

@show autodiff(Enzyme.Reverse, x -> rand(DiracDelta(x)), Active, Active(1.0)) # ((0.0,),)
```
Zero gradients are returned because the outer `rand` call is marked inactive:

https://github.com/EnzymeAD/Enzyme.jl/blob/1bf16f8217f2f0e516666f5dff2deb27a653302d/src/internal_rules.jl#L69

PR #1388, opened 2024-04-15
Try narrowing inactive rules for `rand`

gaurav-arya:
Addresses #1387 

These new rules could potentially be too narrow for some cases. Also, for reference, here are the ones in `ChainRules.jl`: https://github.com/JuliaDiff/ChainRules.jl/blob/f9f0722558e065ee8bbdf042c2d99a9d56ba0070/src/rulesets/Random/random.jl#L15-L25. It seemed like we might not need some of them, and also the generic `rand!` ones there would be dangerous for Enzyme when differentiating parameter-dependent mutable sampling.


wsmoses (approved):
LGTM, maybe add a test for the case you opened an issue on?

Issue #1417, opened 2024-05-07
Autodiff Deferred Thunk is broken


wsmoses:
@vchuravy @michel2323 I'm not sure what's happening here. The non-deferred version succeeeds. This is a minimization of the segfault from  https://github.com/JuliaGPU/KernelAbstractions.jl/pull/476

```julia
using KernelAbstractions
using Test
using Enzyme
using EnzymeCore
# using KernelAbstractions.EnzymeExt
Enzyme.API.printall!(true)

@kernel function square!(A)
    I = @index(Global, Linear)
    @inbounds A[I] *= A[I]
end


    A = Array{Float64}(undef, 64)
    dA = Array{Float64}(undef, 64)

    A .= (1:1:64)
    dA .= 1

    import KernelAbstractions: Kernel, StaticSize, launch_config, __groupsize, __groupindex, blocks, mkcontext, CompilerMetadata, CPU

    function aug_fwd(ctx, f::FT, ::Val{ModifiedBetween}, args...) where {ModifiedBetween, FT}
        TapeType = EnzymeCore.tape_type(ReverseSplitModified(ReverseSplitWithPrimal, Val(ModifiedBetween)), Const{Core.Typeof(f)}, Const, Const{Core.Typeof(ctx)}, map(Core.Typeof, args)...)
        @show TapeType
        forward, reverse = EnzymeCore.autodiff_deferred_thunk(ReverseSplitModified(ReverseSplitWithPrimal, Val(ModifiedBetween)), TapeType, Const{Core.Typeof(f)}, Const, Const{Core.Typeof(ctx)}, map(Core.Typeof, args)...)
        
        # Non deferred works
        # forward, reverse = EnzymeCore.autodiff_thunk(ReverseSplitModified(ReverseSplitWithPrimal, Val(ModifiedBetween)), Const{Core.Typeof(f)}, Const, Const{Core.Typeof(ctx)}, map(Core.Typeof, args)...)
        forward(Const(f), Const(ctx), args...)[1]
        return nothing
    end

    ndrange = (1,)
    workgroupsize = nothing
    func = Enzyme.Const(square!(CPU()))
        kernel = func.val
        f = kernel.f

        ndrange, workgroupsize, iterspace, dynamic = launch_config(kernel, ndrange, workgroupsize)


        # TODO autodiff_deferred on the func.val
        ModifiedBetween = Val((false, false, true))

        aug_kernel = similar(kernel, aug_fwd)

[...]

PR #1426, opened 2024-05-11
fix deferred

wsmoses:


vchuravy:
Can you add the test from #1417

Issue #1437, opened 2024-05-13
StaticArrays + Onehot mismatch


wsmoses:
similar creates an marray rather than sarray 

https://discourse.julialang.org/t/noise-with-gradient-works/114178/10

PR #1438, opened 2024-05-13
Fix static arrays on forward mode gradient call

wsmoses:


wsmoses:
Fixes https://github.com/EnzymeAD/Enzyme.jl/issues/1437

PR #1443, opened 2024-05-13
Revamp interpreter

michel2323:


wsmoses:
@vchuravy @michel2323 is this to be able to automatically upgrade the use of autodiff in nested AD into autodiff deferred?

That would be amazing.

vchuravy:
Yeah that's the goal

wsmoses:
bump here @michel2323 

FYI https://github.com/EnzymeAD/Reactant.jl/pull/40 may be helpful here

wsmoses:
@vchuravy what else is needed here?

PR #1462, opened 2024-05-22
Fix unused shadow return type

wsmoses:


wsmoses:
Fixes https://github.com/EnzymeAD/Enzyme.jl/issues/1461

PR #1463, opened 2024-05-22
Faster type analysis (no JIT)

wsmoses:


wsmoses:
Fixes https://github.com/EnzymeAD/Enzyme.jl/issues/1461

wsmoses:
@swilliamson7 @vchuravy  I wonder if/how much this speeds things up (though is useful even without for simplifying debugging)

PR #1470, opened 2024-05-25
Nicer CUDA errors

wsmoses:


wsmoses:
Resolves https://github.com/EnzymeAD/Enzyme.jl/issues/1102

PR #1478, opened 2024-05-27
Update `CompatHelper` workflow

prbzrg:
I added other directories to check by `CompatHelper`. Also, I enabled the check for jlls.
Before merging this, you should know that at first, it will create many PRs.

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -8,19 +8,28 @@ jobs:
   CompatHelper:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - name: Get Julia compatibility
         id: julia_compat
         # NOTE: this requires a julia compat lower-bound with minor version!
         run : |
           version=$(grep '^julia = ' Project.toml | grep -o '".*"' | cut -d '"' -f2)
           echo "::set-output name=version::$version"
-      - uses: julia-actions/setup-julia@latest
+      - uses: julia-actions/setup-julia@v2
         with:
           version: ${{ steps.julia_compat.outputs.version }}
-      - name: Pkg.add("CompatHelper")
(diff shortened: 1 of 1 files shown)


.github/workflows/CompatHelper.yml
```diff
@@ -8,19 +8,28 @@ jobs:
...
+      - name: Pkg.add
+        shell: julia --color=yes {0}
+        run: |
+          import Pkg
+          Pkg.add("CompatHelper")
+      - name: CompatHelper.main
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: julia -e 'using CompatHelper; CompatHelper.main()'
+          COMPATHELPER_PRIV: ${{ secrets.DOCUMENTER_KEY }}
```
vchuravy, line 31: Why this change?

vchuravy, line 31:
```suggestion
```

prbzrg: If I remember correctly, without it, there was an error about permissions. Let's delete it for now.

prbzrg: Albeit, `CompatHelper` has it
https://github.com/JuliaRegistries/CompatHelper.jl/blob/e72bd6b815a69d14ccdb0c85b3589bed2a2af704/.github/workflows/CompatHelper.yml#L44

vchuravy (approved):

wsmoses:
@vchuravy @prbzrg can this not open PR's for test which have compat for the main repo?

We shouldn't keep separate compats for both

vchuravy:
Yeah I agree that is suboptimal.

vchuravy:
There is an upstream issue for that: https://github.com/JuliaRegistries/CompatHelper.jl/issues/294

vchuravy:
I disabled it for `test` again

PR #1497, opened 2024-06-06
Workaround breakage in llvm.jl

wsmoses:


src/api.jl
```diff
@@ -104,6 +104,7 @@ struct CFnTypeInfo
 end
 
 
+@static isdefined(LLVM, InstructionMetadataDict)
```
vchuravy, line 107:
```suggestion
@static if isdefined(LLVM, InstructionMetadataDict)
```

src/api.jl
```diff
@@ -104,6 +104,7 @@ struct CFnTypeInfo
 end
 
 
+@static if isdefined(LLVM, InstructionMetadataDict)
```
vchuravy, line 107:
```suggestion
@static if isdefined(LLVM, :InstructionMetadataDict)
```

vchuravy:
x-ref: https://github.com/maleadt/LLVM.jl/pull/414

vchuravy (approved):

wsmoses:
@vchuravy apparenrlt we need to backported to 11 too see slack

maleadt:
I pushed & tagged a deprecation, so this should be fixed for users.

I would recommend pirating internals like that though. It's also not clear to me why you're overriding the function in the first place, https://github.com/EnzymeAD/Enzyme/blob/79e829910a63618e1bb5843b8564031a8345305c/enzyme/Enzyme/CApi.cpp#L913-L919 seems wrong (metadata can be attached to function values, not only instructions).

wsmoses:
Yeah but we only needed it to attach to instructions. But it’s awesome that
we won’t need to vendor that

vchuravy:
> apparenrlt we need to backported to 11 too see slack

Enzyme.jl 0.11 only ever supported LLVM.jl v6 so no?

PR #1518, opened 2024-06-06
Make zero in place

wsmoses:
It's 2 am so I'lls tart this later, but something like this.

cc @ChrisRackauckas

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -19,7 +19,7 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 [compat]
 CEnum = "0.4, 0.5"
 ChainRulesCore = "1"
-EnzymeCore = "0.7.3"
+EnzymeCore = "0.7.4"
 Enzyme_jll = "0.0.119"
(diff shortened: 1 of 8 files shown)


Project.toml
```diff
@@ -19,7 +19,7 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 [compat]
 CEnum = "0.4, 0.5"
 ChainRulesCore = "1"
-EnzymeCore = "0.7.3"
+EnzymeCore = "0.7.3, 0.7.4"
```
vchuravy, line 22: This needs to be `0.7.4` since otherwise you will get a "EnzymeCore.make_zero!" not defined.

wsmoses: i was just wondering this

ChrisRackauckas:
I feel like make_zero and make_zero! should be moved out of Enzyme as a general Julia tool, even to Base as the standard `zero` function extension. I kind of just use it everywhere...

wsmoses:
Not opposed, but have to start somewhere.

PR #1545, opened 2024-06-18
Implement `pick_batchsize`

### LINKED ISSUE #1542: Heuristic for picking the chunk/batch size?
ForwardDiff has a heuristic for picking chunk size, with a default threshold of 12 dictated by memory bandwidth:

https://github.com/JuliaDiff/ForwardDiff.jl/blob/ff56092ed2960717ce45f53a90584898c232e74b/src/prelude.jl#L24-L34

https://github.com/JuliaDiff/ForwardDiff.jl/blob/ff56092ed2960717ce45f53a90584898c232e74b/src/prelude.jl#L8

Does Enzyme have something similar I could use?

gdalle:
Fix #1542 by implementing a constant heuristic for picking chunk size.


src/utils.jl
```diff
@@ -1,3 +1,13 @@
+"""
+    pick_chunksize()
+
+Return a reasonable chunk size for batched differentiation, given as `Val{chunk}()`.
+
+!!! warning
+    This function is experimental, and not part of the public API.
+"""
+pick_chunksize() = Val(16)
```
wsmoses, line 9: Can this take a mode, function type, and argument types?

wsmoses: and also possibly a val for the total number of iterations

gdalle: what do you mean by iterations?

wsmoses: Like we’re chunking a total of N iterations into chunks or size C

gdalle: are argument types sufficient? in general we may need the actual argument values, for instance to ensure that the chunk size is smaller than the input length

wsmoses: Yeah that’s why I also recommended passing in the total number of iterations too

gdalle: perhaps `dimension` is a better word?

wsmoses: maybe but that feels ambiguous in a different way

vchuravy: Why does this need to return a `Val`?

gdalle: Because Enzyme takes this parameter as a `Val` for `jacobian` and (forward) `gradient`, so I thought this would save the eventual users one conversion. In addition, because it is constant, it is also type-inferrable (for now).

vchuravy: >  In addition, because it is constant, it is also type-inferrable (for now).

If it ever becomes non-constant this will be a source of latency. Julia will constant-evaluate `Val(chunksize())` just fine, but the Val conversion should be left to the user.

gdalle: I added the arguments and a `Val` called `totalsize`, is that how you imagined it?

src/utils.jl
```diff
@@ -1,3 +1,20 @@
+"""
+    pick_chunksize(::Val{totalsize}, mode, ftype, argtypes...)
+
+Return a reasonable chunk size for batched differentiation.
+
+!!! warning
+    This function is experimental, and not part of the public API.
+"""
+function pick_chunksize(
+    ::Val{totalsize},
```
vchuravy, line 10: Don't overuse Val like this. If the compiler constant propagates it, it will happen in anyways. If it doesn't this causes dispatch for every possible totalsize parameter.

src/utils.jl
```diff
@@ -1,3 +1,20 @@
...
+Return a reasonable chunk size for batched differentiation.
+
+!!! warning
+    This function is experimental, and not part of the public API.
+"""
+function pick_chunksize(
+    totalsize::Integer,
+    mode::Mode,
+    ftype::Type,
+    argtypes::Vararg{Type{<:Annotation}, Nargs}
```
wsmoses, line 13: I think also you'll need the return type annotation here too

gdalle: done in the last commit

gdalle:
Should this be renamed to `pick_batchsize`? On the one hand, the type is `BatchDuplicated`, but on the other hand, `jacobian` and `gradient` call it chunk

wsmoses:
Sure, batchsize sgtm.

wsmoses:
Otherwise, this looks good to me once test pass

gdalle:
I think test failures are unrelated? The tests I introduced pass locally

gdalle:
is this acceptable? it's hard to know given all the failing tests

wsmoses:
@gdalle CI here seems really just blocked by this PR.


e.g.

```
BatchMixedDuplicated vector{float64} call |    3      3
Test Summary: | Pass  Total
TypeTree      |   20     20
ERROR: LoadError: LoadError: ArgumentError: Package EnzymeCore not found in current path:
- Run `import Pkg; Pkg.add("EnzymeCore")` to install the EnzymeCore package.

Stacktrace:
 [1] require(into::Module, mod::Symbol)
   @ Base ./loading.jl:893
 [2] include(fname::String)
   @ Base.MainInclude ./client.jl:444
 [3] top-level scope
   @ ~/work/Enzyme.jl/Enzyme.jl/test/runtests.jl:87
 [4] include(fname::String)
   @ Base.MainInclude ./client.jl:444
 [5] top-level scope
   @ none:6
in expression starting at /home/runner/work/Enzyme.jl/Enzyme.jl/test/utils.jl:2
```

gdalle:
My bad, I thought EnzymeCore was available for testing. I removed it from the test file

gdalle:
@wsmoses this should be good to go?

gdalle:
@vchuravy this looks like an easy merge?

src/utils.jl
```diff
@@ -1,3 +1,19 @@
+"""
```
wsmoses, line 1: Sorry, just noticing this wasn't merged.

But one more quick comment before then. Can you rename this file sugar.jl or something?

We already confusingly have several other utils.jl files in subdirectories that make this a bit confusing.

wsmoses (approved):

gdalle:
Is this mergeable?

wsmoses:
sorry for all the constant shenanigans, but sadly this now has a merge conflict. I'll merge right after and ping me if its not (sorry this functionality isn't currently used in Enzyme or fixes an issue so I keep forgetting to hit merge after walking away as CI starts)

gdalle:
Alternatively, or in complement to this PR, I've been thinking of adding a batch size parameter to the Enzyme AD type, so that it can be statically inferred like in AutoForwardDiff. But maybe I should read the Width from the mode object instead? Is it the same?

wsmoses:
Yeah the width parameter is indeed what you'd expect

gdalle:
On the other hand maybe this should be an adaptive function which dispatches on the array type, so that static arrays can get special treatment?

gdalle:
Actually I'm gonna put this on hold until I get a better sense of what it means to choose a batch size. I'll try to write something up for the DI docs and then go from there. Please don't merge in the meantime.

PR #1546, opened 2024-06-18
Update LLVM.jl

wsmoses:


giordano:
I'm probably missing some context, but there's no LLVM.jl v8 yet: https://github.com/JuliaRegistries/General/blob/cdbfd773a76c85cbfaf1d32bb11468ad6ce705c0/L/LLVM/Versions.toml?

Issue #1562, opened 2024-06-24
Enzyme v0.12.17 fails the DifferentiationInterface test suite


gdalle:
The new version v0.12.17 fails DI tests which used to pass before, as I realized in a completely unrelated PR (https://github.com/gdalle/DifferentiationInterface.jl/pull/327).

Something is wrong with the part of our tests related to the pullback of a scalar-to-vector function:

https://github.com/gdalle/DifferentiationInterface.jl/blob/4390675e46150273c60a6f493d8b48cf04e57e5b/DifferentiationInterfaceTest/src/scenarios/default.jl#L41-L49

It now returns `0` instead of the correct result.

MWE (needs the `main` branch of the DI repo, not the latest release):

```julia
using Pkg

Pkg.add(
    url="https://github.com/gdalle/DifferentiationInterface.jl",
    subdir="DifferentiationInterface"
)
    
Pkg.add(
    url="https://github.com/gdalle/DifferentiationInterface.jl",
    subdir="DifferentiationInterfaceTest"
)

using DifferentiationInterface, DifferentiationInterfaceTest, Enzyme

test_differentiation(AutoEnzyme(); excluded=[:derivative, :gradient, :hessian, :hvp, :jacobian, :pushforward, :second_derivative], logging=true)
```

wsmoses:
This was the result of a bad jll uptake yesterday and has been fixed on main (and being released today)

Issue #1401, opened 2024-04-22
Reverse mode is type-unstable in v0.12


gdalle:
I noticed the following change in https://github.com/gdalle/DifferentiationInterface.jl/pull/196:

Enzyme v0.11: type-stable

```julia
(@v1.10) pkg> activate --temp
  Activating new project at `/tmp/jl_xGmbHM`

(jl_xGmbHM) pkg> add JET, Enzyme@0.11
   Resolving package versions...
    Updating `/tmp/jl_xGmbHM/Project.toml`
⌃ [7da242da] + Enzyme v0.11.20
  [c3a54625] + JET v0.8.29
    Updating `/tmp/jl_xGmbHM/Manifest.toml`
    ...

julia> using JET, Enzyme

julia> @test_opt Enzyme.gradient(Enzyme.Reverse, sum, [1.0])
Test Passed
```

Enzyme v0.12: type-unstable

```julia
(@v1.10) pkg> activate --temp
  Activating new project at `/tmp/jl_JGHF4T`

(jl_JGHF4T) pkg> add JET, Enzyme@0.12
   Resolving package versions...
    Updating `/tmp/jl_JGHF4T/Project.toml`
  [7da242da] + Enzyme v0.12.0
  [c3a54625] + JET v0.8.29
    Updating `/tmp/jl_JGHF4T/Manifest.toml`
    ...

julia> using JET, Enzyme

julia> @test_opt Enzyme.gradient(Enzyme.Reverse, sum, [1.0])
JET-test failed at REPL[5]:1
  Expression: #= REPL[5]:1 =# JET.@test_opt Enzyme.gradient(Enzyme.Reverse, sum, [1.0])
  ═════ 49 possible errors found ═════
  ┌ gradient(::ReverseMode{false, FFIABI, false}, f::typeof(sum), x::Vector{Float64}) @ Enzyme /home/gdalle/.julia/packages/Enzyme/MIIMf/src/Enzyme.jl:932
  │┌ active_reg_inner(::Type{Vector{Float64}}, seen::Tuple{}, world::Nothing, ::Val{true}) @ Enzyme.Compiler /home/gdalle/.julia/packages/Enzyme/MIIMf/src/compiler.jl:367
[...]

Issue #1391, opened 2024-04-17
Reverse mode `jacobian` is wrong for matrix input


gdalle:
This seems related to the wrong jacobian shape of #1332 but it turns out even a reshape is not enough:

```julia
julia> using Enzyme

julia> x = float.(reshape(1:6, 2, 3))
2×3 Matrix{Float64}:
 1.0  3.0  5.0
 2.0  4.0  6.0

julia> f(x) = fill(sum(abs2, x), 6);

julia> jac_forward = Enzyme.jacobian(Enzyme.Forward, f, x)
6×6 Matrix{Float64}:
 2.0  4.0  6.0  8.0  10.0  12.0
 2.0  4.0  6.0  8.0  10.0  12.0
 2.0  4.0  6.0  8.0  10.0  12.0
 2.0  4.0  6.0  8.0  10.0  12.0
 2.0  4.0  6.0  8.0  10.0  12.0
 2.0  4.0  6.0  8.0  10.0  12.0

julia> jac_reverse = reshape(Enzyme.jacobian(Enzyme.Reverse, f, x, Val(6), Val(1)), 6, 6)
6×6 Matrix{Float64}:
  2.0   2.0   2.0   4.0   4.0   4.0
  6.0   6.0   6.0   8.0   8.0   8.0
 10.0  10.0  10.0  12.0  12.0  12.0
  2.0   2.0   2.0   4.0   4.0   4.0
  6.0   6.0   6.0   8.0   8.0   8.0
 10.0  10.0  10.0  12.0  12.0  12.0
```

gdalle:
This is now solved, although the shape of the forward Jacobian has changed:

```julia
julia> using Enzyme

julia> x = float.(reshape(1:6, 2, 3))
2×3 Matrix{Float64}:
 1.0  3.0  5.0
 2.0  4.0  6.0

julia> f(x) = fill(sum(abs2, x), 6);

julia> jac_forward = Enzyme.jacobian(Enzyme.Forward, f, x)

6×2×3 Array{Float64, 3}:
[:, :, 1] =
 2.0  4.0
 2.0  4.0
 2.0  4.0
 2.0  4.0
 2.0  4.0
 2.0  4.0

[:, :, 2] =
 6.0  8.0
 6.0  8.0
 6.0  8.0
 6.0  8.0
 6.0  8.0
 6.0  8.0

[:, :, 3] =
 10.0  12.0
 10.0  12.0
 10.0  12.0
[...]

PR #1563, opened 2024-06-24
Add DifferentiationInterface test suite

gdalle:
Prompted by #1562, this PR adds some tests from DifferentiationInterfaceTest to the Enzyme test suite.
Our test suite has caught [numerous bugs in the past](https://github.com/gdalle/DifferentiationInterface.jl/issues/99), so it would be a great asset to check that new Enzyme releases don't jeopardize existing behavior.

### DIFF
--- a/test/Project.toml
+++ b/test/Project.toml
@@ -2,6 +2,8 @@
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 ChainRules = "082447d4-558c-5d27-93f4-14fc19e9eca2"
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
+DifferentiationInterface = "a0c0ee7d-e4b9-4e03-894e-1c5f64a51d63"
+DifferentiationInterfaceTest = "a82114a7-5aa3-49a8-9643-716bb13727a3"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 EnzymeTestUtils = "12d8515a-0907-448a-8884-5fe00fdf1c5a"
 Enzyme_jll = "7cc45869-7501-5eee-bdea-0790c847d4ef"
--- a/test/differentiationinterface.jl
+++ b/test/differentiationinterface.jl
@@ -1,0 +1,6 @@
+using DifferentiationInterface
+using DifferentiationInterfaceTest
+using Enzyme: Enzyme
+
+test_differentiation([AutoEnzyme(; mode=Enzyme.Forward), AutoEnzyme(; mode=Enzyme.Reverse)];
+                     second_order=false, logging=false)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -100,6 +100,10 @@ end
     include("blas.jl")
 end
 
+@static if VERSION >= v"1.10"
+    include("differentiationinterface.jl")
+end
+
 f0(x) = 1.0 + x
 function vrec(start, x)
     if start > length(x)
@@ -3411,4 +3415,3 @@ end
 
 
 end
-


wsmoses:
This was the result of a bad jll uptake yesterday, already caught by our existing CI, and has been fixed on main (and being released today).

I'm not opposed to this in principle but the tests need to provide actionable feedback of Enzyme-specific MWE's.

In any case since the existing suite caught this already, I'm going to assume the need for is moot and close. Feel free to re-open if you'd like.

gdalle:
I can't reopen a PR you closed.

> In any case since the existing suite caught this already, I'm going to assume the need for is moot and close.

This specific bug was caught by your CI, but others may not be, and others have not been int he past. For instance, by using our type stability checks, you would have been able to catch https://github.com/EnzymeAD/Enzyme.jl/issues/1401 right away.

> I'm not opposed to this in principle but the tests need to provide actionable feedback of Enzyme-specific MWE's.

What do you mean by this? Test failures in DifferentiationInterfaceTest are already fairly verbose and precise, take a look at the stack trace here for instance: https://github.com/gdalle/DifferentiationInterface.jl/actions/runs/9644116325/job/26595528876

wsmoses:
We do generally add tests for any failures caught (so as an example the test you showed earlier led to us adding some inferred tests that should check the same).

I guess the source of my hesitation here is that while adding these as integration tests is useful as a signal that everything is fine or not, the feedback isn't actionable to an enzyme dev.

For example looking at the posted log:
```
Cotangent value: Test Failed at /home/runner/work/DifferentiationInterface.jl/DifferentiationInterface.jl/DifferentiationInterfaceTest/src/tests/correctness.jl:197
  Expression: dx1 ≈ scen.res1
   Evaluated: 0.0 ≈ 9.020996451098895

Stacktrace:
 [1] macro expansion
   @ /opt/hostedtoolcache/julia/1.10.4/x64/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
 [2] macro expansion
   @ ~/work/DifferentiationInterface.jl/DifferentiationInterface.jl/DifferentiationInterfaceTest/src/tests/correctness.jl:197 [inlined]
 [3] macro expansion
   @ /opt/hostedtoolcache/julia/1.10.4/x64/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [4] macro expansion
   @ ~/work/DifferentiationInterface.jl/DifferentiationInterface.jl/DifferentiationInterfaceTest/src/tests/correctness.jl:197 [inlined]
 [5] macro expansion
   @ /opt/hostedtoolcache/julia/1.10.4/x64/share/julia/stdlib/v1.10/Test/src/Test.jl:1669 [inlined]
 [6] test_correctness(ba::AutoEnzyme{Nothing}, scen::DifferentiationInterfaceTest.Scenario{:pullback, 1, :outofplace, typeof(DifferentiationInterfaceTest.num_to_arr_vector), Float64, Vector{Float64}, Vector{Float64}, Float64, Nothing}; isapprox::typeof(isapprox), atol::Int64, rtol::Float64)
   @ DifferentiationInterfaceTest ~/work/DifferentiationInterface.jl/DifferentiationInterface.jl/DifferentiationInterfaceTest/src/tests/correctness.jl:180
   ```
   
I have no idea what the isolated function is, or the corresponding Enzyme.autodiff call which would correspond to it (required to successfully debug). If you would be able to print out a corresponding MWE with corresponding Enzyme.autodiff call which triggers that would be immensely useful.

Also fwiw, adding this test suite here would not catch the bug before added, since, like mentioned this happened downstream in the jll.

**Comment by gdalle** [comment]
> I have no idea what the isolated function is, or the corresponding Enzyme.autodiff call which would correspond to it (required to successfully debug). If you would be able to print out a corresponding MWE with corresponding Enzyme.autodiff call which triggers that would be immensely useful.

Yeah, the stack trace is not very useful here, what I meant was the test summary at the end.
The summary below tells you that the failure happened during tests of `AutoEnzyme()` (which is either forward or reverse mode), for the `pullback` operator, in a `Scenario` involving the function `num_to_arr_vector`, where the operator was applied out-of-place.

image

**Comment by gdalle** [comment]
> If you would be able to print out a corresponding MWE with corresponding Enzyme.autodiff call which triggers that would be immensely useful.

The issue is that it would require storing the result of every test somehow, instead of just relying on the nested `@testset` macros to do their job. We've had a lot of discussions with @adrhill on how to do that, and decided it was best to leave it alone, see https://github.com/gdalle/DifferentiationInterface.jl/issues/109. Otherwise we'd have to define our own custom test set types, and everything becomes infinitely more complicated.

**Comment by wsmoses** [comment]
Yeah I mean like I said before that's nice as an integration test, but that level of summary isn't sufficient to be able to debug -- which really requires an MWE of the fn and the autodiff.

My hesitation here is that if someone has a PR which causes that to fail, they'd also have to look through a separate repo and have to try to disect what the failing code is, and how to reproduce it (after of course confirming whether that is indeed a bug or intended).

Given that the bar to contribution is already relatively high, I'm hesitant to force devs to have to bear that burden (hence the ask for it to spit out an isolatable MWE). Otherwise it'll just get ignored, at which point there isn't much value in a check (like the present formatter) which is always assumed failing and never checked.

**Comment by gdalle** [comment]
Would it be enough if the test displayed a custom error message that gives you an MWE in DifferentiationInterface syntax? Something like:

---
_To reproduce this behavior, run the following Julia code and inspect the result:_
```julia
import Enzyme
using DifferentiationInterface, DifferentiationInterfaceTest

backend = AutoEnzyme()
f = DifferentiationInterfaceTest.num_to_arr_vector
x = 1.0
dy = float.(1:6)
pullback(f, backend, x, dy)
```

wsmoses:
I think here I'd really like to see the Enzyme autodiff / syntax [just to rule out false positives if from DI]

wsmoses:
If you can do that [and very minor make this a separate github CI target rather than default part of test suite], I'd be happy to merge!

gdalle:
How would this be possible for operators that are not defined in Enzyme? For instance if DI.hessian fails?
Even for operators that are defined directly using Enzyme, it would be really hard to pull off without some heavy metaprogramming skills which I lack. Maybe point to the URL of the corresponding source in the DI Enzyme extension?

gdalle:
Even for existing operators, there are some I cannot use because of (apparent?) buggy behavior

https://github.com/EnzymeAD/Enzyme.jl/issues/1391

wsmoses:
Yeah I think the "translate into an Enzyme.autodiff call" would only really be critical for a gradient. More complex ops, like hessian, etc I think would be completely fine to leave as DI.hessian/etc

wsmoses:
including jacobian, btw!

MilesCranmer:
+1 for more integration tests. Even if such tests don't provide Enzyme devs with immediate actionable feedback into what went wrong, they act as a canary in a coal mine – something somewhere broke, and should be investigated before a new release is made.

As a compromise, I think you could pin the version of external packages in integration tests. That way, any failure is likely to indicate an internal regression, rather than simply indicating a new feature in a third-party library.

gdalle:
Closing for now because DITest is undergoing some major changes, but I'll be back

PR #1575, opened 2024-06-26
First stab at nongenabi

wsmoses:
@vchuravy what are your thoughts on a design like this for supporting non-gen abi

### DIFF
--- a/src/Enzyme.jl
+++ b/src/Enzyme.jl
@@ -247,11 +247,11 @@ Enzyme.autodiff(ReverseWithPrimal, x->x*x, Active(3.0))
         eltype(A)    
     end
 
-    opt_mi = RABI <: NonGenABI ? Compiler.fspec(eltype(FA), tt′) : nothing
+    opt_mi = RABI <: NonGenABI ? Compiler.fspec(eltype(FA), tt′) : Val(world)
 
     if A <: Active
         if (!allocatedinline(rt) || rt isa Union) && rt != Union{}
-            forward, adjoint = Enzyme.Compiler.thunk(opt_mi, Val(world), FA, Duplicated{rt}, tt′, #=Split=# Val(API.DEM_ReverseModeGradient), Val(width), ModifiedBetween, #=ReturnPrimal=#Val(ReturnPrimal), #=ShadowInit=#Val(true), RABI)
+            forward, adjoint = Enzyme.Compiler.thunk(opt_mi, FA, Duplicated{rt}, tt′, #=Split=# Val(API.DEM_ReverseModeGradient), Val(width), ModifiedBetween, #=ReturnPrimal=#Val(ReturnPrimal), #=ShadowInit=#Val(true), RABI)
             res = forward(f, args...)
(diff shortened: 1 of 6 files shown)


src/compiler.jl
```diff
@@ -6710,9 +6710,8 @@ end
 @inline remove_innerty(::Type{<:MixedDuplicated}) = MixedDuplicated
 @inline remove_innerty(::Type{<:BatchMixedDuplicated}) = MixedDuplicated
 
-@inline @generated function thunk(::Val{World}, ::Type{FA}, ::Type{A}, tt::Type{TT},::Val{Mode}, ::Val{width}, ::Val{ModifiedBetween}, ::Val{ReturnPrimal}, ::Val{ShadowInit}, ::Type{ABI}) where {FA<:Annotation, A<:Annotation, TT, Mode, ModifiedBetween, width, ReturnPrimal, ShadowInit, World, ABI}   
+@inline function thunkbase(mi::Core.MethodInstance, ::Val{World}, ::Type{FA}, ::Type{A}, tt::Type{TT},::Val{Mode}, ::Val{width}, ::Val{ModifiedBetween}, ::Val{ReturnPrimal}, ::Val{ShadowInit}, ::Type{ABI}) where {FA<:Annotation, A<:Annotation, TT, Mode, ModifiedBetween, width, ReturnPrimal, ShadowInit, World, ABI}
```
vchuravy, line 6713: It would be important that this function doesn't take world as an input.

wsmoses: Okay untested but done, regardless does this design make sense?

src/Enzyme.jl
```diff
@@ -412,7 +414,9 @@ f(x) = x*x
     tt    = Tuple{map(T->eltype(Core.Typeof(T)), args)...}
     world = codegen_world_age(Core.Typeof(f.val), tt)
 
-    thunk = Enzyme.Compiler.thunk(Val(world), FA, RT, tt′, #=Mode=# Val(API.DEM_ForwardMode), Val(width),
+    opt_mi = RABI <: NonGenABI ? Compiler.fspec(eltype(FA), tt′) : Val(world)
```
vchuravy, line 417: This observation of world is also an issue.

vchuravy:
In order for code to relocatable you may not observe world as a compile time constant. You may look at the tls_world_age, but you may not cache that or put it into a Val.

src/Enzyme.jl
wsmoses: But it shouldn't occur for the nongen case, no ?

Or do I need to literally separate into two functionsfor the different ABI's?

vchuravy: I would write it as:

```
    opt_mi = if RABI <: NonGenABI
        Compiler.fspec(eltype(FA), tt′)
    else
        Val(codegen_world_age(Core.Typeof(f.val), tt))
    end
```

To ensure that world is not accidentally captured. And then look at all uses of `codegen_world_age` and do the same.

Issue #1580, opened 2024-06-27
LLVM version issue with latest release?


ChrisRackauckas:
```julia
Continuous and discrete costs: Error During Test at /home/runner/.julia/packages/SafeTestsets/raUNr/src/SafeTestsets.jl:30
  Got exception outside of a @test
  LoadError: MethodError: no method matching cpu_features!(::LLVM.ModulePassManager)
  
  Closest candidates are:
    cpu_features!(::LLVM.Module)
     @ GPUCompiler ~/.julia/packages/GPUCompiler/05oYT/src/optim.jl:606
  
  Stacktrace:
    [1] (::Enzyme.Compiler.var"#28472#28473"{LLVM.Module, LLVM.TargetMachine})(pm::LLVM.ModulePassManager)
      @ Enzyme.Compiler ~/.julia/packages/Enzyme/qd8AI/src/compiler/optimize.jl:1941
    [2] LLVM.ModulePassManager(::Enzyme.Compiler.var"#28472#28473"{LLVM.Module, LLVM.TargetMachine}; kwargs::@Kwargs{})
      @ LLVM ~/.julia/packages/LLVM/5DlHM/src/passmanager.jl:33
    [3] LLVM.ModulePassManager(::Function)
      @ LLVM ~/.julia/packages/LLVM/5DlHM/src/passmanager.jl:30
    [4] optimize!(mod::LLVM.Module, tm::LLVM.TargetMachine)
      @ Enzyme.Compiler ~/.julia/packages/Enzyme/qd8AI/src/compiler/optimize.jl:1930
    [5] codegen(output::Symbol, job::GPUCompiler.CompilerJob{Enzyme.Compiler.EnzymeTarget, Enzyme.Compiler.EnzymeCompilerParams}; libraries::Bool, deferred_codegen::Bool, optimize::Bool, toplevel::Bool, strip::Bool, validate::Bool, only_entry::Bool, parent_job::Nothing)
      @ Enzyme.Compiler ~/.julia/packages/Enzyme/qd8AI/src/compiler.jl:5773
    [6] _thunk(job::GPUCompiler.CompilerJob{Enzyme.Compiler.EnzymeTarget, Enzyme.Compiler.EnzymeCompilerParams}, postopt::Bool)
      @ Enzyme.Compiler ~/.julia/packages/Enzyme/qd8AI/src/compiler.jl:6658
    [7] cached_compilation
[...]

PR #1581, opened 2024-06-27
Fix LLVM feature check

wsmoses:


vchuravy:
Fixes #1580

src/compiler/optimize.jl
```diff
@@ -1937,8 +1937,12 @@ function optimize!(mod::LLVM.Module, tm)
         basic_alias_analysis!(pm)
         cfgsimplification!(pm)
         dce!(pm)
-@static if isdefined(GPUCompiler, :cpu_features!)
+@static if isdefined(LLVM, :cpu_features!)
+        LLVM.cpu_features!(pm)
+else
+@staticif isdefined(GPUCompiler, :cpu_features!)
```
avik-pal, line 1943:
```suggestion
@static if isdefined(GPUCompiler, :cpu_features!)
```

src/compiler/optimize.jl
```diff
@@ -1937,8 +1937,12 @@ function optimize!(mod::LLVM.Module, tm)
         basic_alias_analysis!(pm)
         cfgsimplification!(pm)
         dce!(pm)
+@static if isdefined(LLVM, :cpu_features!)
+        LLVM.cpu_features!(pm)
+else
 @static if isdefined(GPUCompiler, :cpu_features!)
```
vchuravy, line 1943:
```suggestion
elseif isdefined(GPUCompiler, :cpu_features!)
```

?

wsmoses: does static still apply to the else tho?

that was my worry

vchuravy: Yeah it should

Issue #1205, opened 2023-12-18
Hessian vector product guidance


cmhamel:
What is the proper approach to get performant hessian vector products for a scalar function with a vector valued input?

wsmoses:
With the new hvp utilities and docs, closing

PR #1583, opened 2024-06-28
Add hvp helpers

wsmoses:


wsmoses:
Fixes https://github.com/EnzymeAD/Enzyme.jl/issues/1205

wsmoses:
@vchuravy after a few attempts I can't seem to get the doctest filter syntax right if you know the right incantation

Project.toml
```diff
@@ -20,7 +20,7 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 CEnum = "0.4, 0.5"
 ChainRulesCore = "1"
 EnzymeCore = "0.7.5"
-Enzyme_jll = "0.0.127"
+Enzyme_jll = "0.0.126"
```
giordano, line 23: Why is this necessary?

wsmoses: oh CI wasn't running at all without the other package registered, so i did this temporarily to see if it would run

giordano: Maybe there was a delay in the registry?  [v0.0.127](https://github.com/JuliaRegistries/General/pull/109889) was registered over 24 hours ago.

PR #1632, opened 2024-07-12
Simplify no derivative message

wsmoses:


wsmoses:
cc @mhauru @yebai feel free to follow up this PR with changes to the text of the error message to be more informative.

Nevertheless this should no longer print the surrounding function scope in these cases.

PR #1655, opened 2024-07-21
Add internal forward-mode rules for ranges

ChrisRackauckas:
This is part 1 one solving https://github.com/EnzymeAD/Enzyme.jl/issues/274. It does the forward mode rules as those are simpler. A separate PR will do the WIP reverse mode rules as that seems to be a bit more complex.


src/internal_rules.jl
```diff
@@ -815,3 +815,17 @@ function EnzymeRules.forward(func::Const{typeof(ldiv!)},
...
     end
 end
+
+# Ranges
+# Float64 ranges in Julia use bitwise `&` with higher precision
+# to correct for numerical error, thus we put rules over the
+# operations as this is not directly differentiable
+
+getval(x) = hasproperty(x, :val) ? x.val : x
+function EnzymeRules.forward(func::Const{Colon}, ::Type{<:Duplicated}, start::Union{Const, Active}, step::Union{Const, Active}, stop::Union{Const, Active})
```
wsmoses, line 825: Can you remove the union{const,active} restriction, make Type{<:Duplicated} more general, and in doing so add batchduplicated/duplicatednoneed?

See https://github.com/EnzymeAD/Enzyme.jl/blob/d6a3ddcd4122d45b3b245b436f35b5d81d84fe1c/src/internal_rules.jl#L582 / https://github.com/EnzymeAD/Enzyme.jl/blob/d6a3ddcd4122d45b3b245b436f35b5d81d84fe1c/src/internal_rules.jl#L600 for example

ChrisRackauckas:
How's that?

src/internal_rules.jl
```diff
@@ -815,3 +815,25 @@ function EnzymeRules.forward(func::Const{typeof(ldiv!)},
...
     end
 end
+
+# Ranges
+# Float64 ranges in Julia use bitwise `&` with higher precision
+# to correct for numerical error, thus we put rules over the
+# operations as this is not directly differentiable
+
+getval(x) = hasproperty(x, :val) ? x.val : x
+function EnzymeRules.forward(func::Const{Colon}, RT::Type{<:Union{Const, DuplicatedNoNeed, Duplicated}}, start, step, stop)
```
wsmoses, line 825: this still needs to add batchduplicated/batchduplicatednoneed/etc , but otherwise almost there!

src/internal_rules.jl
```diff
@@ -815,3 +815,25 @@ function EnzymeRules.forward(func::Const{typeof(ldiv!)},
...
+    ret = func.val(getval.((start, step, stop))...)
```
wsmoses, line 826: Since it's three things can you just do .val for all of them [there isn't a question that they would have the property all Enzyme.Annotation values have .val\

ChrisRackauckas, line 826:
```suggestion
    ret = func.val(start.val, step.val, stop.val)
```

ChrisRackauckas:
How's this? Is there anything like this `batchsize` utility?

wsmoses:
You should be able to just call width(..). Also per above you should probably do something like

ntuple(Val(width(RT))) do i 
  Base.@_inline_meta
  Thing
End

There are several examples of this in this file

PR #1656, opened 2024-07-21
WIP: Add internal reverse-mode rules for ranges

ChrisRackauckas:
This is the second PR to fix https://github.com/EnzymeAD/Enzyme.jl/issues/274. It's separated as I think the forward mode one can just be merged no problem, and this one may take a little bit more time.

The crux of why this one is hard is because of how Julia deals with malformed ranges.

```
Basically dret.val = 182.0:156.0:26.0, the 26.0 is not the true value. Same as

julia> 10:1:1
10:1:9
```

Because of that behavior, the reverse `dret` does not actually have the information as to what its final point is, and its length is "incorrect" as it's changed by the constructor. In order to "fix" the reverse, we'd want to swap the `step` to negative and then use the same start/stop, but that information is already lost so it cannot be fixed within the rule. You can see the commented out code that would do the fixing if the information is there, and without that we cannot get a correctly sized reversed range for the rule.

But it's a bit puzzling to figure out how to remove that behavior. In Base Julia it seems to be done in the `function (:)(start::T, step::T, stop::T) where T<:IEEEFloat`, and as I showed in the issue, I can overload that function and the behavior goes away, but Enzyme's constructed range still has that truncation behavior, which means I missed spot or something.

### DIFF
--- a/src/internal_rules.jl
+++ b/src/internal_rules.jl
@@ -875,7 +875,7 @@ function EnzymeRules.augmented_primal(config, func::Const{Colon}, ::Type{<:Activ
 end
 
 function EnzymeRules.reverse(config, func::Const{Colon}, dret, tape::Nothing,
-                 start::Annotation{<:AbstractFloat}, step::Annotation{<:AbstractFloat}, stop::Annotation{<:AbstractFloat})
+                 start::Annotation{T1}, step::Annotation{T2}, stop::Annotation{T3}) where {T1<:AbstractFloat, T2<:AbstractFloat, T3<:AbstractFloat}
 
     primal = func.val(start.val, step.val, stop.val)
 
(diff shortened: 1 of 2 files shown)


src/internal_rules.jl
```diff
@@ -815,3 +815,42 @@ function EnzymeRules.forward(func::Const{typeof(ldiv!)},
...
+getval(x) = hasproperty(x, :val) ? x.val : x
+function EnzymeRules.forward(func::Const{Colon}, ::Type{<:Duplicated}, start::Union{Const, Active}, step::Union{Const, Active}, stop::Union{Const, Active})
+    ret = func.val(getval.((start, step, stop))...)
+    dstart = start isa Const ? zero(eltype(ret)) : one(eltype(ret))
+    dstep = step isa Const ? zero(eltype(ret)) : one(eltype(ret))
+
+    return Duplicated(ret, range(dstart, step=dstep, length=length(ret)))
+end
+
+function EnzymeRules.augmented_primal(config::EnzymeRules.ConfigWidth{1}, func::Const{Colon}, ::Type{<:Active},
```
wsmoses, line 833: This doesn't need to be limited to configwidth 1 here, it would equally work for width n

src/internal_rules.jl
```diff
@@ -815,3 +815,42 @@ function EnzymeRules.forward(func::Const{typeof(ldiv!)},
...
+        primal = nothing
+    end
+    return EnzymeRules.AugmentedReturn(primal, nothing, nothing)
+end
+
+function EnzymeRules.reverse(config::EnzymeRules.ConfigWidth{1}, func::Const{Colon}, dret, tape::Nothing,
+                 start, step, stop)
+
+    #fixedreverse = if _dret.start > _dret.stop && _dret.step > 0
+    #     _dret.stop:_dret.step:_dret.start
```
wsmoses, line 848: can this just be an expr new constructor [and bypass checks/rewriting]?

ChrisRackauckas: I don't know what you mean by an expr new constructor. Is that some Enzyme thing to bypass the inner constructor?

ChrisRackauckas: We can't do it in the rule though because the information about what the true stop is is already lost by the time it hits the rule. The rule has already given us a `dret` which has no valid stop from reversing the step.

src/internal_rules.jl
ChrisRackauckas: Yes, just trying to get things working one step at a time.

src/internal_rules.jl
wsmoses: something like this: https://github.com/EnzymeAD/Enzyme.jl/blob/8a850c77be67604e9b9429be19f2876db29899b3/src/rules/typeunstablerules.jl#L54

PR #1663, opened 2024-07-22
Pushforward dvals in the ranges forward rule

ChrisRackauckas:
I just realized before it merged that we're actually missing a part of the pushforward here. Even though the solution to the derivative is just the range itself, we forgot to multiply it by the dval to pushforward the derivative. It implicitly had it as one, calculating the derivative, instead of the full pushforward. The test should be updated to catch this.

### DIFF
--- a/src/internal_rules.jl
+++ b/src/internal_rules.jl
@@ -825,9 +825,9 @@ function EnzymeRules.forward(func::Const{Colon}, RT::Type{<:Union{Const, Duplica
     dstart = if start isa Const 
         zero(eltype(ret)) 
     elseif start isa Duplicated || start isa DuplicatedNoNeed
-        one(eltype(ret))
+        start.dval * one(eltype(ret))
     elseif start isa BatchDuplicated || start isa BatchDuplicatedNoNeed
-        ntuple(x->one(eltype(ret)), Val(width(RT)))
+        ntuple(i->start.dval[i] * one(eltype(ret)), Val(width(RT)))
     else
         error("Annotation type $(typeof(start)) not supported for range start. Please open an issue")
     end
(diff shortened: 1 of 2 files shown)


src/internal_rules.jl
```diff
@@ -825,19 +825,19 @@ function EnzymeRules.forward(func::Const{Colon}, RT::Type{<:Union{Const, Duplica
     dstart = if start isa Const 
         zero(eltype(ret)) 
     elseif start isa Duplicated || start isa DuplicatedNoNeed
-        one(eltype(ret))
+        start.dval * one(eltype(ret))
```
wsmoses, line 828: why multiply against one here then?

wsmoses (commented):
LGTM but yeah please update the test

src/internal_rules.jl
ChrisRackauckas: type promotion

wsmoses: why not just do a type cast then?

test/internal_rules.jl
```diff
@@ -620,25 +620,29 @@ end
...
     function f4(x)
+	x = 25.0x
         ts = Array(0.0:.25:x)
         sum(ts)
     end
-    @test Enzyme.autodiff(Forward, f1,  Duplicated(0.25, 1.0)) == (78,)
-    @test Enzyme.autodiff(Forward, f2,  Duplicated(0.25, 1.0)) == (1.0,)
-    @test Enzyme.autodiff(Forward, f3,  Duplicated(0.25, 1.0)) == (12,)
-    @test Enzyme.autodiff(Forward, f4,  Duplicated(3.0, 1.0)) == (0,)
+    @test Enzyme.autodiff(Forward, f1,  Duplicated(0.1, 1.0)) == (78,)
```
wsmoses, line 642: While at it, can we also test with batching?

Issue #1674, opened 2024-07-24
Missing 2-argument method for `bitcast!`


MilesCranmer:
I think this line:

https://github.com/EnzymeAD/Enzyme.jl/blob/5fc23099cdac3dbdfa9d413ee3f17b34759bb6f2/src/compiler.jl#L6622

calls a method that does not exist? We are running into an error for it here: https://github.com/SymbolicML/DynamicExpressions.jl/pull/91#issuecomment-2248464734

```julia
Test enzyme derivatives: Error During Test at /home/runner/.julia/packages/SafeTestsets/raUNr/src/SafeTestsets.jl:30
  Got exception outside of a @test
  LoadError: MethodError: no method matching bitcast!(::LLVM.IRBuilder, ::LLVM.PointerType)
  
  Closest candidates are:
    bitcast!(::LLVM.IRBuilder, ::LLVM.Value, ::LLVM.LLVMType, ::String)
     @ LLVM ~/.julia/packages/LLVM/5DlHM/src/irbuilder.jl:343
    bitcast!(::LLVM.IRBuilder, ::LLVM.Value, ::LLVM.LLVMType)
     @ LLVM ~/.julia/packages/LLVM/5DlHM/src/irbuilder.jl:343
  
  Stacktrace:
    [1] #28366
      @ ~/.julia/packages/Enzyme/aEyGD/src/compiler.jl:6622 [inlined]
    [2] JuliaContext(f::Enzyme.Compiler.var"#28366#28371"{Bool, DataType, DataType, DataType, Int64, DataType, Int64, Bool, DataType, DataType}; kwargs::@Kwargs{})
      @ GPUCompiler ~/.julia/packages/GPUCompiler/Y4hSX/src/driver.jl:52
    [3] JuliaContext(f::Function)
      @ GPUCompiler ~/.julia/packages/GPUCompiler/Y4hSX/src/driver.jl:42
    [4] #s2028#28365
      @ ~/.julia/packages/Enzyme/aEyGD/src/compiler.jl:6276 [inlined]
    [5] var"#s2028#28365"(RawCall::Any, PT::Any, FA::Any, T::Any, RT::Any, TapeType::Any, N::Any, CC::Any, width::Any, returnPrimal::Any, ::Any, ::Type, fptr::Any, ::Type, ::Type, ::Type, tt::Any, rt::Any, fn::Any, ::Any, args::Any)
      @ Enzyme.Compiler ./none:0
[...]

Issue #1662, opened 2024-07-22
Hitting new segfaults in DynamicExpressions.jl


MilesCranmer:
I have a integration test of Enzyme in DynamicExpressions.jl that has been working fine for the past few months but recently started hitting segfaults (without changes to DynamicExpressions itself as far as I can tell).

Here's the test:

```julia
using Pkg
pkg"add Test Preferences Enzyme DynamicExpressions"
using Test
using Preferences
using Enzyme
set_preferences!("DynamicExpressions", "instability_check" => "error")
# ^ Seems to be necessary to trigger error

using DynamicExpressions

operators = OperatorEnum(; binary_operators=(+, -, *, /), unary_operators=(cos, sin))
const eval_options = (turbo=Val(false),)
x1, x2 = (i -> Node(Float64; feature=i)).(1:2)

tree = Node(1, x1, Node(1, x2))

X = randn(3, 100);
dX = zero(X)

function f(tree, X, operators, output)
    output[] = sum(eval_tree_array(tree, X, operators; eval_options...)[1])
    return nothing
end

output = [0.0]
doutput = [1.0]

autodiff(
    Reverse,
    f,
    Const(tree),
    Duplicated(X, dX),
    Const(operators),
    Duplicated(output, doutput),
)

true_dX = cat(ones(100), -sin.(X[2, :]), zeros(100); dims=2)'
@test true_dX ≈ dX
```

I noticed today that I started getting segfaults:

```
[33909] signal (11.2): Segmentation fault: 11
in expression starting at REPL[18]:1
typekeyvalue_hash at /Users/julia/.julia/scratchspaces/a66863c6-20e8-4ff4-8a62-49f30b1f605e/agent-cache/default-honeycrisp-XG3Q6T6R70.0/build/default-honeycrisp-XG3Q6T6R70-0/julialang/julia-release-1-dot-10/src/jltypes.c:1631 [inlined]
lookup_typevalue at /Users/julia/.julia/scratchspaces/a66863c6-20e8-4ff4-8a62-49f30b1f605e/agent-cache/default-honeycrisp-XG3Q6T6R70.0/build/default-honeycrisp-XG3Q6T6R70-0/julialang/julia-release-1-dot-10/src/jltypes.c:1059
[...]

Issue #1018, opened 2023-08-19
Hanging and task switch errors when differentiating DynamicExpressions.jl


MilesCranmer:
In trying to get DynamicExpressions.jl to be compatible with Enzyme.jl, I've taken the approach of using `Base.Cartesian.@nif` over the space of operators – so that the compiler knows exactly what operators are being used at the forward pass. This fixes all of the remaining type stability issues.

However it seems like I am running into other issues with this change, in particular a StackOverflowError (or simply having the evaluation hang – which happens in this MWE).

Here's a MWE. I tried to reduce it more but I wasn't able to replicate the error, so unfortunately it is quite long. My intuition is that these nested `Base.Cartesian.@nif` (used for fusing operators) are too many branches for Enzyme to handle. What do you think?

```julia
using Enzyme

################################################################################
### OperatorEnum.jl
################################################################################
struct OperatorEnum{B,U}
    binops::B
    unaops::U
end
################################################################################

################################################################################
### Equation.jl
################################################################################
mutable struct Node{T}
    degree::UInt8  # 0 for constant/variable, 1 for cos/sin, 2 for +/* etc.
    constant::Bool  # false if variable
    val::Union{T,Nothing}  # If is a constant, this stores the actual value
    # ------------------- (possibly undefined below)
    feature::UInt16  # If is a variable (e.g., x in cos(x)), this stores the feature index.
    op::UInt8  # If operator, this is the index of the operator in operators.binops, or operators.unaops
    l::Node{T}  # Left child node. Only defined for degree=1 or degree=2.
    r::Node{T}  # Right child node. Only defined for degree=2. 
    Node(d::Integer, c::Bool, v::_T) where {_T} = new{_T}(UInt8(d), c, v)
    Node(::Type{_T}, d::Integer, c::Bool, v::_T) where {_T} = new{_T}(UInt8(d), c, v)
    Node(::Type{_T}, d::Integer, c::Bool, v::Nothing, f::Integer) where {_T} = new{_T}(UInt8(d), c, v, UInt16(f))
    Node(d::Integer, c::Bool, v::Nothing, f::Integer, o::Integer, l::Node{_T}) where {_T} = new{_T}(UInt8(d), c, v, UInt16(f), UInt8(o), l)
[...]

Issue #1009, opened 2023-08-17
Catching assertions / debugging mode?


MilesCranmer:
I'm happy to share that SymbolicRegression.jl integration is pretty much working, thanks to help from @wsmoses and @vchuravy across several issues! https://github.com/MilesCranmer/SymbolicRegression.jl/pull/254. It uses Enzyme.jl inside an Optim.jl loop to optimize constants in runtime-generated expressions.

However, there are still some issues I am seeing which show up for certain generated expressions. The issue is that I don't know which expressions cause this, as the errors I see crash Julia, so I am unable to create a MWE. For example, the one at the bottom of this issue. I was wondering if there was a debugging mode which lets me safely catch these issues (which seem to be assertion errors in the C code) so I can generate some debugging info?

<details><summary>Error message (no MWE available)</summary>

```
<analysis>
i64 -1: {[-1]:Anything}, intvals: {-1,}
{} addrspace(12)* addrspacecast ({}* inttoptr (i64 4311547912 to {}*) to {} addrspace(12)*): {[-1]:Anything}, intvals: {}
i8* getelementptr inbounds ([11 x i8], [11 x i8]* @_j_str1, i32 0, i32 0): {[-1]:Pointer, [-1,-1]:Anything}, intvals: {}
i8 112: {[-1]:Integer}, intvals: {112,}
i8 121: {[-1]:Integer}, intvals: {121,}
i8 115: {[-1]:Integer}, intvals: {115,}
i8 97: {[-1]:Integer}, intvals: {97,}
i8 114: {[-1]:Integer}, intvals: {114,}
i8 116: {[-1]:Integer}, intvals: {116,}
i8 101: {[-1]:Integer}, intvals: {101,}
  %1 = call noalias nonnull dereferenceable(3) dereferenceable_or_null(3) i8* @malloc(i64 3), !enzyme_fromstack !9: {[-1]:Pointer}, intvals: {}
  %16 = call {} addrspace(10)* @julia.typeof({} addrspace(10)* nonnull %15) #16, !dbg !60: {}, intvals: {}
  %17 = addrspacecast {} addrspace(10)* %16 to {} addrspace(11)*, !dbg !60: {}, intvals: {}
  %8 = load i8, i8 addrspace(11)* %7, align 8, !dbg !18, !tbaa !22, !alias.scope !26, !noalias !29: {[-1]:Integer}, intvals: {}
[...]

PR #1675, opened 2024-07-24
ci: add DynamicExpressions integration tests

MilesCranmer:
DynamicExpressions.jl seems to be good at flagging issues in Enzyme, such as the below, so this PR therefore adds a simple integration test for it so that these issues will get flagged immediately.

- #548
- #552
- #816
- #999
- #1009
- #1018
- #1156
- #1662
- #1674

The CI also pins DynamicExpressions to a specific version where it has been known to work, so any failures would indicate some kind of internal regression. Also, this is a separate CI run so won't be recorded in the main test suite.

This is also important for the stability of the PySR/SymbolicRegression.jl ecosystem which have a large number of non-technical users; I would want to prevent breakages for them when I eventually make Enzyme part of the default search code.

I restructured the CI so that packages can add their own integration tests fairly easily: 

```yaml
    matrix:
      version:
        - '1'
      os:
        - ubuntu-latest
      test:
        - DynamicExpressions
    steps:
      - uses: actions/checkout@v4
      - uses: julia-actions/setup-julia@v1
        with:
      - uses: julia-actions/cache@v1
      - uses: julia-actions/julia-buildpkg@v1
      - name: "Run tests"
        run: |
            julia --color=yes --project=. -e 'using Pkg; pkg"instantiate"'
            julia --color=yes --project=test/integration -e 'using Pkg; pkg"dev ."; pkg"instantiate"'
            julia --color=yes --project=test/integration --threads=auto --check-bounds=yes -e test/integration/${{ matrix.test }}.jl
        shell: bash
```

All they need to do is create a new `test/integration/{PackageName}.jl` file and add it to the list.

@gdalle maybe you could rebase #1563 on this so we can have them in the same format?

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -234,7 +234,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1'
+          - '1.10'
         os:
           - ubuntu-latest
         test:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -248,8 +248,8 @@ jobs:
       - uses: julia-actions/julia-buildpkg@v1
       - name: "Run tests"
         run: |
-            julia --color=yes --project=test/integration -e 'using Pkg; Pkg.instantiate(); Pkg.develop([PackageSpec(; path) for path in (".", "lib/EnzymeCore")]); Pkg.instantiate()'
-            julia --color=yes --project=test/integration --threads=auto --check-bounds=yes -e test/integration/${{ matrix.test }}.jl
+            julia --color=yes --project=test/integration -e 'using Pkg; Pkg.develop([PackageSpec(; path) for path in (".", "lib/EnzymeCore")]); Pkg.instantiate()'
+            julia --color=yes --project=test/integration --threads=auto --check-bounds=yes test/integration/${{ matrix.test }}.jl
         shell: bash
   docs:
     name: Documentation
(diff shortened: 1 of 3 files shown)


.github/workflows/CI.yml
```diff
@@ -225,6 +225,33 @@ jobs:
...
+  integration:
+    name: Integration Tests - ${{ matrix.test }}
+    runs-on: ${{ matrix.os }}
+    env:
+      JULIA_PKG_SERVER_REGISTRY_PREFERENCE: eager
+    strategy:
+      fail-fast: false
+      matrix:
+        version:
+          - '1'
```
wsmoses, line 237: Minor but can you change to 1.10 explicitly for now [will bump once we have 1.11 support]

wsmoses (approved):

.github/workflows/CI.yml
MilesCranmer: Will do

wsmoses:
SGTM!

.github/workflows/CI.yml
```diff
@@ -225,6 +225,33 @@ jobs:
...
+          - DynamicExpressions
+    steps:
+      - uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: ${{ matrix.version }}
+      - uses: julia-actions/cache@v1
+      - uses: julia-actions/julia-buildpkg@v1
+      - name: "Run tests"
+        run: |
```
wsmoses, line 250: oh you may also need to dev lib/EnzymeCore [see above]

MilesCranmer: Thanks, added

MilesCranmer:
Cool, seems to work!

(I think the #1674 is a different issue)

PR #1701, opened 2024-08-05
Add optional [but presently on] check if function contains written-to data

wsmoses:
This is conservative, since it will yell if data could be written -- meaning it might not be written, and also it may not be active data which is written.

cc @gdalle @ChrisRackauckas


gdalle:
Thank you for implementing this! Will take a look tomorrow and test it with ADTypes

PR #1716, opened 2024-08-09
fix: typetree_inner is not in Compiler module

avik-pal:
https://buildkite.com/julialang/luxlib-dot-jl/builds/1115#01913493-6aff-48e4-af64-3212902f9a52/301-1734


wsmoses (approved):

vchuravy:
fixes #1720

PR #1752, opened 2024-08-25
Handle const addr casts

wsmoses:


test/runtests.jl
```diff
@@ -780,6 +780,38 @@ end
     @test hess[1][2] ≈ 1.0
     @test hess[2][1] ≈ 1.0
     @test hess[2][2] ≈ 0.0
+
+    function f_ip(x, tmp)
+        tmp .= identity.(x)
```
danielwe, line 785: Not sure you'll trigger the issue with `identity`, I think the broadcasting machinery specializes on that one to produce the same code as `tmp .= x`. That's why I used the custom `g(x) = x`---equally trivial logic, but without the specialized code path.

danielwe: Confirming that `identity` fails to trigger the issue on my end. If you don't want to define the extra function you can also replace with something like `tmp .= x ./ 2`.

PR #1758, opened 2024-08-28
fix LLVM 8 depwarns

isaacsas:
SciML has moved to failing CI for depwarns, and Enzyme is a dependency of many SciML packages, so it would be nice to get a release that doesn't have these two depwarns.

I don't know the best practice for checking the version of a dependency (LLVM in this case), but would be happy to modify this PR if I should be doing something else.

Thanks!

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -15,7 +15,6 @@ ObjectFile = "d8793406-e978-5875-9003-1fc021f44a92"
 Preferences = "21216c6a-2e73-6563-6e65-726566657250"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
-TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
 
 [weakdeps]
 BFloat16s = "ab4f0b2a-ad5b-11e8-123f-65d77653426b"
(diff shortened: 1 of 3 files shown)


vchuravy:
`pkgversion` is 1.9 and above only.

isaacsas:
Do you know a pre-1.9 way to check the version of a package?

isaacsas:
I'll update to just go back to the old code pre-1.9. That hopefully shouldn't impact SciML anyways as we've moved to 1.10 for most packages.

isaacsas:
Ahh, actually it is 1.8. I'll search for how to check versions before pkgversion existed.

wsmoses:
Why not just make this unconditionally a Julia version check?

isaacsas:
LLVM v8 supports Julia 1.8 but `pkgversion` is Julia 1.9 and up. The deprecated functions are still present in earlier LLVM versions, which are listed as supported in Enzyme's Project.toml. So doesn't this mean one needs both Julia and LLVM version info to determine what to do here?

isaacsas:
I think this should hopefully fix it.

wsmoses:
yeah but the code in older llvm.jl versions just did a VERSION >= v"1.x" or whatever, so we might as well just do that here instead of depending on toml parsing

isaacsas:
OK, so what Julia version should I switch at? 1.8?

wsmoses:
Not sure, look at what the defn is of these fns from LLVM.jl before they dropped 1.6/1.9/etc?

isaacsas:
Thanks! If there is any chance to get a bug fix release with this it would help in unblocking some failing SciML library CI.

ChrisRackauckas:
Is this getting a release?

wsmoses:
done

PR #1760, opened 2024-08-29
non-breaking, easier to diff version of gradient-jacbian PR

ExpandingMan:
This is an alternate version of #1733 as requested there.  This version *should* be non-breaking in terms of API (though I'll still have to work through old julia versions if it doesn't work there).

**Please note!** Basically the whole idea of #1733 was to make the internals more consistent, which this PR must deliberately spoil in order to not break!  Therefore there are some pieces of this which are quite wonky but which make more sense when the `gradient` and `jacobian` return behavior is made consistent.

I have left as comments most of what would be needed to change to the behavior of #1733, the rest of the changes are the trivial re-arrangement of the files, so these two PR's together I believe fulfill the requests made there.

This does *NOT* include the new unit tests in #1733 of which some, but not all, would fail.


src/Enzyme.jl
```diff
@@ -1006,64 +1006,75 @@ end
...
-Besides arrays, for struct `x` it returns another instance of the same type,
-whose fields contain the components of the gradient.
-In the result, `grad.a` contains `∂f/∂x.a` for any differential `x.a`,
-while `grad.c == x.c` for other types.
+@inline tupleconcat(x) = x
+@inline tupleconcat(x, y) = (x..., y...)
+@inline tupleconcat(x, y, z...) = (x..., tupleconcat(y, z...)...)
 
-Examples:
+@inline function derivative(mode::ForwardMode, f, x; shadow=onehot(x))
```
wsmoses, line 1026: Since this is a new method, can we add tests?

src/Enzyme.jl
```diff
@@ -1006,64 +1006,75 @@ end
...
+    values(only(autodiff_deferred(mode, f, BatchDuplicatedNoNeed, BatchDuplicated(x, shadow))))
+end
 
-```jldoctest gradient
-f(x) = x[1]*x[2]
+@inline _chunkcheck(::Val{0}) = throw(ArgumentError("Cannot differentiate with a batch size of 0"))
+@inline _chunkcheck(::Val) = nothing
 
-grad = gradient(Reverse, f, [2.0, 3.0])
+@inline function derivative(mode::ForwardMode, f::F, x::X, ::Val{chunk};
```
wsmoses, line 1036: what is the semantics for derivative? I would've thought it be one input one output. If it needs multiple inputs though, why not just have this call gradient?

wsmoses: Ah I see, those now call this

src/Enzyme.jl
```diff
@@ -1108,228 +1118,39 @@ Like [`gradient!`](@ref), except it using deferred mode.
...
+@inline function jacobian(mode::ForwardMode, f, x; shadow=onehot(x))
+    #NOTE: these requires special handling to prevent breaking API
+    #gradient(mode, f, x; shadow)
+    df = derivative(mode, f, x; shadow)
+    jacobian_output_forward(df, df[1], x)
 end
 
-@inline function chunkedonehot(x::AbstractFloat, ::Val{chunk}) where chunk
-    return ((one(x),),)
+@inline function jacobian(mode::ForwardMode, f::F, x::X, ::Val{chunk};
```
wsmoses, line 1135: Can you restore all the docs/doctests?

src/Enzyme.jl
```diff
@@ -1108,228 +1118,39 @@ Like [`gradient!`](@ref), except it using deferred mode.
...
 
-@inline function chunkedonehot(x, ::Val{chunk}) where chunk
-    sz = length(x)
-    num = ((sz + chunk - 1) ÷ chunk)
-    ntuple(Val(num)) do i
-        Base.@_inline_meta
-        onehot(x, (i-1)*chunk+1, i == num ? sz : (i*chunk) )
-    end
+@inline function jacobian(mode::ForwardMode, f, x; shadow=onehot(x))
+    #NOTE: these requires special handling to prevent breaking API
```
wsmoses, line 1129: this note doesn't apply here, no?

src/Enzyme.jl
ExpandingMan: Yeah, the basic idea is that all `gradient` and `jacobian` do is re-arrange output of this, so it should be its own function.  The name `derivative` is perhaps too grandiose, I didn't have any more sophisticated idea of what it should be than the thing the returns of which are manipulated by those.

src/Enzyme.jl
ExpandingMan: Oops, removing them was not deliberate.

src/Enzyme.jl
ExpandingMan: It does, in the breaking version this is merely `gradient`, but in `main` currently there is a discrepancy between how `gradient` and `jacobian` handle outputs (owing to `gradient` sometimes returning tuples).  Therefore I had to make special accommodations to split the behavior here so it wouldn't break the API.

src/Enzyme.jl
ExpandingMan: done

src/Enzyme.jl
wsmoses: Jacobian should also conditionally return tuples, for example, if inputs or outputs are not arrays/scalars

ExpandingMan: This does that, as can be seen in the fallback where it returns the output of `derivative` `df` directly.  However I can only attest to the behavior being exactly the same in cases which are tested.  It was not entirely clear to me exactly what the semantics of the outputs are, one of the goals of the original PR was to help define them more clearly.

ext/EnzymeStaticArraysExt.jl
```diff
@@ -21,4 +21,11 @@ end
     end
 end
 
+
```
wsmoses, line 24: since this is completely commented out, let's remove it.

Also how it is api breaking. if the change is now that a staticarray is returned instead of an array, I'm fine with that change (the tuple one, people already use quite heavily so I'd want to keep that in a breaking change to make it easy for folks).

src/Enzyme.jl
wsmoses: Obviously likely undertested already, but since we're changing the implementation a bunch (and adding new fn's), would it be possible to add tests that confirm that all functions of the form input in (Float, Vector, Matrix, Struct) and output in (Float, Vector, Matrix, Struct)   work for as expected for (derivative, gradient, gradient!, jacobian) x ("", deferred)?

I realize this is kind of an obnoxious list, but also this is now the API surface (and presumably not sufficiently tested to start with).

If things don't break and work as intended, great!

If things break, we should fix it (or mark it as xfail, depending).

If things don't break, but are counter-intuitive, we should see if it would amount to a bugfix or breaking change for users, and fix it either immediately or follow up.

Nevertheless, for at least my sake reviewing it would be good to see this explicitly since the number of options is getting a bit unclear and large to think through all combinations (e.g. the old implementation explicitly had a switch for float vs vector vs matrix vs others that I'm now not as confident would be supported).

src/Enzyme.jl
```diff
@@ -1380,17 +1281,24 @@ of shape `size(output)` of values of the input type.
...
         tt′ = Tuple{BatchDuplicated{Core.Typeof(x), last_size}}
-        primal2, adjoint2 = Enzyme.Compiler.thunk(opt_mi, FA, BatchDuplicatedNoNeed{rt}, tt′, #=Split=# Val(API.DEM_ReverseModeGradient), #=width=#Val(last_size), ModifiedBetween, #=ReturnPrimal=#Val(false), #=ShadowInit=#Val(false), RABI, Val(ErrIfFuncWritten))
+        primal2, adjoint2 = Enzyme.Compiler.thunk(opt_mi, FA, BatchDuplicatedNoNeed{rt}, tt′,
+                                                  #=Split=# Val(API.DEM_ReverseModeGradient),
+                                                  #=width=#Val(last_size), ModifiedBetween,
+                                                  #=ReturnPrimal=#Val(false), #=ShadowInit=#Val(false),
+                                                  RABI, Val(ErrIfFuncWritten))
     end
-
+    #TODO: this is broken for static arrays
```
wsmoses, line 1301: yeah this function presently assumes that the returned array is mutable, which perhaps should be added to the docstring (and later remedied).

src/Enzyme.jl
```diff
@@ -1454,7 +1339,11 @@ end
     else
         Val(codegen_world_age(Core.Typeof(f), tt))
     end
-    primal, adjoint = Enzyme.Compiler.thunk(opt_mi, FA, DuplicatedNoNeed{rt}, tt′, #=Split=# Val(API.DEM_ReverseModeGradient), #=width=#Val(1), ModifiedBetween, #=ReturnPrimal=#Val(false), #=ShadowInit=#Val(false), RABI, Val(ErrIfFuncWritten))
+    primal, adjoint = Enzyme.Compiler.thunk(opt_mi, FA, DuplicatedNoNeed{rt}, tt′,
+                                            #=Split=# Val(API.DEM_ReverseModeGradient),
+                                            #=width=#Val(1), ModifiedBetween, #=ReturnPrimal=#Val(false),
+                                            #=ShadowInit=#Val(false), RABI, Val(ErrIfFuncWritten))
+    #TODO: this is broken for static arrays
```
wsmoses, line 1346: same comment as above re mutable assumption in this implementation.

src/Enzyme.jl
ExpandingMan: Yeah this was somewhat a note for myself as I definitely intend to make versions of this for static arrays at some point.  This is part of why I think it would be nice to have more autodiff methods, we could get it working there once and for all and not worry about separate `gradient` methods breaking it.

ext/EnzymeStaticArraysExt.jl
ExpandingMan: The issue here is entirely with the `gradient` methods that are currently returning `Tuple` instead of array.  If this is uncommented this will return `StaticArray` in cases where the `Array` equivalent would return `Tuple` (this is a good example of why I think ultimately the breaking change is called for).

Are you sure this should be removed?  If this PR to get merged, it seems like a clear TODO that this should be handled at some point and having it here would make it a lot easier to keep track of.

src/Enzyme.jl
ExpandingMan: I just added a bunch of tests from the other PR that cover part of this, but I still need to go through and comprehensively check for every method when I get a chance.  I already did some of this in testing on my end (not in the repo) so it shouldn't be too bad, though I'm probably missing some edge cases like different chunk sizes.

ext/EnzymeStaticArraysExt.jl
wsmoses: I suppose that's fair. My concern here is just that commented out code is never run, and thus can quickly get out of data (and confuse folks).

I'm okay with leaving it here, or putting it in an issue with the NOTE for todo, or whatever you prefer

src/Enzyme.jl
wsmoses: Okay if that's the case, it would definitely be helpful to have a docstring stating what the semantics are (as distinct from others, since I can see people trying to use derivative and then getting confused).

We could also pick a different name here, like unstructured_derivative or something (which would likely dissuade folks from using it in contrast to gradient or jacobian).

wsmoses: Also this function will fail on a chunk size of 1 (duplicated, not batchduplicated should be use for such cases).

src/Enzyme.jl
wsmoses: I guess I'm not sure what change to autodiff would enable that? (i.e. you can use autodiff to implement this, the difference being for mutable (e.g. duplicated) vs immutable (e.g. active) data you need to pass it as an argument vs get it as a result). The reason is described a bit here: https://enzyme.mit.edu/julia/stable/faq/#Mixed-activity

But basically (and feel free to extend docs to explain this more if unclear), suppose you have a square function you want to differentiate, that takes a float

```julia
function grad_square(x, dx)
    dx += 2 * x
    return nothing
end
x = 3.14
dx = make_zero(x)
grad_square(x, dx)
@show dx # 0.0, not the intended 2 * 3.14
```

if it's a float dx is passed by value so we won't see it changed in the outer scope.

MixedDuplicated is a funky combo of the two that we've implemented to support code in either convention (including a struct of a float and an array), though at a slight perf loss.

wsmoses: those docs should probably be updated now to say we have a MixedDuplicated type which means you don't need to change your original code to be ref-based, and you'd pass `MixedDuplicated(x, Ref(dx))` as an input.

ext/EnzymeStaticArraysExt.jl
ExpandingMan: Well, I suppose we are going to need to tackle the subject of how exactly breaking changes will actually work.  (Of course this could be uncommented then.)  By this point I assume from your interest in these PR's that you agree that there are some (mildly) breaking changes needed here for sake of consistency, but I don't know what the time frame for that would look like.  Presumably when we get through these PR's we can have a more detailed (and visible) conversation about that.

src/Enzyme.jl
ExpandingMan: Ok, fixed, see below.

test/runtests.jl
```diff
@@ -2851,6 +2851,71 @@ end
...
+    @test tplapprox(Enzyme.unstructured_derivative(Enzyme.Forward, x -> 2*x, 3.0, Val(2)), (2.0,))
+    @test tplapprox(Enzyme.unstructured_derivative(Enzyme.Forward, x -> [x, 2*x], 3.0), ([1.0,2.0],))
+    @test tplapprox(Enzyme.unstructured_derivative(Enzyme.Forward, x -> [x, 2*x], 3.0, Val(2)), ([1.0,2.0],))
+    @test tplapprox(Enzyme.unstructured_derivative(Enzyme.Forward, x -> sum(abs2, x), [2.0, 3.0]), (4.0, 6.0))
+    @test tplapprox(Enzyme.unstructured_derivative(Enzyme.Forward, x -> sum(abs2, x), [2.0, 3.0], Val(2)), (4.0, 6.0))
+    @test tplapprox(Enzyme.unstructured_derivative(Enzyme.Forward, x -> [-x[2], x[1]], [-2.0, 1.0]), ([0.0,1.0],[-1.0,0.0]))
+    @test tplapprox(Enzyme.unstructured_derivative(Enzyme.Forward, x -> [-x[2], x[1]], [-2.0, 1.0], Val(2)), ([0.0,1.0],[-1.0,0.0]))
+
+    x = 3.0
+
```
wsmoses, line 2868: let's add some comments just so its clear, 

This is scalar -> scalar

test/runtests.jl
```diff
@@ -2851,6 +2851,71 @@ end
...
+    @test Enzyme.gradient(Enzyme.Forward, x -> x^2, x) ≈ 6.0
+    @test Enzyme.gradient(Enzyme.Reverse, x -> x^2, x) ≈ 6.0
+
+    @test Enzyme.gradient(Enzyme.Forward, x -> [x, x^2], x) ≈ [1.0, 6.0]
```
wsmoses, line 2872: scalar -> vector

test/runtests.jl
```diff
@@ -2851,6 +2851,71 @@ end
...
+    @test_broken Enzyme.gradient(Enzyme.Reverse, x -> [x, x^2], x) ≈ [1.0, 6.0]
+
```
wsmoses, line 2874: can we also add scalar -> vector

and also scalar -> struct

test/runtests.jl
```diff
@@ -2851,6 +2851,71 @@ end
...
+    #NOTE: the discrepancy here is the same issue with the interface that causes there to be much
+    #special handling of gradients and jacobians.  Presumably these should be made consistent at
+    #some point but it would be a breaking change.
+    @test tplapprox(Enzyme.gradient(Enzyme.Forward, x -> sum(2*x), [1.0,1.0]), (2.0, 2.0))
```
wsmoses, line 2878: okay we now have vector -> scalar

test/runtests.jl
```diff
@@ -2851,6 +2851,71 @@ end
...
+    @test Enzyme.gradient(Enzyme.Reverse, x -> sum(2*x), [1.0,1.0]) ≈ [2.0, 2.0]
```
wsmoses, line 2879: can we also add matrix -> scalar

test/runtests.jl
```diff
@@ -2851,6 +2851,71 @@ end
...
+
+    @test Enzyme.gradient(Enzyme.Forward, x -> 2*x, [1.0,1.0]) ≈ Float64[2 0; 0 2]
```
wsmoses, line 2881: we now have vector->vector

test/runtests.jl
```diff
@@ -2851,6 +2851,71 @@ end
...
+    @test_broken Enzyme.gradient(Enzyme.Reverse, x -> 2*x, [1.0,1.0]) ≈ Float64[2 0; 0 2]
+
```
wsmoses, line 2883: can we add matrix->matrix?

test/runtests.jl
```diff
@@ -2851,6 +2851,71 @@ end
...
+    x = 1.0
```
wsmoses, line 2884: and can we add struct -> scalar
struct -> vector
struct -> matrix
struct -> struct

test/runtests.jl
```diff
@@ -2851,6 +2851,71 @@ end
...
+    @test gradient(Forward, x -> 2*x, x) ≈ 2.0
+    @test gradient(Reverse, x -> 2*x, x) ≈ 2.0
```
wsmoses, line 2886: we may also need versions of the above with different chunk sizes, deferred, and also returning the original return vs not (when supported in mode)

test/runtests.jl
```diff
@@ -2851,6 +2851,71 @@ end
...
+    @test tplapprox(Enzyme.unstructured_derivative(Enzyme.Forward, x -> 2*x, 3.0), (2.0,))
+    @test tplapprox(Enzyme.unstructured_derivative(Enzyme.Forward, x -> 2*x, 3.0, Val(2)), (2.0,))
+    @test tplapprox(Enzyme.unstructured_derivative(Enzyme.Forward, x -> [x, 2*x], 3.0), ([1.0,2.0],))
+    @test tplapprox(Enzyme.unstructured_derivative(Enzyme.Forward, x -> [x, 2*x], 3.0, Val(2)), ([1.0,2.0],))
+    @test tplapprox(Enzyme.unstructured_derivative(Enzyme.Forward, x -> sum(abs2, x), [2.0, 3.0]), (4.0, 6.0))
+    @test tplapprox(Enzyme.unstructured_derivative(Enzyme.Forward, x -> sum(abs2, x), [2.0, 3.0], Val(2)), (4.0, 6.0))
+    @test tplapprox(Enzyme.unstructured_derivative(Enzyme.Forward, x -> [-x[2], x[1]], [-2.0, 1.0]), ([0.0,1.0],[-1.0,0.0]))
+    @test tplapprox(Enzyme.unstructured_derivative(Enzyme.Forward, x -> [-x[2], x[1]], [-2.0, 1.0], Val(2)), ([0.0,1.0],[-1.0,0.0]))
+
+    x = 3.0
```
wsmoses, line 2867: for ease can we just call this scalar, and the other one vector, so they don't need overwriting?

ExpandingMan:
I'm working on adding all the various edge cases, but I've hit a bit of a snag.  The $\partial\textrm{matrix}/\partial\textrm{vector}$ case is failing with some very strange stack overflow errors inside `autodiff`.  So far I'm having a very hard time understanding how I could have introduced these, but (unlike all other cases) I cannot reproduce them on latest tagged.

It could be some unrelated change that happened more recently than latest tagged.  Will take me a bit to sort this out.

ExpandingMan:
Ah, I have confirmed that I did *NOT* introduce these errors, they are happening on latest commits to `main`.  I will explicitly mark them out for you in the PR.  You may want to test them out and fix them before trying to merge this.

wsmoses:
Hm maybe in this case it’s wise to split up this PR even further to just
add the tests (which confirm current main), then we add this PR stop that

ExpandingMan:
I suppose that's not hard to do, but there is a slight caveat: there is a trivial mistake in which a `ReverseMode` constructor is being called with too few type parameters in `main` which breaks a whole bunch of cases which this PR also fixes.  This is not breaking because it was clearly a bug in the first place, but it would cause a whole bunch of these tests to fail.  I think the easiest thing to do would be to fix those (very trivial) errors in the tests PR so as not having to mark a whole bunch of reverse mode stuff as broken.

wsmoses:
I believe this is since superceeded by API fixes on main, reopen and let me know if otherwise.

PR #1773, opened 2024-09-02
more comprehensive unit tests for gradient and jacobian

ExpandingMan:
The need for this arose out of #1760 as a stepping stone to that PR.

This also includes a small fix of a trivial issue in reverse mode gradient and jacobian which was causing a bunch of errors.

@wsmoses see the `WARN` comments.  These were recently broken.  Could you perhaps fix them?  If you can I will fix and expand them here.


test/runtests.jl
```diff
@@ -2851,6 +2851,126 @@ end
...
+    i3::Float64
+end
+struct OutStruct
+    i1::Float64
+    i2::Float64
+    i3::Float64
+end
+
+@testset "Gradient and Jacobian Outputs" begin
+    x = 3.0
```
wsmoses, line 2870: Can you call this scalar?

ExpandingMan: Not sure what you mean.

wsmoses:
@ExpandingMan I expanded and rewrote the tests which all look good to me.

Can you investigate/fix why 1.6 fails. And for 1.10, hvcat (aka array constructor syntax) was broken by SparseArrays, can the tests be rewritten to allocate an array then set the results instead)?

ExpandingMan:
Both of these things are baffling to me.

There is no apparent reason for 1.6 to fail that I can see, (albeit with a bunch of spurious output, which is maybe indicative of an improperly handled exception?), at the end of the log it merely says `Package Enzyme errored during testing (exit code: 3221225477)`.

The 1.10 error also makes no sense to me, and is a lot more concerning, since the way it is written seems like a very real use case and ideally not something to be fudged over in tests.  I don't understand how it's hitting that stack overflow, why is it hitting SparseArrays methods?  I'm reasonably sure calls like this were working a few commits ago, or possibly I was looking at some subtle variation of them which I now cannot distinguish.

vchuravy:
1.6 looks like:

```
Gradient and Jacobian Outputs: Error During Test at /home/runner/work/Enzyme.jl/Enzyme.jl/test/runtests.jl:2984
  Test threw exception
  Expression: Enzyme.jacobian(Enzyme.Forward, (x->begin
                [x[1] * x[2] cos(x[1]) + x[2]; exp(x[2]) x[1]]
            end), vector) ≈ [vector[2] -(sin(vector[1])); 0.0 1.0; vector[1] 1.0; exp(vector[2]) 0.0]
  DimensionMismatch("dimensions must match: a has dims (Base.OneTo(2), Base.OneTo(2), Base.OneTo(2)), b has dims (Base.OneTo(4), Base.OneTo(2)), mismatch at 1")
  Stacktrace:
   [1] promote_shape(a::Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}, Base.OneTo{Int64}}, b::Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}})
     @ Base ./indices.jl:178
   [2] promote_shape
     @ ./indices.jl:169 [inlined]
   [3] -(A::Array{Float64, 3}, B::Matrix{Float64})
     @ Base ./arraymath.jl:38
   [4] isapprox(x::Array{Float64, 3}, y::Matrix{Float64}; atol::Int64, rtol::Float64, nans::Bool, norm::typeof(norm))
     @ LinearAlgebra /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/generic.jl:1657
   [5] isapprox(x::Array{Float64, 3}, y::Matrix{Float64})
     @ LinearAlgebra /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/generic.jl:1657
   [6] eval_test(evaluated::Expr, quoted::Expr, source::LineNumberNode, negate::Bool)
     @ Test /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:277
   [7] macro expansion
     @ ~/work/Enzyme.jl/Enzyme.jl/test/runtests.jl:2984 [inlined]
   [8] macro expansion
     @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1151 [inlined]
   [9] top-level scope
     @ ~/work/Enzyme.jl/Enzyme.jl/test/runtests.jl:2911
```

(Best to use Linux not Windows for logs)

ExpandingMan:
I just did the following
- Replaced the half-assed `tplapprox` with `≃` (`\simeq`) which should behave better.  This was breaking a whole bunch of examples in mysterious ways.  (I have a new appreciation for `isapprox` which actually needs to do a little bit of math and therefore know something about the algebra of its arguments, hence why it does not "just work" on tuples and all sorts of other things.)
- Replaced explicit array constructors within anonymous functions by `mkarray` hack (essentially force it to use `vcat` instead of `hvcat`).  This feels awful, and I'm not yet sure it'll fix 1.6 as of writing, we'll see.  @wsmoses are you *sure* this is the fault of `Base` and stdlibs?  It seems very likely people are going to hit this.  On any earlier version I'd say who cares, but 1.10 is LTS and not sure whether it'll get fixed there in the foreseeable future.

ExpandingMan:
It looks like this fixes ("fixes") 1.10 but all 1.6 are now failing because a lot of these jobs for some reason hit an error where they can't find `libEnzyme` (which wasn't the case last time).

wsmoses:
1.6-1.9 are now presently broken by an issue in Yggdrassil: https://buildkite.com/julialang/yggdrasil/builds/13010#0191c04d-60fc-438f-a5fb-67945bfa910b

@vchuravy my attempted workaround was to permit the old and new jlls [required for recent 1.10+], but 1.6 still picked up the new jll and hit the above error. What is the right way to workaround the Yggy failure atm?

wsmoses:
cc @giordano

wsmoses:
@ExpandingMan so you can debug while yggy remains broken I pinned you to the old jll for now, which should let tests actually run on old versions

test/runtests.jl
```diff
@@ -2899,6 +2908,253 @@ end
...
+    i2::Float64
+    i3::Float64
+end
+struct OutStruct
+    i1::Float64
+    i2::Float64
+    i3::Float64
+end
+
+for (A, B) ∈ Iterators.product((:InpStruct, :OutStruct), (:InpStruct, :OutStruct))
```
wsmoses, line 2924: InpStructt should only compare against InpStruct and not Outstruct

ExpandingMan:
The test failures on 1.10 are from it seemingly ignoring the rules in the tests in `rules.jl`, which seems unlikely to be related to this PR.

ExpandingMan:
Ok I *think* all that's left is whatever is going on with the rules now.

ExpandingMan:
Not sure what the deal is with the segfaults but the failures here probably are unrelated?

wsmoses:
yeah this is part of the "weird things per release infra changes", what is expected to pass currently passes.

Issue #1812, opened 2024-09-12
No method matching Duplicated(::Base.MappingRF{...}, ::Base.RefValue{Base.MappingRF{…}})


mhauru:
MWE:
```julia
using Random: randn
using Enzyme: Enzyme
using Turing: Turing

Enzyme.API.runtimeActivity!(true)

Turing.@model function MvDirichletWithManualAccumulation(w, doc)
    β ~ Turing.filldist(Turing.Dirichlet([1.0, 1.0]), 2)
    log_product = log.(β)
    Turing.@addlogprob! sum(log_product[CartesianIndex.(w, doc)])
end

model = MvDirichletWithManualAccumulation([1, 1, 1, 1], [1, 1, 2, 2])
ctx = Turing.DefaultContext()
vi = Turing.VarInfo(model)
vi_linked = Turing.link(vi, model)
ldp = Turing.LogDensityFunction(vi_linked, model, ctx)
f = Base.Fix1(Turing.LogDensityProblems.logdensity, ldp)
d = Turing.LogDensityProblems.dimension(ldp)
x = randn(d)

Enzyme.gradient(Enzyme.Reverse, Enzyme.Const(f), x)
```

Output:
```julia
ERROR: LoadError: MethodError: no method matching EnzymeCore.Duplicated(::Base.MappingRF{Base.Fix1{…}, Base.BottomRF{…}}, ::Base.RefValue{Base.MappingRF{…}})

Closest candidates are:
  EnzymeCore.Duplicated(::T1, ::T1) where T1
   @ EnzymeCore ~/.julia/packages/EnzymeCore/jSKtE/src/EnzymeCore.jl:66
  EnzymeCore.Duplicated(::T1, ::T1, ::Bool) where T1
   @ EnzymeCore ~/.julia/packages/EnzymeCore/jSKtE/src/EnzymeCore.jl:66

Stacktrace:
  [1] runtime_generic_augfwd(activity::Type{…}, width::Val{…}, ModifiedBetween::Val{…}, RT::Val{…}, f::Base.MappingRF{…}, df::Base.RefValue{…}, primal_1::Base._InitialValue, shadow_1_1::Nothing, primal_2::SubArray{…}, shadow_2_1::SubArray{…})
    @ Enzyme.Compiler ~/.julia/packages/Enzyme/TiboG/src/rules/jitrules.jl:338
  [2] _foldl_impl
    @ ./reduce.jl:58
  [3] foldl_impl
[...]

Issue #1811, opened 2024-09-12
AssertionError: RT <: Array


mhauru:
MWE:
```julia
using Enzyme: Enzyme
using LinearAlgebra: Diagonal, transpose
using Turing: Turing

Turing.@model function demo(x=transpose([1.5 2.0;]))
    m ~ Turing.MvNormal(Diagonal([1.0, 1.0]))
    x .~ Turing.MvNormal(m, Diagonal([1.0, 1.0]))
    return nothing
end

model = demo()
ctx = Turing.DefaultContext()
vi = Turing.VarInfo(model)
ldp = Turing.LogDensityFunction(vi, model, ctx)
f = Base.Fix1(Turing.LogDensityProblems.logdensity, ldp)
x = [0.0, 0.0]
Enzyme.gradient(Enzyme.Reverse, Enzyme.Const(f), x)
```

Output
```julia
ERROR: LoadError: AssertionError: RT <: Array
Stacktrace:
  [1] abs_typeof(arg::LLVM.CallInst, partial::Bool)
    @ Enzyme.Compiler ~/.julia/packages/Enzyme/TiboG/src/absint.jl:266
  [2] abs_typeof
    @ ~/.julia/packages/Enzyme/TiboG/src/absint.jl:153 [inlined]
  [3] codegen(output::Symbol, job::GPUCompiler.CompilerJob{…}; libraries::Bool, deferred_codegen::Bool, optimize::Bool, toplevel::Bool, strip::Bool, validate::Bool, only_entry::Bool, parent_job::Nothing)
    @ Enzyme.Compiler ~/.julia/packages/Enzyme/TiboG/src/compiler.jl:6245
  [4] codegen
    @ ~/.julia/packages/Enzyme/TiboG/src/compiler.jl:5614 [inlined]
  [5] _thunk(job::GPUCompiler.CompilerJob{Enzyme.Compiler.EnzymeTarget, Enzyme.Compiler.EnzymeCompilerParams}, postopt::Bool)
    @ Enzyme.Compiler ~/.julia/packages/Enzyme/TiboG/src/compiler.jl:7241
  [6] _thunk
    @ ~/.julia/packages/Enzyme/TiboG/src/compiler.jl:7241 [inlined]
  [7] cached_compilation
    @ ~/.julia/packages/Enzyme/TiboG/src/compiler.jl:7282 [inlined]
[...]

Issue #1807, opened 2024-09-11
Forward mode gradient AssertionError


mhauru:
MWE:
```julia
using Turing: Turing
using Enzyme: Enzyme

Enzyme.API.runtimeActivity!(true)

Turing.@model function demo()
    x ~ Turing.Normal()
    y ~ Turing.Normal()
end

model = demo()
ctx = Turing.DefaultContext()
vi = Turing.VarInfo(model)
ldp = Turing.LogDensityFunction(vi, model, ctx)
x = randn(2)
f = Base.Fix1(Turing.LogDensityProblems.logdensity, ldp)
Enzyme.gradient(Enzyme.Forward, Enzyme.Const(f), x)
```
Output:
```julia
ERROR: LoadError: AssertionError: width == 1
Stacktrace:
  [1]
    @ Enzyme.Compiler ~/.julia/packages/Enzyme/TiboG/src/compiler.jl:4277
  [2] enzyme!(job::GPUCompiler.CompilerJob{…}, mod::LLVM.Module, primalf::LLVM.Function, TT::Type, mode::Enzyme.API.CDerivativeMode, width::Int64, parallel::Bool, actualRetType::Type, wrap::Bool, modifiedBetween::Tuple{…}, returnPrimal::Bool, expectedTapeType::Type, loweredArgs::Set{…}, boxedArgs::Set{…})
    @ Enzyme.Compiler ~/.julia/packages/Enzyme/TiboG/src/compiler.jl:4188
  [3] codegen(output::Symbol, job::GPUCompiler.CompilerJob{…}; libraries::Bool, deferred_codegen::Bool, optimize::Bool, toplevel::Bool, strip::Bool, validate::Bool, only_entry::Bool, parent_job::Nothing)
    @ Enzyme.Compiler ~/.julia/packages/Enzyme/TiboG/src/compiler.jl:6438
  [4] codegen
    @ ~/.julia/packages/Enzyme/TiboG/src/compiler.jl:5614 [inlined]
  [5] _thunk(job::GPUCompiler.CompilerJob{Enzyme.Compiler.EnzymeTarget, Enzyme.Compiler.EnzymeCompilerParams}, postopt::Bool)
    @ Enzyme.Compiler ~/.julia/packages/Enzyme/TiboG/src/compiler.jl:7241
  [6] _thunk
    @ ~/.julia/packages/Enzyme/TiboG/src/compiler.jl:7241 [inlined]
  [7] cached_compilation
[...]

PR #1813, opened 2024-09-12
Add Turing integration tests

mhauru:
Makes log density functions out of a bunch of Turing.jl models and checks that Enzyme can differentiate them, both in Forward and Reverse modes.

Related issues that should probably be sorted before merging this:
* #1812
* #1811
* #1807


wsmoses:
Can turing and all dependencies be pinned to exact versions

mhauru:
I changed the dependency from Turing to DynamicPPL and Distributions since we don't do any sampling. I pinnned those two and their major depencencies: AbstractMCMC, AbstractPPL, Bijectors, LogDensityProblems, LogDensityProblemsAD. Can pin _all_ the dependencies if you'd like (BangBang, OrderedCollections, etc.), but I wonder if in that case it would make more sense to just commit the Manifest.toml.

wsmoses:
@penelopeysm can you help get this integrated with the modern CI workflow for integration tests?

PR #1816, opened 2024-09-12
Runtime activity in mode

wsmoses:
Note this is a very breaking change for rules.

Specifically now forward rules take a config (to understand runtime activity so they can forward it, e.g. for KA)


wsmoses:
So the issue is RevConfig contains fields that don't apply to fwd mode like
if a var is overwritten.

Am fine with an abstract class though what specifically are you thinking

Issue #1761, opened 2024-08-29
Check if function is being called inside `autodiff`


avik-pal:
### `@avik-pal`

In Enzyme can we check if a function is being called within an autodiff? Something equivalent to the following ChainRules version

```julia
within_gradient(_) = False()
CRC.rrule(::typeof(within_gradient), x) = True(), _ -> (∂∅, ∂∅)
```

I tried defining:

```julia
function EnzymeRules.forward(
        ::EnzymeCore.Const{typeof(within_gradient)}, ::Type{RT}, x) where {RT}
    error("within_gradient")
end
```

### `@wsmoses`

I think this is getting optimized out lol
But yeah we can add something (c++ interface already has so we should bring parity to Julia for that :P)
(Julia will still constprop through things with enzyme rules atm)

avik-pal:
@wsmoses is this as simple as adding a `ccall` somewhere? I can make a PR if you point me to the function

wsmoses:
we have this magic function: https://github.com/EnzymeAD/Enzyme/blob/7f614f43808e5bd3960f3712ac880b38eedc01d6/enzyme/test/Integration/ReverseMode/mycos.c#L39

which has the semantics that __enzyme_iter(x, y) = x if not differentiated, x + y on first order, etc [used for sake of keeping right order of taylor series].

It probably needs a bit of julia integration, but maybe something like this?

avik-pal:
Is this not an "actual" function present in the binary? I tried

```julia
function within_autodiff()
    return ccall((:__enzyme_iter, libEnzyme), UInt64, (UInt64, UInt64), 0, 1) != 0
end
```

But that symbol is not present

wsmoses:
No it’s a fictitious one that Enzyme will replace with a definition (or
rather during AD transform as above).

Just like the enzyme pass we may need to tell enzyme to replace uses after
AD of the function with the first arg.

wsmoses:
hm okay I realize this won't quite work the same in Julia since we won't have control over the non differentiated optimizatio pipeline.

I think the solution here is we just need to tell the abstract interpreter to block constprop for a special function which we rewrite if inside an enzyme differentiated context. cc @vchuravy @aviatesk 

I know some early work on absint overriding was here: https://github.com/EnzymeAD/Enzyme.jl/pull/1443 but I think needs someone to push it forward atm

wsmoses:
[...]

PR #1851, opened 2024-09-18
Add within autodiff cmd

wsmoses:


wsmoses:
@aviatesk what's the right way to do this properly?

I think it's blocking for @avik-pal / @ChrisRackauckas

wsmoses:
https://github.com/EnzymeAD/Enzyme.jl/issues/1761

aviatesk:
If the overlay method table was available, that would be ideal. If that's still difficult, we might have to stick with the current implementation approach.

wsmoses:
So as is this fails atm (presumably because Const prop happens before we can fix this?)

test/abi.jl
```diff
@@ -460,6 +460,11 @@ abssum(x) = sum(abs2, x);
 
 mulsin(x) = sin(x[1] * x[2])
 
+@testset "within_autodiff" begin
+    @test Enzyme.within_autodiff()
```
vchuravy, line 464:
```suggestion
    @test !Enzyme.within_autodiff()
```

test/abi.jl
```diff
@@ -460,6 +460,11 @@ abssum(x) = sum(abs2, x);
...
+    @test Enzyme.autodiff(ForwardWithPrimal, Enzyme.within_autodiff)[1]
```
vchuravy, line 465: This might need: `Enzyme.autodiff(ForwardWithPrimal, ()->Enzyme.within_autodiff())[1]
`

src/compiler/interpreter.jl
```diff
@@ -220,11 +220,23 @@ struct AutodiffCallInfo <: CallInfo
...
         max_methods::Int = get_max_methods(interp, f, sv))
 
     (; fargs, argtypes) = arginfo
+    
+    if f === Enzyme.within_autodiff
+          arginfo2 = ArgInfo(
+            fargs isa Nothing ? nothing : [:(Enzyme.Interpreter.always_true), fargs[2:end]...],
+            [Core.Const(Enzyme.Interpreter.always_true), argtypes[2:end]...]
+          )
+          return abstract_call_known(
```
vchuravy, line 236:
```
    if length(argtypes) != 0
        @static if VERSION < v"1.11.0-"
            return CallMeta(Union{}, Effects(), NoCallInfo())
        else
            return CallMeta(Union{}, Union{}, Effects(), NoCallInfo())
        end
    end
    @static if VERSION < v"1.11.0-"
        return CallMeta(Const(true), CC.EFFECTS_TOTAL, MethodResultPure())
    else
        return CallMeta(Const(true), Union{}, CC.EFFECTS_TOTAL, MethodResultPure())
    end
end
```

src/Enzyme.jl
avik-pal, line 1: Can this be added to Core?

wsmoses: done [though still getting this to work properly atm]

PR #1824, opened 2024-09-14
split up tests

ExpandingMan:
I've split up the unit tests using the following criteria:
- `runtests.jl` now just links to other files, there are a large number of separate test files.
- Every test file **must** be testable and self contained simply by `include`ing that file, no more running full tests every single time.
- No nesting of `include` that's not apparent from `runtests.jl`.  That is, only `runtests.jl` has `include` other than `utils.jl`, so you can see all the files that are run from `runtests.jl` alone.
- `using` statements other than for `Enzyme` and `Test` are contained within the appropriate files, that way you don't necessarily need all the dependencies to run only the files you want.

I tried to organize the files roughly by subject.  This wasn't always easy as the existing tests were pretty mixed up, and I only have a poor understanding of some of the internals myself, you might want to shuffle some of these around.  You can see cases where I gave up in the `misc.jl` file.

I also eliminated some of the most redundant `Enzyme.` qualifiers in cases where they are used over and over.

My goal was **NOT** to change the content of any tests, the only thing I actually changed here was adding a very simple test for `StaticArrays` based on a recent issue.

PR #1868, opened 2024-09-20
split up tests

ExpandingMan:
This is a re-do of #1824 for post 0.13.

To reiterate, the goal here is to roughly group the tests by the functionality they are testing and separate them into a large number of files which should be possible to run entirely independently.  Again the way I have grouped these tests may not always make a lot of sense, it seems likely they'll get shuffled around more in the future (or before this gets merged) but this seems like a decent start.

There is still a major issue that I've so far been unable to resolve: a setfault in `arrays.jl`, which I've seen before when trying to split up the tests.  It sure looks to me like the fact that it segfaults is an enzyme bug, as I don't see any reason it should do that, especially merely as a result of being in a different file.

As of writing, everything else in here passes as expected for me (x86 linux).


test/arrays.jl
```diff
@@ -0,0 +1,385 @@
...
+
+    xmean = mean(x_, dims=(1, 2, 5))
+    xvar = var(x_, corrected=false, mean=xmean, dims=(1, 2, 5))
+
+    return sum(abs2, bc2_affine_normalize(identity, x_, xmean, xvar, scale_, bias_, 1e-5))
+end
+
+#FUCK: I'm *still* getting segfaults here????
+#=
+@testset "Broadcast noalias" begin
```
ExpandingMan, line 246: The segfault I can't explain happens in here.

test/higherorder.jl
```diff
@@ -0,0 +1,74 @@
...
+
+    x = [1.0]
+    v = [-1.0]
+    hv = make_zero(v)
+    tmp = similar(x)
+
+    #WARN: this fails with an assertion error from somewhere deep inside enzyme/compiler and I 
+    # have absolutely no idea why
+    # definitely fails when file is run in isolation, not sure about otherwise
+    f_hvp!(hv, x, v, tmp)
```
ExpandingMan, line 71: The assertion error happens in this call.

ExpandingMan: Turns out this was due to `LinearAlgebra` not being imported, and `dot` not being defined.  The error message was unrecognizable to me, but there was clearly an error introduced by me here, now fixed.

wsmoses:
open a mwe issue for each of these?

ExpandingMan:
#1869 and #1870

(the former is clearly a real bug, the latter is just confusing error messages and probably not a bug)

wsmoses:
I think we need to fix the segfault before this goes in.

Unfortunately I can't reproduce locally. Can you setup a tmate instance with it on it or something, and send me via slack

ExpandingMan:
Ok.  I don't think I've ever let the segfaulting test run in CI/CD, let's see what that does first, I've just uncommented it.

ExpandingMan:
Frankly that could take forever, there is quite a lot of stuff to get split up.  I'd like to get it in, but if you think it's going to be that difficult, I'll leave it to you (if this is indeed something you feel is worth doing).  Even if I did this piecemeal I wouldn't know when to do what.

wsmoses:
maybe do like 3-4 in a first pass (and move out utils)?

ExpandingMan:
Could we maybe just try to keep track of which tests are added between the PR going up and it being merged?  Would it really be that hard?

wsmoses:
I mean sure if you want, but also bear in mind that moving everything together means it'll take longer to review.....which means another test PR might get in before then....necessitating another review...

PR #1875, opened 2024-09-21
fix some cases of gradient/jacobian with StaticArrays

ExpandingMan:
This fixes at least some cases of functions taking and return static arrays being returned as `Array` (and the associated allocations).

An example of a case this fixes is
```julia
f1(x) = @SVector Float64[x[1], 2*x[2]]

x = @SVector ones(2)

gradient(Forward, f1, x)
```

There were two things wrong with the `tupstack` function in current `main`.  The first is that `cat` always returns an `Array`.  This may be a missing method in StaticArrays, but I suspect that the reason it's missing is that it's hard to statically infer the resulting shape (and it would rely on static inference of the value of `dims`).  Suffice it to say, I think it's safe to assume that `cat` can never be used with staticarrays.

The second issue is that when `reshape` is used with static arrays, one must use `Size` arguments to return a static array, to indicate that the size can be statically inferred.

This works in the cases I have tested but looks fragile.  I still need to confirm more thoroughly that this does not break any of the higher rank cases.


wsmoses:
Add a test?

ExpandingMan:
Can we try to get #1868 merged and then I'll add a whole big batch of StaticArrays tests here?

ExpandingMan:
I've now added a bunch of tests.  A couple of things to note:
- I don't actually try to check for allocations here, as it requires a somewhat careful setup to work consistently (at least without `BenchmarkTools`).  In my testing there are no allocations as long as the return types are those I'm testing for.
- Ideally, `gradient` would return an `SVector` instead of a `TupleArray` so that it retains statically inferable shape information (this requires things like `Size` and `Length` and not just type parameters with a different interface).  I believe this could be done without breaking, but I thought it best to leave it for a future PR, since it would require changes inside of the `gradient` functions.  I have included a `convert` method.
- The `broken` reverse mode tests here should *already be broken* in main.  I'm not even sure some of them are _supposed_ to work, I suspect some should not, in which case let me know and I'll remove those tests.

wsmoses:
A tuplevector has a static known shape like an SArray so imo it’s fine

test/runtests.jl
```diff
@@ -2854,6 +2854,51 @@ end
...
+    @test dx isa SVector
+    @test convert(SArray, dx) == [2.0, 2.0]  # test to make sure conversion works
+    @test_broken gradient(Reverse, f1, x)[1] isa SMatrix
+    @test_broken gradient(Reverse, f1, x)[1] == [0 1.0; 0 2.0]
+    @test_broken jacobian(Reverse, f2, x)[1] isa SArray
+    @test_broken jacobian(Reverse, f2, x)[1] == reshape(Float64[0,0,1,2,1,2,0,0], (2,2,2))
+
+    x = @SMatrix Float64[1 2; 3 4]
+
+    @test_broken gradient(Reverse, f1, x)[1] isa SArray
```
wsmoses, line 2896: how is this broken at the moment?

ExpandingMan:
It does in principle, but it does not satisfy the interface that staticarrays uses to know its shapes.  The result will be that if you try to operate on it with functions from static arrays it will treat it as dynamic.  For example
```julia
julia> Size(@SVector([1,2]))
Size(2,)

julia> Size(Enzyme.TupleArray((1,2), (2,)))
Size(StaticArraysCore.Dynamic(),)
```
One could overload `Size` and `Length` but the interface expects it to be of a `StaticArray` type, so I'm not sure how far that'll get.  Calling one of them on the output of `gradient` or `jacobian` and it usually working but suddenly gives you `Dynamic()` on one particular `gradient` does not seem ideal.

wsmoses:
I mean that separately feels worth doing tbh

wsmoses:
@ExpandingMan this fails, once it passes, it seems fine to merge.

ExpandingMan:
`jacobian` is definitely exported but somehow it was saying it's not defined in `runtests.jl`, so weird.  Not going to try to figure it out now.  Hopefully this will pass.

ExpandingMan:
Are these macos and 1.11 failures expected?  They look like they have nothing to do with what is in this PR.

wsmoses:
They are, yes

PR #1895, opened 2024-09-26
Improve documentation of modes

### LINKED ISSUE #1881: Inconsistency in mode type parameters for `ReverseModeSplit`
The docstring and the type definition disagree on the order of type parameters (switch between `Width` and `RuntimeActivity`)

https://github.com/EnzymeAD/Enzyme.jl/blob/0c36c5af6e7ed02a25e5bb7485d6477f97ca7eed/lib/EnzymeCore/src/EnzymeCore.jl#L248-L261

gdalle:
Fix #1881 by adding systematic documentation for the `Mode` subtypes and helper functions. 
This PR doesn't change a single line of actual code.

### DIFF
--- a/lib/EnzymeCore/src/EnzymeCore.jl
+++ b/lib/EnzymeCore/src/EnzymeCore.jl
@@ -337,8 +337,18 @@ const ReverseHolomorphicWithPrimal = ReverseMode{true,false,DefaultABI, true, fa
 @inline set_runtime_activity(::ReverseMode{ReturnPrimal,RuntimeActivity,ABI,Holomorphic,ErrIfFuncWritten}, rt::Bool) where {ReturnPrimal,RuntimeActivity,ABI,Holomorphic,ErrIfFuncWritten} = ReverseMode{ReturnPrimal,rt,ABI,Holomorphic,ErrIfFuncWritten}()
 @inline clear_runtime_activity(::ReverseMode{ReturnPrimal,RuntimeActivity,ABI,Holomorphic,ErrIfFuncWritten}) where {ReturnPrimal,RuntimeActivity,ABI,Holomorphic,ErrIfFuncWritten} = ReverseMode{ReturnPrimal,false,ABI,Holomorphic,ErrIfFuncWritten}()
 
+"""
+    WithPrimal(::Mode)
+
+Return a new mode which includes the primal value.
+"""
 @inline WithPrimal(::ReverseMode{ReturnPrimal,RuntimeActivity,ABI,Holomorphic,ErrIfFuncWritten}) where {ReturnPrimal,RuntimeActivity,ABI,Holomorphic,ErrIfFuncWritten} = ReverseMode{true,RuntimeActivity,ABI,Holomorphic,ErrIfFuncWritten}()
 
+"""
+    NoPrimal(::Mode)
+
+Return a new mode which excludes the primal value.
+"""
 @inline NoPrimal(::ReverseMode{ReturnPrimal,RuntimeActivity,ABI,Holomorphic,ErrIfFuncWritten}) where {ReturnPrimal,RuntimeActivity,ABI,Holomorphic,ErrIfFuncWritten} = ReverseMode{false,RuntimeActivity,ABI,Holomorphic,ErrIfFuncWritten}()
 
 """
@@ -567,19 +577,5 @@ function clear_runtime_activity end
 Return a new mode with its [`ABI`](@ref) set to the chosen type.
 """
 function set_abi end
-
-"""
-    WithPrimal(::Mode)
-
-Return a new mode which includes the primal value.
-"""
-function WithPrimal end
-
-"""
-    NoPrimal(::Mode)
-
-Return a new mode which excludes the primal value.
-"""
-function NoPrimal end
 
 end # module EnzymeCore


lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -245,18 +349,47 @@ const ReverseHolomorphicWithPrimal = ReverseMode{true,false,DefaultABI, true, fa
...
+# Type parameters
+
+- `ReturnShadow`: whether to return the shadow return value from the augmented-forward.
+- `Width`: batch size (pick `0` to derive it automatically)
+- `ModifiedBetween`: `Tuple` of each argument's "modified between" state (pick `true` to derive it automatically).
+- other parameters: see [`ReverseMode`](@ref)
+
+!!! warning
+    The type parameters of `ReverseModeSplit` are not part of the public API and can change without notice.
+    Please use one of the following concrete instantiations instead:
```
wsmoses, line 373: same comment

lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -245,18 +349,47 @@ const ReverseHolomorphicWithPrimal = ReverseMode{true,false,DefaultABI, true, fa
...
+- `ReturnShadow`: whether to return the shadow return value from the augmented-forward.
+- `Width`: batch size (pick `0` to derive it automatically)
+- `ModifiedBetween`: `Tuple` of each argument's "modified between" state (pick `true` to derive it automatically).
+- other parameters: see [`ReverseMode`](@ref)
+
+!!! warning
+    The type parameters of `ReverseModeSplit` are not part of the public API and can change without notice.
+    Please use one of the following concrete instantiations instead (combined with setter functions like [`set_runtime_activity`](@ref)): 
+    - [`ReverseSplitNoPrimal`](@ref)
+    - [`ReverseSplitWithPrimal`](@ref)
```
wsmoses, line 375: also the helper functions here.

including also ReverseSplitModified and ReverseSplitWidth

gdalle: Shouldn't all helper functions be lowercase or uppercase? Those are easy to confuse with actual structs

wsmoses: No disagreement, but when I asked folks if those should be changed for the breaking change I didn’t see any feedback on.

So they will remain as they are (they significantly predate the other helper functions).

We can try to change in a future breaking release but they should still be referenced presently (for example you didn’t know about the existing width one!)

gdalle: According to Julia's interpretation of semantic versioning, those helper functions were never documented so they were never public in the first place. Thus you don't need a breaking release to change them.
That is precisely why I go to the trouble of adding docstrings everywhere, so that we get a clearer picture of what is public and what is not from now on.

wsmoses: In this context they were public since we told folks to use them. I agree it would've been nice to have had docstrings too -- but in this context I don't think they can be changed til next major bump

gdalle: @vchuravy thoughts?

wsmoses: in any case this is somewhat moot now per https://github.com/EnzymeAD/Enzyme.jl/pull/1898

wsmoses: Totally happy to come up with a good unified design for the helper functions and mark deprecations of the old ones to new choices, but the time for breaking changes that make it in to 0.13 has already passed.

Feel free to open a PR/issue to start designing their successor, but if we're adding docs, we need to document the functions we have now.

gdalle: Sounds good, I'll finish documenting the existing stuff in this PR tomorrow

gdalle:
@wsmoses this is up-to-date with `main` and ready for review nb 2

lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -189,50 +189,168 @@ end
...
+    clear_runtime_activity(::Mode)
+
+Return a new mode where runtime activity analysis is deactivated.
+"""
+function clear_runtime_activity end
+
+"""
+    set_abi(::Mode, ::Type{ABI})
+
+Return a new mode with its ABI set to the chosen type.
```
wsmoses, line 283: maybe make this `[`ABI`](@ref).`

lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -189,50 +189,168 @@ end
...
+
+# Type parameters
+
+- `ReturnPrimal`: whether to return the primal return value from the augmented-forward pass.
+- `Holomorphic`: Whether the complex result function is holomorphic and we should compute `d/dz`
+- other parameters: see [`Mode`](@ref)
+
+!!! warning
+    The type parameters of `ReverseMode` are not part of the public API and can change without notice.
+    Please use one of the following concrete instantiations instead (combined with setter functions like [`set_runtime_activity`](@ref)):
```
wsmoses, line 320: can you list the getter and setter functions here?

lib/EnzymeCore/src/EnzymeCore.jl
wsmoses: same comment to list/reference the helper functions here?

lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -189,50 +189,172 @@ end
...
+
+"""
+    clear_err_if_func_written(::Mode)
+
+Return a new mode which doesn't throw an error for attempts to write into an unannotated function object.
+"""
+function clear_err_if_func_written end
+
+"""
+    set_runtime_activity(::Mode)
```
wsmoses, line 266: This also has a version which takes a bool [whether to turn on or off] as the second param, as well as for a FwdConfig and RevConfig, if those can be doc'd as well

gdalle: done

lib/EnzymeCore/src/EnzymeCore.jl
gdalle: done

gdalle:
all good!

gdalle:
I've merged the latest changes in `main` and tried to minimize the diff for the final review

Issue #1821, opened 2024-09-13
Forward mode gradient error with argument of limited mutability


mhauru:
```julia
julia> f(x) = sum(x)
f (generic function with 1 method)

julia> x = Hermitian(randn(2,2))
2×2 Hermitian{Float64, Matrix{Float64}}:
  0.656689  -0.377647
 -0.377647  -0.505356

julia> Enzyme.gradient(Enzyme.Forward, Enzyme.Const(f), x)
ERROR: ArgumentError: Cannot set a non-diagonal index in a Hermitian matrix
Stacktrace:
 [1] setindex!
   @ ~/.julia/juliaup/julia-1.10.5+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/LinearAlgebra/src/symmetric.jl:264 [inlined]
 [2] _setindex!
   @ ./abstractarray.jl:1431 [inlined]
 [3] setindex!
   @ ./abstractarray.jl:1396 [inlined]
 [4] #93
   @ ~/.julia/packages/Enzyme/TiboG/src/Enzyme.jl:967 [inlined]
 [5] macro expansion
   @ ./ntuple.jl:72 [inlined]
 [6] ntuple(f::Enzyme.var"#93#94"{Hermitian{Float64, Matrix{Float64}}, Int64}, ::Val{4})
   @ Base ./ntuple.jl:69
 [7] onehot
   @ ~/.julia/packages/Enzyme/TiboG/src/Enzyme.jl:963 [inlined]
 [8] gradient(::EnzymeCore.ForwardMode{…}, f::EnzymeCore.Const{…}, x::Hermitian{…})
   @ Enzyme ~/.julia/packages/Enzyme/TiboG/src/Enzyme.jl:1132
 [9] top-level scope
   @ REPL[60]:1
Some type information was truncated. Use `show(err)` to see complete types.
```

Types like `Hermitian` and `Symmetric` break the assumption made in `onehot` that all elements of the array are independent variables that can be `setindex!`ed freely. Not sure if there's some general solution to this, but maybe catering for stdlib types would be enough?

wsmoses:
@ExpandingMan would you be interested in taking a stab at this?

ExpandingMan:
There's a complicated but efficient solution in which `onehot` is smart enough only to take the independent basis elements.  The real difficulty I see with that is that the output side of `gradient` would have to know about it to combine the results, which can potentially significantly increase its complexity.

[...]

PR #1934, opened 2024-10-01
inelegant solution to #1821

ExpandingMan:
This is a particularly inelegant solution to #1821.

What I do here is for `Symmetric` and `Hermitian` take a redundant basis.  This looks a bit strange because I both set the individual elements and also set the type of the resulting matrix.  The reason for this is that if I do not set all the elements of the underlying buffer, Enzyme's autodiff returns the wrong result, but the wrapper type must also be set because currently `BatchDuplicated` only allows arguments and bases to be of the same type.  I also go out of my way to call `similar` here (e.g. rather than `zeros`) to make it less likely we will run afoul of `BatchDuplicated`.

I *think* this works in general but I have not exhaustively checked it.  I'm not sure how enzyme would be expected to deal with complex arguments here, so I've only tested purely real cases so far.

To see an example of how it works, consider symmetric matrix $X$ and
$$f(X) = X_{11}^2 + 2 X_{12}^2 + 3 X_{21}^2 - X_{22}^2$$
For extra clarity, write $X = \left(\matrix{a & b \cr b & c}\right)$ so that
$$f(X) = a^2 + 5 b^2 - c^2$$
The derivative with respect to the off-diagonal element $b$ is clearly $10b$.

I impelement this in Julia via
```julia
f0(x) = x[1,1]^2 + 2*x[1,2]^2 + 3*x[2,1]^2 - x[2,2]^2
```
Then, for example
```julia
X = Hermitian(Float64[1 2; 2 3])
```
so that
```julia
julia> gradient(Forward, f0, X)[1]
2×2 Enzyme.TupleArray{Float64, (2, 2), 4, 2}:
  2.0  20.0
 20.0  -6.0
```
This also works if the underlying data for `X` is `[1.0 2.0; 0.0 3.0]`.

This will remain a draft until I come up with more comprehensive tests for it.

### DIFF
--- a/src/Enzyme.jl
+++ b/src/Enzyme.jl
@@ -1566,6 +1566,26 @@ end
     return (one(x),)
 end
 
+@inline function _herm_sym_onehot(type, x::AbstractMatrix, start=1, endl=length(x))
+    idxs = CartesianIndices(x)
+    ntuple(Val(endl - start + 1)) do i0
+        Base.@_inline_meta
+        i = start + i0 - 1
+        idx = idxs[i]
+        res = zeros(eltype(x), size(x))
+        res[idx] = 1
+        res[idx[2],idx[1]] = 1
+        type(res)
+    end
+end
+
+@inline onehot(x::Hermitian) = _herm_sym_onehot(Hermitian, x)
+@inline onehot(x::Symmetric) = _herm_sym_onehot(Symmetric, x)
+
+@inline onehot(x::Hermitian, start, endl) = _herm_sym_onehot(Hermitian, x, start, endl)
+@inline onehot(x::Symmetric, start, endl) = _herm_sym_onehot(Symmetric, x, start, endl)
+
+
 """
     gradient(::ReverseMode, f, args...)
 
--- a/src/Enzyme.jl
+++ b/src/Enzyme.jl
@@ -1566,15 +1566,20 @@ end
     return (one(x),)
 end
 
+function _is_symmed_index(a::CartesianIndex, b::CartesianIndex)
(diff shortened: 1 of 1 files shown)


ExpandingMan:
Alright, this looks like it works.  Again, while this may be a fix this solution is not very nice and involves a lot of redundancy, but I think anything better would be significantly more complicated.

I certainly would advise giving this a good think before merging.  Keep in mind that if a user is really determined, they can create this on their own using only the existing public API (`gradient` and `shadows`) so I would find it perfectly reasonable if you decided it was better to close this and leave it broken for the foreseeable future rather than adding more behavior that it would be better to later break.

src/Enzyme.jl
```diff
@@ -1566,6 +1566,31 @@ end
...
+
+@inline function _herm_sym_onehot(type, x::AbstractMatrix, start=1, endl=length(x))
+    idxs = CartesianIndices(x)
+    ntuple(Val(endl - start + 1)) do i0
+        Base.@_inline_meta
+        i = start + i0 - 1
+        idx = idxs[i]
+        res = similar(parent(x))
+        for idx2 ∈ CartesianIndices(x)
+            @inbounds res[idx2] = _is_symmed_index(idx, idx2) ? 1 : 0
```
wsmoses, line 1581: why do we need the check if symmed?

ExpandingMan: This is checking if it's either the index `idx` or its transpose.  In my testing enzyme would return the wrong result if I did not set both elements (as if I had not wrapped it in `Symmetric` even though I had).  I don't understand why that is.

PR #1943, opened 2024-10-08
Static array return for forward gradient

wsmoses:
@ExpandingMan


ExpandingMan:
Thanks.  Haven't tested it but LGTM.

PR #1954, opened 2024-10-10
Fix code coverage & update action versions

gdalle:
Hi there!
I looked into the lack of code coverage (e.g. in [this CI log](https://github.com/EnzymeAD/Enzyme.jl/actions/runs/11247429913/job/31270909247)), and it turns out you're using a very outdated version of the Codecov action, which severely limits your uploads.
The solution is to update the action and add a secret `CODECOV_TOKEN` to the repo, following the procedure I described here: <https://discourse.julialang.org/t/psa-new-version-of-codecov-action-requires-additional-setup/109857>.
I used the opportunity to update other action versions.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -54,13 +54,13 @@ jobs:
             version: '1.11'
             assertions: true
     steps:
-      - uses: actions/checkout@v2
-      - uses: julia-actions/setup-julia@v1
+      - uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@v2
         if: ${{ ! matrix.assertions }}
         with:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         if: ${{ matrix.assertions }}
         with:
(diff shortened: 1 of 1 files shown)


wsmoses (approved):

wsmoses:
@gdalle if I gave you relevant repo/org permissions (and let me know which ones) would you be able to setup the key?

gdalle:
Sure but it's literally a 15 second procedure and I explain it in great detail on the discourse post, so don't feel compelled to bring me along. I'd be honored though

gdalle:
The right way is to add an organization wide secret

gdalle:
And actually since we've seen a codecov report in the present PR, it looks like the setup already works?

wsmoses:
Cool, okay will merge

wsmoses:
@gdalle this is breaking CI with say "Error: Unable to resolve action `julia-actions/cache@v4`, unable to find version `v4`" for docs, etc

I'm going to revert and let you reopen with a fix.

gdalle:
done in #1977, sorry for the typo, it was v2 and not v4
using dependabot might help for these kinds of updates btw

Issue #1935, opened 2024-10-02
`active_reg_inner` with `justActive = true` incorrect with immutable types that can be incompletely initialized


danielwe:
In the process of achieving comprehensive test coverage for #1852 I've stumbled upon the following corner case:

```julia-repl
julia> using Enzyme

julia> struct Foo
           x::Float64
           y::Core.Box
           dummy(x) = new(x)
       end

julia> Enzyme.Compiler.active_reg_inner(Tuple{Foo}, (), nothing, Val(false)#=Val{justActive}=#)  # correct
MixedState::ActivityState = 3

julia> Enzyme.Compiler.active_reg_inner(Tuple{Foo}, (), nothing, Val(true)#=Val{justActive}=#)  # incorrect, should be ActiveState
AnyState::ActivityState = 0
```

As far as I understand, `justActive = true` is supposed to map MixedState to ActiveState and DupState to AnyState, for when you just want to know whether or not the type contains immutable active storage. This example breaks that mapping.

The problem remains if I replace `dummy` with `Foo`. I'm using `dummy` here to emphasize that it's the incomplete use of `new` in the struct definition that triggers the error, rather than something specific to inner constructor methods. The problem also remains if I add a fully initializing inner constructor next to the incomplete one.

The issue disappears with any of the following changes:

* Removing `dummy(x)`
* _Only_ invoking `new` in a fully initializing way, e.g., `new(x, Core.Box(nothing))`

wsmoses:
your understanding is correct and this is a bug needing investigation

danielwe:
Changing to `ismutabletype` does not break any tests locally. However, it's a very different condition, and it's unclear to me whether it's the correct one. For one, most abstract types are `!allocatedinline` but are not `ismutabletype`, so the set of field types that would end up in this branch is quite different.

Is this branch just a short-circuiting early return, or is it necessary for correctness? If it's the former, the condition only has to match a subset of the types that could use the early return, so `ismutabletype` might be an adequate solution.

PR #1958, opened 2024-10-10
make_zero! for immutable

vchuravy:
Recurse through the immutable and set the mutable fields to zero.

### DIFF
--- a/src/make_zero.jl
+++ b/src/make_zero.jl
@@ -198,7 +198,7 @@ function make_zero_immutable!(prev::T, seen::S)::T where {T<:Tuple,S}
 end
 
 function make_zero_immutable!(prev::NamedTuple{a,b}, seen::S)::NamedTuple{a,b} where {a,b,S}
-    NamedTuple{a,b}(ntuple(Val(length(T.parameters))) do i
+    NamedTuple{a,b}(ntuple(Val(length(a))) do i
         Base.@_inline_meta
         make_zero_immutable!(prev[a[i]], seen)
     end)
--- a/src/make_zero.jl
+++ b/src/make_zero.jl
@@ -384,6 +384,12 @@ end
 
     push!(seen, prev)
 
+    # For make_zero!(NamedTuple) we want to recurse and zero out
+    # the storage
+    if !Base.ismutabletype(T)
+        return Base.make_zero_immutable!(prev, seen)
+    end
+
     for i = 1:nf
         if isdefined(prev, i)
             xi = getfield(prev, i)
--- a/src/make_zero.jl
+++ b/src/make_zero.jl
@@ -387,7 +387,7 @@ end
     # For make_zero!(NamedTuple) we want to recurse and zero out
     # the storage
     if !Base.ismutabletype(T)
-        return Base.make_zero_immutable!(prev, seen)
+        return make_zero_immutable!(prev, seen)
     end
 
     for i = 1:nf
--- a/src/make_zero.jl
+++ b/src/make_zero.jl
@@ -387,7 +387,8 @@ end
     # For make_zero!(NamedTuple) we want to recurse and zero out
     # the storage
     if !Base.ismutabletype(T)
-        return make_zero_immutable!(prev, seen)
+        make_zero_immutable!(prev, seen)
+        return nothing
     end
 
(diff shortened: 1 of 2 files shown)


danielwe:
Hi, just want to mention that I've spent a lot of time with make_zero! lately for #1852, which is almost ready. There are several bugs (missing branches in some `make_zero_immutable!` methods, several `in(seen, prev)` that should be `in(prev, seen)`, et cetera), but I'm not sure a separate `!ismutable` branch in the generic method is needed. These cases should go through line 404. What's the error you're fixing?

You can check out this commit on a branch in my fork for fixes to all the bugs I found in `make_zero` and `make_zero!`: https://github.com/danielwe/Enzyme.jl/commit/7a6ca9ff332b413c80a2cc540a92c79eb026f2eb. I haven't PRd it because it's redundant with #1852, but maybe you'll find the fix you need there?

vchuravy:
> What's the error you're seeing?

If `prev` is an immutable the line 
```
setfield!(prev, i, make_zero_immutable!(xi, seen))
```

in the generic handler will fail.

danielwe:
> If `prev` is an immutable the line
> 
> ```
> setfield!(prev, i, make_zero_immutable!(xi, seen))
> ```

That's only the case if the immutable contains immutable differentiable values, like `(; a=1.0, b=[1.0])` (ActiveState or MixedState), in which case `make_zero!` should fail anyway. If `make_zero!` succeeds on an immutable type, that's because every differentiable value is contained within mutable storage, like `(; a=[1.0], b=[1.0])` or the test case in this PR (DupState or, trivially, AnyState). In these cases, you will never hit the `setfield!` branch, you'll hit the branch below that's recursively calling `make_zero!`.

I.e., the test case added in this PR would never hit the `setfield!` branch anyway.

However, if you have a MixedState NamedTuple contained within a mutable container, such as `[(; a=1.0, b=[1.0])]`, then you'll hit the `make_zero_immutable!(::Type{<:NamedTuple}, ...)` method, and this one has a bug where it fails to call back into `make_zero!` for the mutable fields. That's the bug fixed in this part of the PR I linked: https://github.com/danielwe/Enzyme.jl/commit/7a6ca9ff332b413c80a2cc540a92c79eb026f2eb#diff-10924b07e20c4a235afd58dea795108f564d5f535318adb73991e4e57144e09dR211-R223

Is this what you've been running into?

danielwe:
Just realized the proposed fix here could also be a way to mask #1935 (because the bug in `active_reg_inner` makes the recursion take the wrong branch and call `make_zero!` when it should have called `make_zero_immutable!`), but I think the proper fix for that should be #1936 or something similar. The fix proposed here enables silent bugs where `make_zero!` fails to throw an error on unsupported types, while the bug in `active_reg_inner` remains unaddressed and may show up in other contexts.

danielwe:
...and any issue addressed by this PR that aren't caused by #1935 should be fixed by #1961. If that's not the case, please send me an MWE so I can fix that.

wsmoses:
@vchuravy @danielwe is this PR still live or has it been partially covered elsewhere (please forgive the delay, am getting back to this after the big compile time fix)

danielwe:
Pretty sure the underlying issues here were fixed by #1961. The test suite introduced there should cover every failure mode this PR aimed to address. @vchuravy please let me know if you have an MWE that still fails.

That said, I was a bit too confident in the discussion above. The issue was a bit more nuanced and #1961 didn't get it right until after the full combinatorial nested wrapper testset was added, which was after this discussion. (The solution was to use `active_reg_inner` less and `ismutabletype` more systematically throughout the `make_zero!` methods, as seen for example in this section of the diff: https://github.com/EnzymeAD/Enzyme.jl/pull/1961/files#diff-10924b07e20c4a235afd58dea795108f564d5f535318adb73991e4e57144e09dL441-R500)

wsmoses:
okay, I'll close this for now then (and definitely reopen if needed)

vchuravy:
We definilty need one of the PRs that @danielwe is working on.

danielwe:
#1961 was already merged, so all this should work in the latest releases!

The bigger #1852 will be ready soon(TM), but that one shouldn't affect make_zero(!) correctness now.

PR #1967, opened 2024-10-16
Fix Codecov badge

gdalle:
The links in the README were pointing to @wsmoses's GitHub profile, which is not a problem for GitHub but confused CodeCov

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,10 +2,10 @@
 
 [![Stable](https://img.shields.io/badge/docs-stable-blue.svg)](https://enzyme.mit.edu/julia/stable)
 [![Dev](https://img.shields.io/badge/docs-dev-blue.svg)](https://enzyme.mit.edu/julia/dev)
-[![Build Status](https://github.com/wsmoses/Enzyme.jl/workflows/CI/badge.svg)](https://github.com/wsmoses/Enzyme.jl/actions)
-[![Coverage](https://codecov.io/gh/wsmoses/Enzyme.jl/branch/master/graph/badge.svg)](https://codecov.io/gh/wsmoses/Enzyme.jl)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,10 +2,10 @@
 
 [![Stable](https://img.shields.io/badge/docs-stable-blue.svg)](https://enzyme.mit.edu/julia/stable)
 [![Dev](https://img.shields.io/badge/docs-dev-blue.svg)](https://enzyme.mit.edu/julia/dev)
-[![Build Status](https://github.com/wsmoses/Enzyme.jl/workflows/CI/badge.svg)](https://github.com/wsmoses/Enzyme.jl/actions)
-[![Coverage](https://codecov.io/gh/wsmoses/Enzyme.jl/branch/master/graph/badge.svg)](https://codecov.io/gh/wsmoses/Enzyme.jl)
+[![Build Status](https://github.com/EnzymeAD/Enzyme.jl/workflows/CI/badge.svg)](https://github.com/EnzymeAD/Enzyme.jl/actions)
+[![Coverage](https://codecov.io/gh/EnzymeAD/Enzyme.jl/branch/master/graph/badge.svg)](https://codecov.io/gh/EnzymeAD/Enzyme.jl)
```
wsmoses, line 6: the master branch doesn't exist, we use main

wsmoses: I'm also not quite certain how helpful codecov here is tbh, since sadly Enzyme rules and the likes don't count for codeconv

wsmoses: cc @vchuravy

gdalle: I replaced `master` with `main`. As for the rules, I figured that such a metaprogramming-heavy would be tricky to handle. Still, I think it is worth adding because

- It gives you precious information on parts of the code that are not rules
- You can add more granularity afterwards with the settings of the [`julia-processcoverage`](https://github.com/julia-actions/julia-processcoverage) action, for instance by ignoring certain folders.

PR #1979, opened 2024-10-17
Utilities for splitting and unsplitting mode objects

### LINKED ISSUE #1938: Utilities to translate from `ReverseMode` to `ReverseModeSplit`
In DI, the `pullback` functions use split reverse mode by default, so they disregard the `mode` field of `AutoEnzyme` (because I figure no one is going to pass a `ReverseSplitWithPrimal` anyway). But still I'd like to respect some of the attributes of the `ReverseMode` that the user provides.
Would you welcome a PR with functions that translate a `ReverseMode` to the corresponding `ReverseModeSplit`, setting reasonable defaults for the remaining type parameters?

gdalle:
Fix #1938 

I can also add more tests for these mode modifiers in a separate PR, if you're interested


lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -384,7 +384,7 @@ Subtype of [`Mode`](@ref) for split reverse mode differentiation, to use in [`au
     - [`set_abi`](@ref)
     - [`ReverseSplitModified`](@ref), [`ReverseSplitWidth`](@ref)
 """
-struct ReverseModeSplit{ReturnPrimal,ReturnShadow,Width,RuntimeActivity,ModifiedBetween,ABI,Holomorphic,ErrIfFuncWritten,ShadowInit} <: Mode{ABI, ErrIfFuncWritten,RuntimeActivity} end
+struct ReverseModeSplit{ReturnPrimal,ReturnShadow,RuntimeActivity,Width,ModifiedBetween,ABI,Holomorphic,ErrIfFuncWritten,ShadowInit} <: Mode{ABI, ErrIfFuncWritten,RuntimeActivity} end
```
gdalle, line 387: I think the type parameters were switched. It has no concrete effect because the struct might as well be defined as
```julia
struct ReverseModeSplit{T1,T2,T3,...}
```
but fixing it makes reading the code easier

lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -578,4 +578,88 @@ Return a new mode with its [`ABI`](@ref) set to the chosen type.
...
+    ReturnPrimal,
+    RuntimeActivity,
+    ABI,
+    Holomorphic,
+    ErrIfFuncWritten
+}
+    ReturnShadow = true
+    Width = 0
+    ModifiedBetween = true
+    ShadowInit = false
```
gdalle, line 606: I took the default values from `ReverseSplitWithPrimal`

lib/EnzymeCore/src/EnzymeCore.jl
wsmoses, line 387: Probably keep this the same?

lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -578,4 +578,88 @@ Return a new mode with its [`ABI`](@ref) set to the chosen type.
...
+    ErrIfFuncWritten
+}) where {
+    ReturnPrimal,
+    RuntimeActivity,
+    ABI,
+    Holomorphic,
+    ErrIfFuncWritten
+}
+    ReturnShadow = true
+    Width = 0
```
wsmoses, line 604: Can these be (perhaps default) args to this fn?

lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -578,4 +578,88 @@ Return a new mode with its [`ABI`](@ref) set to the chosen type.
...
+"""
+    unsplit_mode(::ReverseMode)
+
+Turn a [`ReverseModeSplit`](@ref) object into a [`ReverseMode`](@ref) object while preserving as many of the settings as possible.
+
+This function acts as the identity on a [`ReverseMode`](@ref).
+
+See also [`split_mode`](@ref).
+"""
+function unsplit_mode(::ReverseModeSplit{
```
wsmoses, line 632: Combined mode is usually what we refer to it as

lib/EnzymeCore/test/mode_modification.jl
```diff
@@ -0,0 +1,15 @@
...
+using Test
+
+@testset "Split / unsplit mode" begin
+    @test split_mode(Reverse) == ReverseSplitNoPrimal
+    @test split_mode(ReverseWithPrimal) == ReverseSplitWithPrimal
+    @test split_mode(ReverseSplitNoPrimal) == ReverseSplitNoPrimal
+    @test split_mode(ReverseSplitWithPrimal) == ReverseSplitWithPrimal
+
+    @test unsplit_mode(Reverse) == Reverse
+    @test unsplit_mode(ReverseWithPrimal) == ReverseWithPrimal
```
wsmoses, line 12: Can you also test the use of runtime activity and err if find written

lib/EnzymeCore/src/EnzymeCore.jl
gdalle: As explained in [this comment](https://github.com/EnzymeAD/Enzyme.jl/pull/1979#discussion_r1805084780), this just fixes a typo that doesn't change functionality but makes the code harder to understand. Semantically, the runtime activity parameter comes before width in all the uses of this struct, even though they are "labelled" wrong here.

lib/EnzymeCore/test/mode_modification.jl
gdalle: done

lib/EnzymeCore/src/EnzymeCore.jl
gdalle: renamed it

lib/EnzymeCore/src/EnzymeCore.jl
gdalle: I added them as positional arguments as `Val` types with default values, to ensure type stability

lib/EnzymeCore/test/mode_modification.jl
```diff
@@ -0,0 +1,23 @@
...
+
+    @test split_mode(Reverse, Val(:ReturnShadow), Val(:Width), Val(:ModifiedBetween), Val(:ShadowInit)) == ReverseModeSplit{false,:ReturnShadow,false,:Width,:ModifiedBetween,EnzymeCore.DefaultABI,false,false,:ShadowInit}()
+
+    @test combined_mode(Reverse) == Reverse
+    @test combined_mode(ReverseWithPrimal) == ReverseWithPrimal
+    @test combined_mode(ReverseSplitNoPrimal) == Reverse
+    @test combined_mode(ReverseSplitWithPrimal) == ReverseWithPrimal
+
+    @test combined_mode(set_runtime_activity(ReverseSplitNoPrimal)) == set_runtime_activity(Reverse)
+    @test combined_mode(set_err_if_func_written(ReverseSplitNoPrimal)) == set_err_if_func_written(Reverse)
```
wsmoses, line 22: Oh sorry one more, if you could also confirm that ABI is preserved too

wsmoses (approved):
One more test then LGTM!

lib/EnzymeCore/test/mode_modification.jl
gdalle: I tried that but `set_abi` isn't defined yet for `ReverseModeSplit`, want me to add it?

wsmoses: Yeah go for it, probably an oversight

gdalle: done

gdalle:
Silly question: do the tests of EnzymeCore ever run? Because that's where I put these new checks

wsmoses:
....good question. Probably we should just add a CI job for EnzymeCore

gdalle:
See https://github.com/EnzymeAD/Enzyme.jl/pull/2009

gdalle:
Since #2009 has been merged by @vchuravy, I have integrated the changes here and I think this one is good to go.

gdalle:
Is anything more needed on my end?

gdalle:
Gentle bump on this one @wsmoses

lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -421,6 +421,8 @@ Return a new instance of [`ReverseModeSplit`](@ref) mode where `Width` is set to
...
 @inline clear_runtime_activity(::ReverseModeSplit{ReturnPrimal,ReturnShadow,RuntimeActivity,Width,ModifiedBetween,ABI, Holomorphic, ErrIfFuncWritten, ShadowInit}) where {ReturnPrimal,ReturnShadow,RuntimeActivity,Width,ModifiedBetween,ABI, Holomorphic, ErrIfFuncWritten, ShadowInit} = ReverseModeSplit{ReturnPrimal,ReturnShadow,false,Width,ModifiedBetween,ABI, Holomorphic, ErrIfFuncWritten, ShadowInit}()
 
+@inline set_abi(::ReverseModeSplit{ReturnPrimal,ReturnShadow,RuntimeActivity,Width,ModifiedBetween,OldABI,Holomorphic,ErrIfFuncWritten,ShadowInit}, ::Type{NewABI}) where {ReturnPrimal,ReturnShadow,RuntimeActivity,Width,ModifiedBetween,OldABI,Holomorphic,ErrIfFuncWritten,ShadowInit,NewABI<:ABI} = ReverseModeSplit{ReturnPrimal,ReturnShadow,RuntimeActivity,Width,ModifiedBetween,NewABI,Holomorphic,ErrIfFuncWritten,ShadowInit}()
```
wsmoses, line 424: just noticed this is also missing a version for set_abi on the type [not just value], otherwise lgtm

wsmoses (approved):

lib/EnzymeCore/src/EnzymeCore.jl
gdalle: I'm not sure what you mean, the signature is
```
set_abi(::Mode, ::Type{A}) where {A<:ABI}
```
and it is implemented consistently for all modes. Do you mean you also want
```
set_abi(::Type{M}, ::Type{A}) where {M<:Mode,A<:ABI}
```
everywhere?

wsmoses: like this https://github.com/EnzymeAD/Enzyme.jl/commit/03d70353863e245ceae1dc04a201575073509be2

wsmoses: you add set_abi methods for split, and it should come with both type and mode

gdalle: done

lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -594,6 +597,100 @@ Return a new mode with its [`ABI`](@ref) set to the chosen type.
...
+"""
+    combined_mode(::ReverseMode)
+
+Turn a [`ReverseModeSplit`](@ref) object into a [`ReverseMode`](@ref) object while preserving as many of the settings as possible.
+
+This function acts as the identity on a [`ReverseMode`](@ref).
+
+See also [`split_mode`](@ref).
+"""
+function combined_mode(
```
wsmoses, line 660: @vchuravy should we have this be lowercase or upper case. I'm actually now thinking maybe uppercase. I like the WithPrimal(Forward) == ForwardWithPrimal property.

wsmoses: so then similarly here Split(ReverseWithPrimal) == ReverseSplitWithPrimal

gdalle: Whenever possible, it would be nice to keep [consistency with Base Julia](https://docs.julialang.org/en/v1/manual/style-guide/#Use-naming-conventions-consistent-with-Julia-base/): uppercase for types, lowercase for functions.
Of course this is already broken by some of the existing mode utilities, and it's too late to go back, but maybe adopting a minimal style guide would help? That way, contributors know right away how to comply, instead of following changes of hearts from maintainers.

gdalle:
Hi @wsmoses @vchuravy, any updates on this? It got two approvals already but then more changes were requested every time.
Sorry to be insistent, it is just important for DI. Some operators use split mode instead of combined mode (e.g. pullback when the output is not a number). At the moment, for lack of translating utilities, DI completely [disregards the mode information](https://github.com/JuliaDiff/DifferentiationInterface.jl/blob/675b4bd8885b5b120f8640e47e90c370fa93f843/DifferentiationInterface/ext/DifferentiationInterfaceEnzymeExt/utils.jl#L63-L66) provided by the user in the backend. So even when they give a mode like `set_runtime_activity(Reverse)`, DI will fall back on the default `ReverseSplitNoPrimal`. It's not ideal but since backend type parameters are internals, the only thing I can do is wait for translators to be part of your API.

wsmoses:
I think we’re still waiting for feedback from Valentin here.

otherwise I personally say change it to be consistent with WithPrimal then merge it

wsmoses:
Re DI, perhaps in the interim you should throw an error in those cases (rather than risk incorrect results)

gdalle:
I can't easily throw an error in these cases. Indeed, detecting these cases would require accessing the mode's type parameters, which are Enzyme internals (that's why we made the mode setters in the first place).

gdalle:
@wsmoses I renamed `split_mode` to `Split` and `combined_mode` to `Combined` as requested. Do you still want to wait for @vchuravy?

vchuravy (approved):
Looks reasonable to me.

gdalle:
@wsmoses is this good to go?

gdalle:
Thanks!

PR #1987, opened 2024-10-19
Update CIs to use `julia-actions/install-juliaup`

prbzrg:
Also replaced the `actions/cache` with `julia-actions/cache`.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -21,8 +21,8 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - 'release'
-          - 'lts'
+          - '1.10'
+          - '1.11'
           - 'nightly'
         os:
           - ubuntu-20.04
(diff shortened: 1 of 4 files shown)


wsmoses:
what's the advantage of juliaup here? Also we should keep the names 1.10, 1.11, etc explicitly

prbzrg:
`juliaup` is the recommended way to install and use Julia.
The benefits are few but IMHO, we should choose it over `setup-julia`.

wsmoses:
…but for CI where versions are always fixed?

imo we should use whatever reduces setup latency since these jobs are run constantly.

cc @vchuravy if you have thoughts here

prbzrg:
I looked into some CI runs, and it seems that `install-juliaup` is the faster setup. (1~2 seconds less)

wsmoses:
If that's the case I really don't see the value add here?

My understanding of juliaup is that essentially it's a script which download's the right julia version binary as required, then dispatches to that. Which, is all good for making users lives' easier.

But in this CI context, we'll never need that functionality, and it feels like it would add extra points of failure/latency, no?

prbzrg:
I don't think it would be any more failures. but feel free to close this one.
If in the future, there were more points, we can open this again.

PR #1989, opened 2024-10-20
1.11: more methodinstance stuff

wsmoses:


src/utils.jl
```diff
@@ -253,24 +253,73 @@ export codegen_world_age
...
-    mi === nothing && throw(MethodError(ft, tt, world))
-    mi = mi::MethodInstance
-
-    # `jl_method_lookup_by_tt` and `jl_method_lookup` can return a unspecialized mi
-    if !Base.isdispatchtuple(mi.specTypes)
-        mi = Core.Compiler.specialize_method(mi.def, sig, mi.sparam_vals)::MethodInstance
+    if Base.isdispatchtuple(sig)   # JuliaLang/julia#52233
+        return GPUCompiler.methodinstance(ft, tt, world)
+    else
+        return prevmethodinstance(ft, tt, world)
```
vchuravy, line 321: You just want to call `mi = Core.Compiler.specialize_method(mi.def, sig, mi.sparam_vals)::MethodInstance` in this case, but you need a call to which to obtain `def`

vchuravy: Edit: Maybe? This is all pretty much unsound.

wsmoses: I mean happy to defer that to the experts, it's now just falling back to the old implementation

wsmoses: Old from 1.10 in gpucompiler*** while concurrently I opened a PR to gpucompiler jl to actually fix and not have us have to vendor it

wsmoses: https://github.com/JuliaGPU/GPUCompiler.jl/pull/644

PR #2009, opened 2024-10-23
Add CI testing for EnzymeCore

gdalle:
As discussed in #1979, this PR adds a CI job for testing EnzymeCore and reporting coverage, following the model of EnzymeTestUtils.
It also skips two tests which seem broken but were never run before so we didn't know.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -188,6 +188,68 @@ jobs:
           files: lcov.info
           token: ${{ secrets.CODECOV_TOKEN }}
           fail_ci_if_error: false  # or true if you want CI to fail when Codecov fails
+  enzymecore:
+    name: EnzymeCore - Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ matrix.libEnzyme }} libEnzyme - ${{ github.event_name }}
+    runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.version == 'nightly' }}
+    env:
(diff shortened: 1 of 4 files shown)


lib/EnzymeCore/test/misc.jl
```diff
@@ -0,0 +1,27 @@
...
+    return Const(f(; kwargs...))
+end
+
+@test has_frule_from_sig(Base.signature_type(f, Tuple{}))
+
+data = [1.0, 2.0, 3.0, 4.0]
+
+d = @view data[2:end]
+y = @view data[3:end]
+@test_skip @test_throws AssertionError Duplicated(d, y)
```
gdalle, line 24: this failed

lib/EnzymeCore/test/misc.jl
```diff
@@ -0,0 +1,27 @@
...
+
+@test_throws ErrorException Active(data)
+@test_skip@test_throws ErrorException Active(d)
```
gdalle, line 27: this also failed

vchuravy (approved):

.github/workflows/CI.yml
```diff
@@ -188,6 +188,68 @@ jobs:
...
+    runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.version == 'nightly' }}
+    env:
+      JULIA_PROJECT: "lib/EnzymeCore"
+    strategy:
+      fail-fast: false
+      matrix:
+        version:
+          - '1.10'
+          - ~1.11.0-0
```
wsmoses, line 202: This should just be 1.11

PR #2021, opened 2024-10-28
Import `needs_primal` from Reactant.jl

mofeing:
`needs_primal(Mode)` returns `true` if the mode needs to compute the primal.

CC @wsmoses


wsmoses (approved):
lgtm but can you bump the version toml for EnzymeCore? [and add compat for Enzyme.jl itself]

test/runtests.jl
```diff
@@ -3634,6 +3634,15 @@ end
     @test NoPrimal(ReverseSplitWithPrimal) === ReverseSplitNoPrimal
 end
 
+@testset "needs_primal" begin
+    @test needs_primal(Reverse) === false
```
wsmoses, line 3638: Tests fail here, presumably this should be moved into the EnzymeCore tests [not enzyme proper tests]

mofeing: I can do that, but I wrote it here because other parts of EnzymeCore are actually tested here and not in EnzymeCore's tests (e.g. `WithPrimal`).

Also, it should also be solved by exporting `needs_primal`.

wsmoses: yeah we didn't use to have CI for EnzymeCore separately, but since we now do, might as well put in the right place (and ideally move the rest)

Project.toml
```diff
@@ -35,7 +35,7 @@ EnzymeStaticArraysExt = "StaticArrays"
 BFloat16s = "0.2, 0.3, 0.4, 0.5"
 CEnum = "0.4, 0.5"
 ChainRulesCore = "1"
-EnzymeCore = "0.8.4"
```
wsmoses, line 38: can the compat be set to either 0.8.4 or 0.8.5 ?

mofeing: do you mean to support both versions? we can put the `~` specifier in the version

wsmoses: yup so "0.8.4, 0.8.5"

Issue #2036, opened 2024-10-31
Segmentation fault, Bijectors.RadialLayer


mhauru:
MWE:
```julia
module MWE
import Bijectors, Enzyme, StableRNGs
rng = StableRNGs.StableRNG(23)

function wrap()
   b = Bijectors.RadialLayer(3)
   binv = Bijectors.inverse(b)
   f(x) = sum(b(binv(x)))
   x = randn(rng, 3)
   Enzyme.gradient(Enzyme.Forward, Enzyme.Const(f), x)
   Enzyme.gradient(Enzyme.Reverse, Enzyme.Const(f), x)
end

wrap()
end
```

Output:
```julia
┌ Warning: Using fallback BLAS replacements for (["dasum_64_"]), performance may be degraded
└ @ Enzyme.Compiler ~/.julia/packages/GPUCompiler/GnbhK/src/utils.jl:59
┌ Warning: Using fallback BLAS replacements for (["dasum_64_"]), performance may be degraded
└ @ Enzyme.Compiler ~/.julia/packages/GPUCompiler/GnbhK/src/utils.jl:59

[81422] signal (11.2): Segmentation fault: 11
in expression starting at REPL[1]:14
getindex at ./essentials.jl:13 [inlined]
macro expansion at ./reduce.jl:264 [inlined]
macro expansion at ./simdloop.jl:77 [inlined]
mapreduce_impl at ./reduce.jl:263
macro expansion at /Users/mhauru/projects/Enzyme.jl/src/compiler.jl:8163 [inlined]
enzyme_call at /Users/mhauru/projects/Enzyme.jl/src/compiler.jl:7729 [inlined]
CombinedAdjointThunk at /Users/mhauru/projects/Enzyme.jl/src/compiler.jl:7502 [inlined]
autodiff at /Users/mhauru/projects/Enzyme.jl/src/Enzyme.jl:491 [inlined]
macro expansion at /Users/mhauru/projects/Enzyme.jl/src/Enzyme.jl:1720 [inlined]
gradient at /Users/mhauru/projects/Enzyme.jl/src/Enzyme.jl:1661 [inlined]
wrap at ./REPL[1]:11
unknown function (ip: 0x310d94193)
ijl_apply_generic at /Users/mhauru/.julia/juliaup/julia-1.10.5+0.aarch64.apple.darwin14/lib/julia/libjulia-internal.1.10.5.dylib (unknown line)
[...]

Issue #2035, opened 2024-10-31
Illegal type analysis, Bijectors.PlanarLayer


mhauru:
MWE:
```julia
module MWE
import Bijectors, Enzyme, StableRNGs
b = Bijectors.PlanarLayer(3)
binv = Bijectors.inverse(b)
x = randn(StableRNGs.StableRNG(23), (3, 3))
f = x -> sum(b(binv(x)))
Enzyme.gradient(Enzyme.Forward, f, x)
end
```

Output:
```
ERROR: Enzyme compilation failed due to illegal type analysis.
Current scope:
; Function Attrs: mustprogress willreturn
define internal fastcc void @preprocess_julia_init_state_16477([4 x double]* noalias nocapture noundef nonnull sret([4 x double]) align 8 dereferenceable(32) "enzyme_type"="{[-1]:Pointer, [-1,-1]:Float@double}" %0, { [3 x double] } addrspace(11)* nocapture noundef nonnull readonly align 8 dereferenceable(24) "enzyme_type"="{[-1]:Pointer, [-1,0]:Float@double, [-1,8]:Float@double, [-1,16]:Float@double}" "enzymejl_parmtype"="4457150416" "enzymejl_parmtype_ref"="1" %1, double "enzyme_type"="{[-1]:Float@double}" "enzymejl_parmtype"="4770454480" "enzymejl_parmtype_ref"="0" %2, double "enzyme_type"="{[-1]:Float@double}" "enzymejl_parmtype"="4770454480" "enzymejl_parmtype_ref"="0" %3, double "enzyme_type"="{[-1]:Float@double}" "enzymejl_parmtype"="4770454480" "enzymejl_parmtype_ref"="0" %4, double "enzyme_type"="{[-1]:Float@double}" "enzymejl_parmtype"="4770454480" "enzymejl_parmtype_ref"="0" %5) unnamed_addr #59 !dbg !5084 {
top:
  %6 = alloca [4 x double], align 8
  %7 = call {}*** @julia.get_pgcstack() #60
  %ptls_field14 = getelementptr inbounds {}**, {}*** %7, i64 2
  %8 = bitcast {}*** %ptls_field14 to i64***
  %ptls_load1516 = load i64**, i64*** %8, align 8, !tbaa !61
  %9 = getelementptr inbounds i64*, i64** %ptls_load1516, i64 2
  %safepoint = load i64*, i64** %9, align 8, !tbaa !65
[...]

Issue #2034, opened 2024-10-31
Julia crash, illegal Instruction 4, Bijectors PDBijector


mhauru:
MWE:
```julia
module BijectorsIntegrationTests

using Bijectors: Bijectors
using Enzyme: Enzyme
using StableRNGs: StableRNG

rng = StableRNG(23)

function test_grad()
    dim = (3, 3)
    bijector = Bijectors.PDBijector()
    b_inv = Bijectors.inverse(bijector)
    f = x -> sum(bijector(b_inv(x)))
    x = randn(rng, dim)
    Enzyme.gradient(Enzyme.Reverse, f, x)
    return nothing
end

test_grad()

end
```

The fact that this is wrapped in the function `test_grad` is significant. Executed directly in the module body the same code passes without issue.

Output:
```
WARNING: replacing module BijectorsIntegrationTests.

[58771] signal (4.1): Illegal instruction: 4
in expression starting at /Users/mhauru/projects/Enzyme.jl/test/integration/mwe.jl:19
_chol! at /Users/julia/.julia/scratchspaces/a66863c6-20e8-4ff4-8a62-49f30b1f605e/agent-cache/default-honeycrisp-R17H3W25T9.0/build/default-honeycrisp-R17H3W25T9-0/julialang/julia-release-1-dot-10/usr/share/julia/stdlib/v1.10/LinearAlgebra/src/cholesky.jl:191 [inlined]
#cholesky!#140 at /Users/julia/.julia/scratchspaces/a66863c6-20e8-4ff4-8a62-49f30b1f605e/agent-cache/default-honeycrisp-R17H3W25T9.0/build/default-honeycrisp-R17H3W25T9-0/julialang/julia-release-1-dot-10/usr/share/julia/stdlib/v1.10/LinearAlgebra/src/cholesky.jl:268 [inlined]
cholesky! at /Users/julia/.julia/scratchspaces/a66863c6-20e8-4ff4-8a62-49f30b1f605e/agent-cache/default-honeycrisp-R17H3W25T9.0/build/default-honeycrisp-R17H3W25T9-0/julialang/julia-release-1-dot-10/usr/share/julia/stdlib/v1.10/LinearAlgebra/src/cholesky.jl:267 [inlined]
[...]

Issue #2033, opened 2024-10-31
Bijectors, reverse mode hangs


mhauru:
MWE:
```julia
module MWE
import Bijectors, Enzyme
b = Bijectors.CorrBijector()
binv = Bijectors.inverse(b)
f(x) = sum(b(binv(x)))
x = randn((0, 0))
Enzyme.gradient(Enzyme.Reverse, f, x)
end
```

Output:
```
WARNING: replacing module MWE.
 ** On entry to DTRMM  parameter number  9 had an illegal value
 ** On entry to DLASCL parameter number  9 had an illegal value
 ** On entry to DTRSM  parameter number  9 had an illegal value
 ** On entry to DTRSM  parameter number  9 had an illegal value
```

It seems to just hang after that. I ran out of patience after 5 minutes.

Enzyme v0.13.12.

wsmoses:
Okay this one requires minimization and isn't resolved by current dev branches

wsmoses:
@penelopeysm and also would you be able to help reduce this one too?

mhauru:
Giving this a shot now

mhauru:
It's an issue with Cholesky on a zero-dim matrix:
```julia
module MWE
import Enzyme, LinearAlgebra

function cholesky!(A)
    C, info = LinearAlgebra._chol!(A, LinearAlgebra.UpperTriangular)
    return C
end

f(x) = sum(cholesky!(copy(x)))
x = zeros((0, 0))
Enzyme.gradient(Enzyme.Reverse, f, x)
end
```

wsmoses:
fixed on main

PR #2037, opened 2024-10-31
Add integration tests for Bijectors

mhauru:
There have been some regressions that cause various crashes and errors with Bijectors.jl and Enzyme v0.13. To guard against this in the future, I think it would be useful to introduce some integration tests for Bijectors, like we are doing in #1813 and #1819 for Turing.jl and Distributions.jl.

I've added a basic check of running forward and reverse mode on all bijector types with at least one argument, plus some extra ones that have caused trouble in the past. I've marked all failing ones, and reported them as separate issues. Here's a list:
* https://github.com/EnzymeAD/Enzyme.jl/issues/2035
* https://github.com/EnzymeAD/Enzyme.jl/issues/2030
* https://github.com/EnzymeAD/Enzyme.jl/issues/2033
* https://github.com/EnzymeAD/Enzyme.jl/issues/2034
* https://github.com/EnzymeAD/Enzyme.jl/issues/2036
* https://github.com/EnzymeAD/Enzyme.jl/issues/2029
* https://github.com/EnzymeAD/Enzyme.jl/issues/2041

cc @yebai


test/integration/Bijectors/runtests.jl
```diff
@@ -0,0 +1,222 @@
...
+    r_mode = if (case.runtime_activity === Both || case.runtime_activity === Reverse)
+        Enzyme.set_runtime_activity(Enzyme.Reverse)
+    else
+        Enzyme.Reverse
+    end
+
+    if !(case.skip === Forward) && !(case.skip === Both)
+        if case.broken === Both || case.broken === Forward
+            @test_broken(
+                collect(Enzyme.gradient(f_mode, Enzyme.Const(f), x...)[1]) ≈ finitediff,
```
wsmoses, line 77: why are these collect's needed?

mhauru: They are not. Removed. Thanks for spotting.

mhauru:
Had to mark one more test as broken, CI now passes. As Hong said, I won't have time to minimise the issues I opened in the near future. Would you be happy to merge as is? Should be better than nothing, and we can add in more tests as we close the individual issues.

wsmoses:
I think the segfaults need further reductions at least, lest we not be able to mark them as broken

mhauru:
I added back in a bunch of tests that were unblocked by the recent fixes. Now the only one marked with `skip=Reverse` is the one that hangs.

mhauru:
There's now a new seg fault I'm unable to reproduce locally. @wsmoses (or anyone else), would you have a Linux machine to try it on? I'm on Mac, CI runs on Ubuntu 22.04.

test/integration/Bijectors/runtests.jl
```diff
@@ -0,0 +1,210 @@
...
+end
+
+@testset "Bijectors integration tests" begin
+    test_cases = TestCase[
+        sum_b_binv_test_case(Bijectors.VecCorrBijector(), 3),
+        sum_b_binv_test_case(Bijectors.VecCorrBijector(), 0),
+        sum_b_binv_test_case(Bijectors.CorrBijector(), (3, 3)),
+        # TODO(mhauru) Skip Reverse because of
+        # https://github.com/EnzymeAD/Enzyme.jl/issues/2033
+        sum_b_binv_test_case(Bijectors.CorrBijector(), (0, 0); skip=Reverse),
```
wsmoses, line 130: this should no longer be skipped

test/integration/Bijectors/runtests.jl
```diff
@@ -0,0 +1,210 @@
...
+        sum_b_binv_test_case(
+            Bijectors.Permute([
+                0 1 0;
+                1 0 0;
+                0 0 1
+            ]),
+            (3, 3),
+        ),
+        # TODO(mhauru) Both modes broken because of
+        # https://github.com/EnzymeAD/Enzyme.jl/issues/2035
```
wsmoses, line 150: So this issue is essentially due to Enzyme 0.13 not being supported by the custom rule on bijectors on all uses of roots.jl to define rules.

x/ref: https://github.com/TuringLang/Bijectors.jl/issues/339

I think this needs to be fixed in bijectors.jl before we can take this in here

test/integration/Bijectors/runtests.jl
mhauru: Done.

test/integration/Bijectors/runtests.jl
mhauru: I would propose merging this as is while we fix this, to get the vast majority of the tests in this PR into CI. Bijectors is pinned to a specific version, so once the necessary changes have been made in Bijectors we can release a new version of IT, and then make a PR that marks this unbroken as it bumps the Bijectors version of these tests.

mhauru:
Tests pass and nothing is marked as `skip`ped any more. Happy to merge @wsmoses?

PR #2079, opened 2024-11-11
Improve broadcast index analysis

wsmoses:


vchuravy:
I didn't see this code go in, but I am not a fan of override like this since this assumes that the implementation underneath is not changing. 

`Base.copyto!` is often replaced by custom broadcast implementations. So I would be worried that we are replacing a custom user implementation with our implementations, I see the safe guards in abstract_call_known, but this feels like a very fragile cludge.

PR #2100, opened 2024-11-17
Handle case for SparseMatrixCSC when output is Const

### LINKED ISSUE #2099: Error in `autodiff` involving `mul!` of `SparseMatrixCSC`
The following example doesn't work

```julia
using LinearAlgebra
using SparseArrays
using Enzyme

# Assume `dudt!` is defined as:
function dudt!(du, u, p, t)
  mul!(du, A, u, p[1], p[2])  # In-place multiplication
  return sum(du)             # Return scalar sum
end

# Example inputs
# const A = rand(5, 5)
const A = sprand(5, 5, 0.4)
du = zeros(5)
u = rand(5)
p = [1.0, 0.5]  # Active parameters
dp = similar(p)
t = 0.0         # Constant parameter

# Define Enzyme gradient computation
grad = Enzyme.autodiff(
  Reverse,
  dudt!,                # Function to differentiate
  Const(du),            # Do not differentiate with respect to `du`
  Const(u),             # `u` is treated as a constant
  Duplicated(p, dp),    # Differentiate with respect to `p`
  Const(t)              # `t` is treated as a constant
)
```

and returns the error

```julia
ERROR: UndefVarError: `dα` not defined in local scope
Suggestion: check for an assignment to a local variable that shadows a global of the same name.
Stacktrace:
  [1] reverse
    @ ~/.julia/packages/Enzyme/azJki/src/internal_rules.jl:849 [inlined]
  [2] dudt!
    @ ~/GitHub/Research/Undef/Autodiff QuantumToolbox/autodiff.jl:93 [inlined]
  [3] diffejulia_dudt__19520wrap
    @ ~/GitHub/Research/Undef/Autodiff QuantumToolbox/autodiff.jl:0
  [4] macro expansion
    @ ~/.julia/packages/Enzyme/azJki/src/compiler.jl:8369 [inlined]
  [5] enzyme_call
    @ ~/.julia/packages/Enzyme/azJki/src/compiler.jl:7932 [inlined]
  [6] CombinedAdjointThunk
    @ ~/.julia/packages/Enzyme/azJki/src/compiler.jl:7705 [inlined]
  [7] autodiff
    @ ~/.julia/packages/Enzyme/azJki/src/Enzyme.jl:491 [inlined]
  [8] autodiff
    @ ~/.julia/packages/Enzyme/azJki/src/Enzyme.jl:537 [inlined]
  [9] autodiff(::ReverseMode{…}, ::typeof(dudt!), ::Const{…}, ::Const{…}, ::Duplicated{…}, ::Const{…})
    @ Enzyme ~/.julia/packages/Enzyme/azJki/src/Enzyme.jl:504
 [10] top-level scope
    @ ~/GitHub/Research/Undef/Autodiff QuantumToolbox/autodiff.jl:108
```

ptiede:
This fixes #2099


ptiede:
@wsmoses, do I need to handle the BatchDuplicated case when the output is Const?

wsmoses:
sorry, in the case that what is batchduplicated?

ptiede:
I don't think there is one. So I should be fine just doing 

```julia
        dα = zero(α.val)
        dβ = zero(β.val)
```
right?

ptiede:
Ok that should be handled correctly again. I think this is good to go.

Issue #2074, opened 2024-11-08
Segfault when calling `rand` with `Bijectors`


Red-Portal:
Hi!

The following code segfaults on 1.10:

```julia
using Bijectors
using LinearAlgebra
using Functors
using Optimisers
using Enzyme
using Random, StableRNGs

struct TestProb1 end

logdensity(::TestProb1, θ) = sum(abs2, θ)

function Bijectors.bijector(::TestProb1)
    return Bijectors.Stacked(
        [Base.Fix1(broadcast, log), identity],
        [1:1, 2:3],
    )
end

struct TestProb2 end

logdensity(::TestProb2, θ) = sum(abs2, θ)

struct MvLocationScale{S,D<:ContinuousDistribution,L,E<:Real} <:
       ContinuousMultivariateDistribution
    location::L
    scale::S
    dist::D
    scale_eps::E
end

Base.length(q::MvLocationScale) = length(q.location)

Functors.@functor MvLocationScale (location, scale)

# This specialization improves AD performance of the sampling path
function Distributions.rand(
    rng::AbstractRNG, q::MvLocationScale{<:Diagonal,L}, num_samples::Int
) where {L}
    (; location, scale) = q
    n_dims = length(location)
    scale_diag = diag(scale)
    return scale_diag .* randn(rng, n_dims, num_samples) .+ location
end

function restructure_ad_forward(restructure, params)
    return restructure(params)::typeof(restructure.model)
end

function estimate_repgradelbo_ad_forward(params′, aux)
    (; rng, problem, restructure) = aux
    q = restructure_ad_forward(restructure, params′)
    zs = rand(rng, q, 10)
    return mean(Base.Fix1(logdensity, problem), eachcol(zs))
end

function main()
    d = 5

    seed = (0x38bef07cf9cc549d)
    rng = StableRNG(seed)

    for prob in [TestProb1(), TestProb2()]
        q = if prob isa TestProb1
            MvLocationScale(zeros(d), Diagonal(ones(d)), Normal(), 1e-5)
        else
            Bijectors.TransformedDistribution(
                MvLocationScale(zeros(d), Diagonal(ones(d)), Normal(), 1e-5),
                inverse(
                    Bijectors.Stacked(
                        [Base.Fix1(broadcast, log), identity],
                        [1:1, 2:d],
                    )
                )
            )
        end

        params, re = Optimisers.destructure(q)
        buf = zero(params)
[...]

PR #2117, opened 2024-11-23
Fix segfault on return type

wsmoses:
Fixes https://github.com/EnzymeAD/Enzyme.jl/issues/2074


lib/EnzymeCore/src/rules.jl
```diff
@@ -208,18 +208,36 @@ function isapplicable(@nospecialize(f), @nospecialize(TT);
...
+        if caller isa Core.MethodInstance
+            add_mt_backedge!(caller, mt, sig)
+        elseif caller isa Core.Compiler.MethodLookupResult
+            for j = 1:Core.Compiler.length(caller)
+                cmatch = Core.Compiler.getindex(caller, j)::Core.MethodMatch
+                cspec = Core.Compiler.specialize_method(cmatch)::Core.MethodInstance
+                add_mt_backedge!(cspec, mt, sig)
+            end
+        end
     end
```
vchuravy, line 221: Can you explain why this code became necessary?

wsmoses: Yeah so basically it was required to fix a bug in the rules tests.

in particular if a function was defined first without a rule, absint would succeed and let’s say inline the function. The only other time we add a back edge is when we take the actual generated llvm functions, which only exist for non Julia inlined code. So we would never add a back edge for the rule currently without a rule.

we would then go and add a rule, but that wouldn’t cause an invalidation and thus end up with the wrong code

src/compiler/interpreter.jl
```diff
@@ -214,21 +214,27 @@ function Core.Compiler.abstract_call_gf_by_type(
     callinfo = ret.info
     method_table = Core.Compiler.method_table(interp)
     specTypes = simplify_kw(atype)
+    caller = if callinfo isa Core.Compiler.MethodMatchInfo && callinfo.results isa Core.Compiler.MethodLookupResult
```
vchuravy, line 217: This seems wrong. We want to recompile the method one-above in the call-graph not the primal method from whence we came?

wsmoses: I don’t follow? If a rule is defined for f we want in invalidation edge exist from inactive(f) -> f, right?

wsmoses: For example if f defines a custom rule, we need to invalidate autodiff(f) to now just call that rule

lib/EnzymeCore/src/rules.jl
vchuravy: Yeah, that was known as a bug...

Julia solves this, by having a new method definition invalidate the method-table. We could change our rule definitions to a macro form that does that.

PR #2118, opened 2024-11-23
Add one-arg method `Duplicated(x)`

mcabbott:
Instead of writing
```
dx = make_zero(x)
xdx = Duplicated(x, dx)
```
this lets you write
```
xdx = Duplicated(x)
```
and later get `dx = xdx.dval` out when you need it.

### DIFF
--- a/lib/EnzymeCore/src/EnzymeCore.jl
+++ b/lib/EnzymeCore/src/EnzymeCore.jl
@@ -55,12 +55,13 @@ Active(i::Integer) = Active(float(i))
 Active(ci::Complex{T}) where T <: Integer = Active(float(ci))
 
 """
-    Duplicated(x, ∂f_∂x)
+    Duplicated(x, ∂f_∂x = make_zero(x))
 
 Mark a function argument `x` of [`autodiff`](@ref Enzyme.autodiff) as duplicated, Enzyme will
 auto-differentiate in respect to such arguments, with `dx` acting as an
 accumulator for gradients (so ``\\partial f / \\partial x`` will be *added to*)
 `∂f_∂x`.
+If the second argument is not provided, it is created by [`make_zero`](@ref).
 """
 struct Duplicated{T} <: Annotation{T}
     val::T
@@ -75,6 +76,8 @@ struct Duplicated{T} <: Annotation{T}
         new{T1}(x, dx)
     end
 end
+
+Duplicated(x) = Duplicated(x, make_zero(x), false)
 
 """
     DuplicatedNoNeed(x, ∂f_∂x)
--- a/lib/EnzymeCore/test/misc.jl
+++ b/lib/EnzymeCore/test/misc.jl
@@ -23,5 +23,8 @@ d = @view data[2:end]
 y = @view data[3:end]
 @test_skip @test_throws AssertionError Duplicated(d, y)
 
+dup = Duplicated(data)
+@test dup isa Duplicated
+
 @test_throws ErrorException Active(data)
 @test_skip @test_throws ErrorException Active(d)


vchuravy (approved):

MasonProtter:
My only concern about this is that it's rather reverse-mode centric. On the other hand, since there's no universally obvious candidate for what to put in there for forward mode, I guess it's fine to make this method favour reverse mode?

MasonProtter:
Yeah, I don't see any obvious use for it for enzyme so it's probably fine.

mcabbott:
Oh no, now I'm worried this PR will turn into a punching bag for API / documentation complaints. 

From a human UI perspective, it does seem a bit odd to re-use the exact same `struct` for forward & reverse. Especially since the forward mode case where you might plausibly construct `Dual(pi, 1)` with your bare hands is precisely the case where `Duplicated` is useless, perhaps even misleading, in reverse mode:
```julia
julia> xdx = Duplicated(3.0, 100.0);

julia> autodiff(Forward, abs2, xdx)  # ok, like dual numbers
(600.0,)

julia> autodiff(Reverse, abs2, Active, xdx)  # maybe this should be an error?
((nothing,),)

julia> xdx  # Duplicated{Float64} is never a useful container for reverse mode
Duplicated{Float64}(3.0, 100.0)

julia> y = (array = [1.0, 2.0], float = 3.0);

julia> ydy = Duplicated(y, make_zero(y));

julia> autodiff(Reverse, y -> sum(y.array .* y.float), Active, ydy) 
((nothing,),)

julia> ydy.dval  # this isn't useless, but does require some understanding
(array = [3.0, 3.0], float = 0.0)
```
You need `ydy = MixedDuplicated(y, Ref(make_zero(y)))` to match the result of the friendly `gradient(Reverse, y -> sum(y.array .* y.float), y)` here. 

Anyway, this PR changes nothing at all about forward mode. 

Please make a separate issue to discuss changing from Duplicated to some new Dual for forward mode. (Or to discuss more clearly documenting things.)

The one thing that is on-topic for this PR is that we could make `Duplicated(x)` give a helpful error on `x::Float64`, maybe all `isbits(x)` or something. That's non-breaking, and would improve friendliness compared to the present methoderror.  `Duplicated(1.0, 0.0)` is entirely useless for reverse mode (you may as well make `Const(1.0)`) as far as I can tell. And should occur only accidentally in forward mode (i.e. you get the dx from somewhere which could plausibly give you a nonzero value) hence you never want the one-argument constructor.

wsmoses:
I'm very down to make/enforce the use of a separate Dual type for forward mode (which the Rust Enzyme stuff does), but yeah that's separate from here, if you want to open a different issue/PR on that.

Obviously that would be quite breaking [and need corresponding checks throughout enzyme]. 

Part of the reason for the one duplicated for both is that from the implementation (as opposed to the user side) they are very much the same, in the sense that a separate shadow data structure is created and maintained (hence the very literal name "duplicated" as in we duplicated the data structure from primal to shadow)

MasonProtter:
> Oh no, now I'm worried this PR will turn into a punching bag for API / documentation complaints.

> Anyway, this PR changes nothing at all about forward mode.

> Please make a separate issue to discuss changing from Duplicated to some new Dual for forward mode. (Or to discuss more clearly documenting things.)

I don't think that's a fair response. Like it or not, `Duplicated` **is** a fundamental part of the API for Forward mode, so IMO it is quite relevant to talk about how changes to `Duplicated` can make the (already highly confusing and second-class) Forward mode API feel even more confusing and second-class. 

If we did disentangle the APIs so Forward mode didn't use `Duplicated` then I think this'd be a valid dismissal, but since they share it I think it's valid to worry about this making the API even more confusing / hard to teach.

PR #2131, opened 2024-11-28
Fewer anonymous funcs

wsmoses:


src/utils.jl
```diff
@@ -5,12 +5,40 @@
...
-@inline unsafe_to_pointer(val::Type{T}) where {T} = ccall(
-    Base.@cfunction(Base.identity, Ptr{Cvoid}, (Ptr{Cvoid},)),
+@inline Base.@nospecializeinfer unsafe_to_pointer(@nospecialize(val::Type)) = @static if sizeof(Int) == sizeof(Int64)
+    Base.llvmcall((
+"""
+declare nonnull {}* @julia.pointer_from_objref({} addrspace(11)*)
+
+define i64 @f({} addrspace(10)* %obj) readnone alwaysinline {
+  %c = addrspacecast {} addrspace(10)* %obj to {} addrspace(11)*
+  %r = call {}* @julia.pointer_from_objref({} addrspace(11)* %c)
```
vchuravy, line 15: Huh? Doesn't this break code?

src/jlrt.jl
```diff
@@ -1,6 +1,6 @@
 # For julia runtime function emission
     
-function emit_allocobj!(
+Base.@nospecializeinfer function emit_allocobj!(
```
vchuravy, line 3: Can we `import Base: @nospecializeinfer`?

src/compiler/validation.jl
```diff
@@ -489,7 +493,11 @@ function check_ir!(@nospecialize(job::CompilerJob), errors::Vector{IRError}, imp
                     newf, _ = get_function!(mod, fname, FT)
                 else
                     found = nothing
-                    for lbb in blocks(initfn), linst in collect(instructions(lbb))
+                    for lbb in blocks(initfn)
+                    liter = LLVM.API.LLVMGetFirstInstruction(lbb)
+                    while liter != C_NULL
```
vchuravy, line 498: Why not simply `for linst in instructions(lbb)`?

src/utils.jl
wsmoses: surprisingly no (I mean this is what the implementation is supposed to do anyways) and now we don't have a cfunction ptr we have to ccall

src/compiler/validation.jl
wsmoses: hm will look at, elsewhere it is required because we may delete the present instruction (so this is akin to the pre increment iterator in llvm)

src/compiler/utils.jl
```diff
@@ -360,6 +360,46 @@ function reinsert_gcmarker!(func::LLVM.Function, @nospecialize(PB::Union{Nothing
     end
 end
 
+@inline enum_attr_kind(kind::String) = LLVM.API.LLVMGetEnumAttributeKindForName(kind, Csize_t(length(kind)))
+
+const swiftself_kind = enum_attr_kind("swiftself")
+
+Base.@assume_effects :removable :foldable :nothrow function has_swiftself(fn::LLVM.Function)::Bool
```
vchuravy, line 367: Why those effects? We shouldn't add them to every function, and I kinda doubt they apply here or are useful?

src/utils.jl
```diff
@@ -81,7 +81,7 @@ end
 export unsafe_to_ptr
 
 # This mimicks literal_pointer_val / literal_pointer_val_slot
-function unsafe_to_llvm(B::LLVM.IRBuilder, @nospecialize(val))::LLVM.Value
+Base.@nospecializeinfer function unsafe_to_llvm(B::LLVM.IRBuilder, @nospecialize(val))::LLVM.Value
```
vchuravy, line 84: Do all these `@nospecializeinfer` actually do anything?

src/utils.jl
vchuravy: We should just use `Base.unsafe_pointer_to_objref`

wsmoses: There was some reason historically why that wasn’t possible, eg some types saying that wasn’t valid for

vchuravy: I just noticed that my brain read something correctly

```
unsafe_pointer_to_objref(x::Ptr) = ccall(:jl_value_ptr, Any, (Ptr{Cvoid},), x

function pointer_from_objref(@nospecialize(x))
    @inline
    ismutable(x) || error("pointer_from_objref cannot be used on immutable objects")
    ccall(:jl_value_ptr, Ptr{Cvoid}, (Any,), x)
en
```

So `unsafe_pointer_from_objref(x) = ccall(:jl_value_ptr, Ptr{Cvoid}, (Any,), x)`

PR #2141, opened 2024-11-29
Cache inference

wsmoses:


src/compiler/interpreter.jl
```diff
@@ -55,7 +74,8 @@ function EnzymeInterpreter(
     reverse_rules::Bool,
     deferred_lower::Bool = true,
     broadcast_rewrite::Bool = true,
-    handler = nothing
+    handler = nothing,
+    local_cache = get_or_create_inference_cache(world, forward_rules, reverse_rules, deferred_lower, broadcast_rewrite, handler)
```
vchuravy, line 78: If I understand correctly, the local_cache ought to be ephermal and can't be shared across multiple abstract interpreter instances.

cc: @aviatesk

src/compiler/interpreter.jl
```diff
@@ -202,15 +223,7 @@ function Core.Compiler.abstract_call_gf_by_type(
...
-    ret = @invoke Core.Compiler.abstract_call_gf_by_type(
-        interp::AbstractInterpreter,
-        f::Any,
-        arginfo::ArgInfo,
-        si::StmtInfo,
-        atype::Any,
-        sv::AbsIntState,
-        max_methods::Int,
-    )
+    ret = Core.invoke(Core.Compiler.abstract_call_gf_by_type, Tuple{AbstractInterpreter, Any, ArgInfo, StmtInfo, Any, AbsIntState, Int}, interp, f, arginfo, si, atype, sv, max_methods)
```
vchuravy, line 226: Why this change?

vchuravy (changes requested):
Unsure that this is correct.

src/compiler/interpreter.jl
wsmoses: Yeah but in this case we actually are effectively caching the absint itself (since all data about it is in the key)

src/compiler/interpreter.jl
wsmoses: We were spending a huge amount of runtime on that function (which appears to compile as jl_invoke). Tried playing around with to fix to no avail

wsmoses:
I spoke with @aviatesk and I think this is fine, but also sadly this did not resolve the issue”we spend like 50 mins in absint” problems

src/compiler/interpreter.jl
vchuravy: Yeah this change doesn't do anything.

src/compiler/interpreter.jl
vchuravy: All caches that ought to survive the absint go into the global cache?

PR #2148, opened 2024-11-30
Don't call specialize_method again

vchuravy:


wsmoses (approved):

wsmoses:
MethodError: no method matching typeinf_type(::Enzyme.Compiler.Interpreter.EnzymeInterpreter{Nothing}, ::Core.MethodInstance)
  
  Closest candidates are:
    typeinf_type(::Core.Compiler.AbstractInterpreter, ::Method, ::Any, ::Core.SimpleVector)
     @ Core compiler/typeinfer.jl:1058
     
     
@vchuravy is there a diff API?

PR #2153, opened 2024-12-01
Add no recur to inference

wsmoses:


vchuravy (commented):
First I ever see this code, but yeah this looks right.

You likely want to do the same song and dance I did instead of directly calling `_return_type` so that you get the correct invalidation tracking.

wsmoses:
these will never be redefined, so I don't think we need the invalidations here [just purely us overloading something for performance/compatibility]

vchuravy:
But won't the primal code be redefined?

```
                fnty = bcty.parameters[3]
                eltys = map(num_or_eltype, bcty.parameters[4].parameters)
                retty = Core.Compiler._return_type(interp, Tuple{fnty, eltys...})
```

I mean I have no clue what this code does.

wsmoses:
it's supposed to be effectively doing a fancy override table

PR #2156, opened 2024-12-02
Don't use ref when unnnecessary for gradient sugar

wsmoses:


wsmoses:
a bit down the line I may ask for help setting up backedges here

PR #2158, opened 2024-12-02
vc/fixup isapplicable use v2

vchuravy:
- **Use right caller for isapplicable usage**
- **use absint directly for applicable check**
- **fix nothing error**
- **fix return type**
- **Fix nmatches type mismatch**
- **forward sv and interp instead**
- **fixup! forward sv and interp instead**


wsmoses:
another excellent indicator here is that we've reduced the CI time from ~55 mins to ~45 minutes. Sadly not quite the ~30 mins I think we had before absint shenanigans [correctly tracking backedges] so maybe worth diving in after this too, but regardless a remarkable improvement.

wsmoses (approved):

wsmoses:
lgtm, but would you mind bumping the enzymecore/enzyme project.toml's

PR #2159, opened 2024-12-02
Fix build

wsmoses:


src/compiler.jl
```diff
@@ -3552,14 +3552,16 @@ function GPUCompiler.codegen(
 
         caller = mi
         if mode == API.DEM_ForwardMode
-            has_custom_rule =
-                EnzymeRules.has_frule_from_sig(specTypes; world, method_table, caller)
+            # This passes nothing which doesn't add an invalidation, but this
+            # is already handled via absint to get here
+            has_custom_rule = EnzymeRules.has_frule_from_sig(interp, specTypes, nothing)
```
vchuravy, line 3557: This should just use the old code with caller=nothing

lib/EnzymeCore/src/rules.jl
```diff
@@ -301,8 +305,8 @@ end
 
 
 function is_inactive_from_sig(@nospecialize(interp::Core.Compiler.AbstractInterpreter),
-                              @nospecialize(TT), sv::Core.Compiler.AbsIntState)
-    return isapplicable(inactive, TT; world, method_table, caller)
+                              @nospecialize(TT), sv::Union{Core.Compiler.AbsIntState, Nothing})
```
vchuravy, line 308: No this doesn't make sense, sv can't be nothing, the other code should drop the edge handling and be used instead

src/compiler.jl
wsmoses: Is there a caching or other advantage to using the interpreter here since it will be the same as inference?

lib/EnzymeCore/src/rules.jl
vchuravy: This code is morally equivalent to `abstract_applicable`

src/compiler.jl
vchuravy: Not really, the key is SV and us batching the edge update

Issue #2147, opened 2024-11-30
Nested AD Errors  Out


turiya4:
The following code (using Julia 1.10.6 and Enzyme 13.17) does simple Nested AD. However this results in an error. 

```
using Enzyme, Lux, Random, ComponentArrays, LinearAlgebra
n = 1
x_batch = randn(2, n)
y_batch = randn(2, n)
model = Chain(Parallel(vcat, Dense(2, 1, tanh), Dense(2, 1, tanh)), Dense(2, 1, tanh))
rng = Random.default_rng()
Random.seed!(rng, 0)
ps, st = Lux.setup(Xoshiro(0), model);
psaxes = getaxes(ComponentArray(ps))

nnfunc(x, y, psarray, st) = first(model((x, y), ComponentArray(psarray, psaxes), st))[1]
psarray = getdata(ComponentArray(ps))

function batch_error(xb, yb, psarray, st)
    val = zeros(n)
    for i = 1 : n
        dx = zeros(2)
        Enzyme.autodiff(Enzyme.Reverse, nnfunc, Active, Duplicated(xb[:, i], dx), Duplicated(yb[:, i], zeros(2)), Duplicated(psarray, zeros(Float32, size(psarray))), Const(st))
        val[i] = sum(dx.^2)
    end
    return sum(val)
end

dpsarray = zeros(Float32, size(psarray))
Enzyme.autodiff(Enzyme.Reverse, batch_error, Active, Duplicated(x_batch,zeros(size(x_batch))), Duplicated(y_batch,zeros(size(y_batch))), Duplicated(psarray, dpsarray), Const(st))
```

```
ERROR: LoadError: Enzyme compilation failed.
Current scope: 
; Function Attrs: mustprogress willreturn
[...]

PR #2161, opened 2024-12-03
Fix higher order codegen

wsmoses:


vchuravy:
What's the GPU story here?

wsmoses:
So the follow up there is basically to also have a similar handler that looks for deferred_(id) and we'll just run gpucompiler codegen there and inline the module similarly to here where we do the mapping variant from the pointer

vchuravy:
I want to remove the `deferred_id` buisness from GPUCompiler so that we can actually precompile GPU kernels,
and I am not keen on maintaining two code-paths.

wsmoses:
Hm how is pre compilation blocked?

but in any case yeah this PR primarily just fixes the “Enzyme.autodiff is the outermost compiler job” context

wsmoses:
More thought/design is merited for when that is not the case

vchuravy:
So this PR is the proposed new API for deferred compilation in GPUCompiler https://github.com/JuliaGPU/GPUCompiler.jl/pull/582 this is without any changes for Enzyme.

The rest of the stack is one possible way to add the necessary bits and bobs to make it Enzyme compatible.
Generally speaking, the rule for precompilation is "you shan't depend on runtime data"  so pointers or incrementing counters are out.

vchuravy:
> Hm how is pre compilation blocked?

We will serialize a Julia CodeInfo that contains a magical pointer/id? That is then used for a dictionary that only exists during the runtime of a particular code.

wsmoses:
hm yeah fair enough. 

In any case even if not a final resolution, would you be opposed to merging?

This net deletes code (including getting rid of a lot of runtime pointers) and removes the autodiff_deferred upgrade in absint, and thus fixes issues like https://github.com/EnzymeAD/Enzyme.jl/issues/2147

wsmoses:
[to be clear we would still need to follow up/continue here, but it also at least unblocks the CPU side higher order stuff]

vchuravy (approved):

PR #2171, opened 2024-12-05
WIP: use IPO cache for results

vchuravy:


src/compiler/interpreter.jl
```diff
@@ -173,6 +173,60 @@ end
...
+        elseif interp.reverse_rules && has_rrule_from_sig(interp, specTypes, linfo)
+            has_rule = true
+        end
+    end
+    @invoke Core.Compiler.finish(interp::Core.Compiler.AbstractInterpreter, opt::Core.Compiler.OptimizationState,
+                      ir::Core.Compiler.IRCode, caller::Core.Compiler.InferenceResult)
+    # Must happen afterwards
+    if inactive || has_rule
+        Core.Compiler.set_inlineable!(src, false)
+    end
```
vchuravy, line 226: @aviatesk does this look reasonable to you? 

The goal is to do the rule check once, per method and not once per call,

PR #2183, opened 2024-12-07
World backedge holder

wsmoses:


wsmoses:
@aviatesk @vchuravy @gbaraldi so the tuple backedge doesn't work. What's the right way to add a generic backedge?

wsmoses:
@vchuravy this really seems to want a MI: https://github.com/JuliaLang/julia/pull/32237/files

wsmoses:
@vtjnash idk if you have any thoughts here either

src/compiler.jl
```diff
@@ -5554,7 +5554,22 @@ function thunk_generator(world::UInt, source::LineNumberNode, @nospecialize(FA::
     # new_ci.min_world = min_world[]
     new_ci.min_world = world
     new_ci.max_world = max_world[]
-    new_ci.edges = Core.MethodInstance[mi]
+
+    edges = Core.MethodInstance[mi]
```
vchuravy, line 5558: edges = Any[mi]

vchuravy:
> @vchuravy this really seems to want a MI: https://github.com/JuliaLang/julia/pull/32237/files

Yeah, Jameson's comment there is correct. This should allow edges more diverse than just the `mi`.
as implemented correctly in the runtime https://github.com/JuliaLang/julia/blob/7192df7e900a6e4562e1670f544b8be8878ee359/src/method.c#L847-L863

src/compiler/interpreter.jl
```diff
@@ -23,6 +24,141 @@ else
...
+    end
+
+    for gen_sig in (
+        Tuple{typeof(EnzymeRules.inactive_noinl), Vararg{Any}},
+        Tuple{typeof(EnzymeRules.noalias), Vararg{Any}},
+        Tuple{typeof(EnzymeRules.inactive_type), Type},
+    )
+        EnzymeRules.add_mt_backedge!(gen_rule_be, ccall(:jl_method_table_for, Any, (Any,), gen_sig)::Core.MethodTable, gen_sig)
+    end
+end
```
vchuravy, line 160: What does this code aim to do? IIUC backedges are only triggered once and after that pruned. So this would only survive one invalidation.

src/compiler/interpreter.jl
```diff
@@ -225,25 +359,35 @@ function Core.Compiler.abstract_call_gf_by_type(
...
+        if interp.forward_rules
+            Core.Compiler.add_backedge!(sv, GPUCompiler.methodinstance(typeof(Enzyme.Compiler.Interpreter.rule_backedge_holder), Tuple{typeof(EnzymeRules.forward)}, interp.world)::Core.MethodInstance)
+            Enzyme.Compiler.Interpreter.rule_backedge_holder(Base.inferencebarrier(EnzymeRules.forward))
+        end
+        if interp.reverse_rules
+            Core.Compiler.add_backedge!(sv, GPUCompiler.methodinstance(typeof(Enzyme.Compiler.Interpreter.rule_backedge_holder), Tuple{typeof(EnzymeRules.augmented_primal)}, interp.world)::Core.MethodInstance)
+            Enzyme.Compiler.Interpreter.rule_backedge_holder(Base.inferencebarrier(EnzymeRules.augmented_primal))
+        end
+        Core.Compiler.add_backedge!(sv, GPUCompiler.methodinstance(typeof(Enzyme.Compiler.Interpreter.rule_backedge_holder), Tuple{typeof(EnzymeRules.inactive)}, interp.world)::Core.MethodInstance)
+        Enzyme.Compiler.Interpreter.rule_backedge_holder(Base.inferencebarrier(typeof(EnzymeRules.inactive)))
```
vchuravy, line 388: This should have had a comment that a `mi->mi` edge is cheaper than a `mi->mt` one.

test/ruleinvalidation.jl
```diff
@@ -42,6 +42,9 @@ end
 # now test invalidation for `inactive`
 inactive(::typeof(issue696), args...) = nothing
 @test autodiff(Forward, issue696, Duplicated(1.0, 1.0))[1] ≈ 0.0
-@test autodiff(Forward, call_issue696, Duplicated(1.0, 1.0))[1] ≈ 0.0
-
+@static if VERSION < v"1.11-"
+    @test_broken autodiff(Forward, call_issue696, Duplicated(1.0, 1.0))[1] ≈ 0.0
+else
+    @test autodiff(Forward, call_issue696, Duplicated(1.0, 1.0))[1] ≈ 0.0
+end
```
vchuravy, line 49: Yeah that tells me we are doing something wrong again.

src/compiler/interpreter.jl
```diff
@@ -225,25 +359,35 @@ function Core.Compiler.abstract_call_gf_by_type(
...
     else
-        # 1. Check if function is inactive
-        if is_inactive_from_sig(interp, specTypes, sv)
+        method_table = Core.Compiler.method_table(interp)
+        if EnzymeRules.is_inactive_from_sig(specTypes; world = interp.world, method_table)
             callinfo = NoInlineCallInfo(callinfo, atype, :inactive)
         else
-            # 2. Check if rule is defined
-            has_rule = get!(interp.rules_cache, specTypes) do
-                if interp.forward_rules && has_frule_from_sig(interp, specTypes, sv)
```
vchuravy, line 234: Probably should have then deleted the code in `tfuncs.jl` since this was the sole user.

src/compiler/interpreter.jl
vchuravy: And generally within inference the right thing is to use something like `abstract_call`. Instead of using the hacky version of `GPUCompiler.methodinstance`

src/compiler/interpreter.jl
wsmoses: ah so just move this inside our interp then? Basically this does the split we discussed, since I failed to add the generic methodtable backedge earlier =/. (I genuinely tried to add it there and failed, so you're going to have to help me add a non methodinstance backedge)

so all functions get a backedge to rule_backedge_holder. And rule_backedge_holder gets a backedge from <all rules>.

src/compiler/interpreter.jl
wsmoses: yeah sorry I don't know the absint well enough to know how to set that up, what's the right way to do?

test/ruleinvalidation.jl
wsmoses: yeah...........but at least not stupid compile time , but yea.....

Issue #2194, opened 2024-12-11
`Reverse` mode incorrectly detecting duplicated returns for a type unstable function


MasonProtter:
MWE:
```
using Enzyme

f_unstable = asin
f_stable(x) = asin(x)

function foo_unstable(x)
    y = sin(x)
    z = f_unstable(y) # untyped global variable
    cos(z)
end;

function foo_stable(x)
    y = sin(x)
    z = f_stable(y) # global const
    cos(z)
end;
```

```julia-repl
julia> autodiff(Reverse, foo_stable, Active(1.0))
((-0.8414709848078965,),)

julia> autodiff(Reverse, foo_unstable, Active(1.0))
ERROR: Duplicated Returns not yet handled
Stacktrace:
 [1] autodiff
   @ ~/.julia/packages/Enzyme/haqjK/src/Enzyme.jl:409 [inlined]
 [2] autodiff
   @ ~/.julia/packages/Enzyme/haqjK/src/Enzyme.jl:544 [inlined]
 [3] autodiff(mode::ReverseMode{false, false, FFIABI, false, false}, f::typeof(foo_unstable), args::Active{Float64})
   @ Enzyme ~/.julia/packages/Enzyme/haqjK/src/Enzyme.jl:516
 [4] top-level scope
   @ REPL[29]:1
```

system info:
```julia-repl
(jl_GwHTSw) pkg> st
Status `/tmp/jl_GwHTSw/Project.toml`
  [7da242da] Enzyme v0.13.22

julia> versioninfo()
Julia Version 1.11.2
Commit 5e9a32e7af2 (2024-12-01 20:02 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 16 × AMD Ryzen 7 7840U w/ Radeon  780M Graphics
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, znver4)
Threads: 8 default, 0 interactive, 4 GC (on 16 virtual cores)
Environment:
  JULIA_NUM_THREADS = 8
  JULIA_EDITOR = emacsclient
```

MasonProtter:
Okay, so update based on what I was able to infer from @wsmoses's comment in #2195, this is only a problem if I don't assert the returned activity:
```julia
julia> autodiff(Reverse, foo_unstable, Active, Active(1.0))
((-0.8414709848078965,),)
```

[...]

PR #2195, opened 2024-12-11
Some tweaks to the Getting Started docs

MasonProtter:
I'd like to get around to some more comprehensive documentation contributions, but thought I'd start with some very low hanging fruit / minor tweaks to the `getting started` page:

* I removed the pattern `autodiff(mode, f, activity, args...)` and replaced it with `autodiff(mode, f, args...)` since telling it the activity mode is no longer needed
* Before delving into various things, I just made it clear that we expect `rosenbrock(1.0, 2.0) == 100.0` and the derivative w.r.t. `x` should be `-400.0` and the derivative w.r.t. `y` should be `200.0`. This should hopefully help users orient themselves a bit quicker when they're trying to interpret the (perhaps confusing) outputs of the various `autodiff` calls. 
* I mentioned what we mean by 'primal' since this isn't necessarily a word everyone who took a calculus course knows

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -31,23 +31,25 @@ Also see [Implementing pullbacks](@ref) on how to implement back-propagation for
 We will try a few things with the following functions:
 
 ```jldoctest rosenbrock
-julia> rosenbrock(x, y) = (1.0 - x)^2 + 100.0 * (y - x^2)^2
-rosenbrock (generic function with 1 method)
+julia> rosenbrock(x, y) = (1.0 - x)^2 + 100.0 * (y - x^2)^2;
 
-julia> rosenbrock_inp(x) = (1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2)^2
-rosenbrock_inp (generic function with 1 method)
+julia> rosenbrock_inp(x) = (1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2)^2;
 ```
+
+where we note for future reference that the value of this function at `x=1.0`, `y=2.0` is `100.0`, and its derivative
+with respect to `x` at that point is `-400.0`, and its derivative with respect to `y` at that point is `200.0`. 
 
 ## Reverse mode
 
 The return value of reverse mode [`autodiff`](@ref) is a tuple that contains as a first value
-the derivative value of the active inputs and optionally the primal return value.
+the derivative value of the active inputs and optionally the _primal_ return value (i.e. the 
+value of the undifferentiated function).
 
 ```jldoctest rosenbrock
-julia> autodiff(Reverse, rosenbrock, Active, Active(1.0), Active(2.0))
+julia> autodiff(Reverse, rosenbrock, Active(1.0), Active(2.0))
 ((-400.0, 200.0),)
 
-julia> autodiff(ReverseWithPrimal, rosenbrock, Active, Active(1.0), Active(2.0))
+julia> autodiff(ReverseWithPrimal, rosenbrock, Active(1.0), Active(2.0))
 ((-400.0, 200.0), 100.0)
 ```
 
@@ -62,7 +64,7 @@ julia> dx = [0.0, 0.0]
  0.0
  0.0
 
-julia> autodiff(Reverse, rosenbrock_inp, Active, Duplicated(x, dx))
+julia> autodiff(Reverse, rosenbrock_inp, Duplicated(x, dx))
 ((nothing,),)
 
 julia> dx
@@ -71,8 +73,9 @@ julia> dx
   200.0
 ```
 
-Both the inplace and "normal" variant return the gradient. The difference is that with
-[`Active`](@ref) the gradient is returned and with [`Duplicated`](@ref) the gradient is accumulated in place.
+Both the inplace and "normal" variant return the gradient. The difference is that with [`Active`](@ref) 
+the gradient is returned and with [`Duplicated`](@ref) the gradient is accumulated in-place into `dx`, 
+and a value of `nothing` is placed in the corresponding slot of the returned `Tuple`.
 
 ## Forward mode
 
@@ -121,7 +124,7 @@ julia> dx = [1.0, 1.0]
  1.0
  1.0
 
-julia> autodiff(ForwardWithPrimal, rosenbrock_inp, Duplicated, Duplicated(x, dx))
+julia> autodiff(ForwardWithPrimal, rosenbrock_inp, Duplicated(x, dx))
 (-400.0, 400.0)
 ```
 


wsmoses:
so telling it the activity is still helpful. Enzyme does its best job to deduce the return activity, but if something is unstable (e.g. like in https://github.com/EnzymeAD/Enzyme.jl/issues/2194) it may fail. In the case of the issue I presume marking the return activity as Active explicitly would allow the type unstable case to succeed.

That said, that doesn't mean we need to introduce the more complex version at the start

wsmoses (approved):

docs/src/index.md
```diff
@@ -31,23 +31,25 @@ Also see [Implementing pullbacks](@ref) on how to implement back-propagation for
...
-julia> rosenbrock(x, y) = (1.0 - x)^2 + 100.0 * (y - x^2)^2
-rosenbrock (generic function with 1 method)
+julia> rosenbrock(x, y) = (1.0 - x)^2 + 100.0 * (y - x^2)^2;
 
-julia> rosenbrock_inp(x) = (1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2)^2
-rosenbrock_inp (generic function with 1 method)
+julia> rosenbrock_inp(x) = (1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2)^2;
 ```
 
+where we note for future reference that the value of this function at `x=1.0`, `y=2.0` is `100.0`, and its derivative
```
mcabbott, line 39: Consider showing this as code instead of prose? 
```
julia> rosenbrock(x, y) = (1.0 - x)^2 + 100.0 * (y - x^2)^2;

julia> rosenbrock_inp(xy) = (1.0 - xy[1])^2 + 100.0 * (xy[2] - xy[1]^2)^2;

julia> rosenbrock(1.0, 2.0) == rosenbrock_inp([1.0, 2.0]) == 100.0
true
```
I also think you should not call the input of `rosenbrock_inp` the same thing, `x == [x, y]` is weird. The name `rosenbrock_inp` also seems a bit weird, maybe it can just be another method, or if that's too confusing, add a suffix more informative than "inp"? (I'm not sure what INP means, maybe input, but why?)

docs/src/index.md
```diff
@@ -31,23 +31,25 @@ Also see [Implementing pullbacks](@ref) on how to implement back-propagation for
...
+with respect to `x` at that point is `-400.0`, and its derivative with respect to `y` at that point is `200.0`. 
+
 ## Reverse mode
 
 The return value of reverse mode [`autodiff`](@ref) is a tuple that contains as a first value
-the derivative value of the active inputs and optionally the primal return value.
+the derivative value of the active inputs and optionally the _primal_ return value (i.e. the 
+value of the undifferentiated function).
```
mcabbott, line 46: Consider not using "value" to mean so many things here?

> is a tuple that contains as a first *element* the *derivatives* of ..., and optionally the primal *value* (i.e. what the function returns).

mcabbott: Maybe "optionally" also seems a bit odd to describe the output not the input. It's not that you may omit this. It's that ReverseWithPrimal tells it to.

docs/src/index.md
wsmoses: I think it was originally in place (@michel2323 were you the one to originally author this doc, just by virtue of it being rosenbrock?)

But either way sure!

docs/src/index.md
wsmoses: yeah we definitely don't need to say "derivative value" and can just say "the derivative of"

wsmoses: and "the value of the undifferentiated function" -> "the result of the original function without differentiation"

mcabbott: Also consider putting the ReverseWithPrimal case first, as without it,  `((-400.0, 200.0),)` seems like a puzzle to count the brackets & guess why.

wsmoses: oh yeah totally fair, if you want to put that in this PR that would be fine with me!

docs/src/index.md
mcabbott: Ah ok. But this function isn't in-place, it's just going to be used somewhere below in a demonstration that Enzyme likes to handle functions which accept Vector by mutating something else. The reader doesn't know that yet.

wsmoses: very true, maybe rosenbrok_array or something? or even just rosenbrock2

docs/src/index.md
```diff
@@ -71,8 +73,9 @@ julia> dx
   200.0
 ```
 
-Both the inplace and "normal" variant return the gradient. The difference is that with
-[`Active`](@ref) the gradient is returned and with [`Duplicated`](@ref) the gradient is accumulated in place.
+Both the inplace and "normal" variant return the gradient. The difference is that with [`Active`](@ref)
```
mcabbott, line 76: This wording seems weird. The inplace version returns `((nothing,),)`, it's written right there. And inplace / "normal" are new terms here, which aren't the terms you need to learn to understand Enzyme.

> The version with `Active` (for immutable inputs like x::Float64) returns the gradient. The version with `Duplicated` (for mutable inputs like `x::Vector{Float64}`) instead writes the gradient into the `Duplicated` object, and returns `nothing` in the corresponding slot of the returned `derivs`.

docs/src/index.md
```diff
@@ -31,23 +31,25 @@ Also see [Implementing pullbacks](@ref) on how to implement back-propagation for
...
-rosenbrock (generic function with 1 method)
+julia> rosenbrock(x, y) = (1.0 - x)^2 + 100.0 * (y - x^2)^2;
 
-julia> rosenbrock_inp(x) = (1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2)^2
-rosenbrock_inp (generic function with 1 method)
+julia> rosenbrock_inp(x) = (1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2)^2;
 ```
 
+where we note for future reference that the value of this function at `x=1.0`, `y=2.0` is `100.0`, and its derivative
+with respect to `x` at that point is `-400.0`, and its derivative with respect to `y` at that point is `200.0`.
```
mcabbott, line 40:
````suggestion
julia> rosenbrock(xy::Vector) = (1.0 - xy[1])^2 + 100.0 * (xy[2] - xy[1]^2)^2;

julia> z = rosenbrock(1.0, 2.0)
100.0

julia> z == rosenbrock([1.0, 2.0])  # Vector method
true
```
We note for future reference that the value of this function at `x=1.0`, `y=2.0` is `z=100.0`. Its derivative with respect to `x` at that point is `-400.0`, and its derivative with respect to `y` is `200.0`. 
````

File: docs/src/index.md (hunk 2 above)
Comment (mcabbott) [reply docs/src/index.md:46]: I can't make suggestions across deleted lines :/  so this is going to be messy...

Comment (mcabbott) [inline docs/src/index.md:46]:
```suggestion
the derivatives with respect to the inputs. The tuple's second element is the _primal_ value (i.e. the result of the original function without differentiation),
but this is omitted if you use `Reverse` instead of `ReverseWithPrimal`:
```

docs/src/index.md
```diff
@@ -31,23 +31,25 @@ Also see [Implementing pullbacks](@ref) on how to implement back-propagation for
...
 ## Reverse mode
 
 The return value of reverse mode [`autodiff`](@ref) is a tuple that contains as a first value
-the derivative value of the active inputs and optionally the primal return value.
+the derivative value of the active inputs and optionally the _primal_ return value (i.e. the 
+value of the undifferentiated function).
 
 ```jldoctest rosenbrock
-julia> autodiff(Reverse, rosenbrock, Active, Active(1.0), Active(2.0))
+julia> autodiff(Reverse, rosenbrock, Active(1.0), Active(2.0))
```
mcabbott, line 49:
```suggestion
julia> derivs, z = autodiff(ReverseWithPrimal, rosenbrock, Active(1.0), Active(2.0))
((-400.0, 200.0), 100.0)

julia> autodiff(Reverse, rosenbrock, Active(1.0), Active(2.0))
```

docs/src/index.md
```diff
@@ -31,23 +31,25 @@ Also see [Implementing pullbacks](@ref) on how to implement back-propagation for
...
 ((-400.0, 200.0),)
 
-julia> autodiff(ReverseWithPrimal, rosenbrock, Active, Active(1.0), Active(2.0))
+julia> autodiff(ReverseWithPrimal, rosenbrock, Active(1.0), Active(2.0))
 ((-400.0, 200.0), 100.0)
```
mcabbott, line 53:
```suggestion
```

docs/src/index.md
```diff
@@ -62,7 +64,7 @@ julia> dx = [0.0, 0.0]
  0.0
  0.0
 
-julia> autodiff(Reverse, rosenbrock_inp, Active, Duplicated(x, dx))
+julia> autodiff(Reverse, rosenbrock_inp, Duplicated(x, dx))
```
mcabbott, line 67:
```suggestion
julia> autodiff(Reverse, rosenbrock, Duplicated(x, dx))
```

docs/src/index.md
mcabbott:
```suggestion
Both versions calculate the same derivatives. The difference is that with [`Active`](@ref) arguments
```

docs/src/index.md
```diff
@@ -71,8 +73,9 @@ julia> dx
   200.0
 ```
 
-Both the inplace and "normal" variant return the gradient. The difference is that with
-[`Active`](@ref) the gradient is returned and with [`Duplicated`](@ref) the gradient is accumulated in place.
+Both the inplace and "normal" variant return the gradient. The difference is that with [`Active`](@ref) 
+the gradient is returned and with [`Duplicated`](@ref) the gradient is accumulated in-place into `dx`, 
+and a value of `nothing` is placed in the corresponding slot of the returned `Tuple`.
```
mcabbott, line 78:
```suggestion
(for immutable inputs like `x::Float64`) returns them as `derivs`, while the version with `Duplicated` (for mutable inputs like `x::Vector{Float64}`) instead writes the gradient into the `Duplicated` object, and returns `nothing` in the corresponding slot of the returned `derivs`. 
In fact it accumulates the gradient, i.e. if you run it again it will double `dx`.
In general, `autodiff` accepts any mix of `Active` and `Duplicated` function arguments, as well as `Const` and various other `Annotation` types.

the gradient is returned and with [`Duplicated`](@ref) the gradient is accumulated in-place into `dx`, 
and a value of `nothing` is placed in the corresponding slot of the returned `Tuple`.
```

docs/src/index.md
```diff
@@ -121,7 +124,7 @@ julia> dx = [1.0, 1.0]
  1.0
  1.0
 
-julia> autodiff(ForwardWithPrimal, rosenbrock_inp, Duplicated, Duplicated(x, dx))
+julia> autodiff(ForwardWithPrimal, rosenbrock_inp, Duplicated(x, dx))
```
mcabbott, line 127:
```suggestion
julia> autodiff(ForwardWithPrimal, rosenbrock, Duplicated(x, dx))
```

mcabbott (commented):
This is messy but I made many suggestions.

* Should it still mention `autodiff(Reverse, f, Active, Duplicated())` i.e. annotating the return type, somewhere? Maybe it's not the first thing to know, but it is surprising to encounter later.

* Should this page also introduce `gradient(Reverse, f, ...)`? Should it do this before or after `autodiff`?

PR #2199, opened 2024-12-14
Fix methodinstance usage and backedges

wsmoses:
Hopefully now that I understand backedges, this fixes pending issues.

Simultaneously this gets rid of a bunch of inner generated functions so hopefully compile time comes down.

Time to first derivative (post precompile) is now:

```julia
julia> @time autodiff(Forward, sin, Duplicated(1.0, 2.0))
  0.000014 seconds
(1.0806046117362795,)

julia> @time autodiff(Forward, sin, Duplicated(1.0, 2.0))
  0.000003 seconds
(1.0806046117362795,)

julia> @time autodiff(Forward, sin, Duplicated(1.0, 2.0))
  0.000002 seconds
(1.0806046117362795,)
```

### DIFF
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -318,7 +318,7 @@ include("llvm/passes.jl")
 include("typeutils/make_zero.jl")
 
 function nested_codegen!(mode::API.CDerivativeMode, mod::LLVM.Module, @nospecialize(f), @nospecialize(tt::Type), world::UInt)
-    funcspec = my_methodinstance(typeof(f), tt, world)
+    funcspec = my_methodinstance(mode == API.DEM_ForwardMode ? Forward : Reverse, typeof(f), tt, world)
     nested_codegen!(mode, mod, funcspec, world)
 end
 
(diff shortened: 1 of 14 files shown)


src/compiler.jl
```diff
@@ -1258,6 +1272,8 @@ Create the methodinstance pair, and lookup the primal return type.
     @nospecialize(TT::Type),
     world::Union{UInt,Nothing} = nothing,
 )
+
+fdsafdsafsa
```
vchuravy, line 1276: ?

src/compiler/interpreter.jl
```diff
@@ -198,6 +176,37 @@ function EnzymeInterpreter(
     else
         InferenceParams(; unoptimize_throw_blocks=false)
     end
+    
+    @static if HAS_INTEGRATED_CACHE
```
vchuravy, line 180: I was also contemplating that Enzyme for 1.11 could simply not use the integrated cache... 
But we might have removed some interface...

src/compiler/tfunc.jl
```diff
@@ -2,30 +2,32 @@ import EnzymeCore: Annotation
 import EnzymeCore.EnzymeRules: FwdConfig, RevConfig, forward, augmented_primal, inactive, _annotate_tt
 
 function has_frule_from_sig(@nospecialize(interp::Core.Compiler.AbstractInterpreter),
-    @nospecialize(TT), sv::Core.Compiler.AbsIntState)::Bool
+    @nospecialize(TT::Type), sv::Core.Compiler.AbsIntState, partialedge::Bool=true)::Bool
```
vchuravy, line 5: What is a `partialedge`?  It seems not used?

PR #2201, opened 2024-12-15
Mark globals as externallt init on nested link

wsmoses:


kdaning:
It's working now, thanks!

PR #2273, opened 2025-01-18
[GHA] Add timeouts and run x86_64 macOS jobs on native hardware

giordano:
x86_64 macOS test jobs are timing out, and the default timeout is an unreasonable 6 hours, which clogs up the queue in the entire GitHub organisation.  This PR sets lower timeouts for individual CI jobs (2 hours for the tests), also makes x86_64 macOS jobs run on native hardware instead of going through Rosetta, which may or may not be the cause of the hangs.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -15,6 +15,7 @@ concurrency:
 
 jobs:
   test:
+    timeout-minutes: 120
     name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ matrix.libEnzyme }} libEnzyme - assertions=${{ matrix.assertions }} - ${{ github.event_name }}
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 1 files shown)


giordano:
> instead of going through Rosetta, which may or may not be the cause of the hangs.

Actually, I'm somewhat positive the issue is using Rosetta, since [the last message printed](https://github.com/EnzymeAD/Enzyme.jl/actions/runs/12835170894/job/35794050875#step:11:5382) is
```
rosetta error: unexpectedly need to EmulateForward on a synchronous exception x86_rip=0x4483051488 arm_pc=0x4491036424 num_insts=4 inst_index=3 x86 instruction bytes: 0x6215344901283465301 0x18125098399031709779
```

giordano:
And just to confirm this does fix the issue: https://github.com/EnzymeAD/Enzyme.jl/actions/runs/12843787914/job/35816114301?pr=2273 finished in 40 minutes without errors 🙂

wsmoses:
oh awesome!

danielwe:
Why not keep the arm64 macOS instances and run natively instead of through Rosetta?

giordano:
This PR was minimally changing and meant only to make existing setup working without errors.

Issue #2296, opened 2025-02-03
Method Error: no method matching MixedDuplicated for Oceananigans run with CATKE


jlk9:
Running this code:

https://github.com/DJ4Earth/Enzymanigans.jl/blob/jlk9/with-simulation/dynamical_core/autodiff_double_gyre.jl

Produces the following method error:

[error_catke_0203.txt](https://github.com/user-attachments/files/18649358/error_catke_0203.txt)

@wsmoses

SouthEndMusic:
I can't open the link so I cannot test whether it works now, is `Enzymanigans.jl` a private repository? You could try yourself on the branch of my PR:

https://github.com/EnzymeAD/Enzyme.jl/pull/2297

jlk9:
https://github.com/DJ4Earth/Enzymanigans.jl/blob/jlk9/reduce-for-catke-0210/dynamical_core/autodiff_double_gyre.jl

wsmoses:
```
using Enzyme

using KernelAbstractions: @kernel, @index, CPU as KCPU

Enzyme.API.looseTypeAnalysis!(true)
Enzyme.API.strictAliasing!(true)

function compute_diffusivities!(J, tracers)

    top_tracer_bcs = NamedTuple(c => tracers[c].boundary_conditions.top for c in propertynames(tracers))

    worksize = (16, 16)
    loop!    = compute_average_surface_buoyancy_flux!(KCPU(), worksize, worksize)

    loop!(J, top_tracer_bcs)

    return nothing
end

@kernel function compute_average_surface_buoyancy_flux!(J, top_tracer_bcs)
    i, j = @index(Global, NTuple)
    J[i, j, 1] = J[i, j, 1]
end

struct CPU     end
struct Bounded end

struct GridHomemade{FT, TX, TY, TZ, CZ, M, MY, FX, FY, VX, VY, Arch}
    architecture :: Arch
    Nx :: Int
    Ny :: Int
    Nz :: Int
    Hx :: Int
    Hy :: Int
    Hz :: Int
    Lx :: FT
    Ly :: FT
    Lz :: FT
    # All directions can be either regular (FX, FY, FZ) <: Number
    # or stretched (FX, FY, FZ) <: AbstractVector
    Δλᶠᵃᵃ :: FX
    Δλᶜᵃᵃ :: FX
    λᶠᵃᵃ  :: VX
    λᶜᵃᵃ  :: VX
[...]

PR #2297, opened 2025-02-04
Fix jitrule

SouthEndMusic:
Fixes https://github.com/EnzymeAD/Enzyme.jl/issues/2296 (?)

### DIFF
--- a/src/rules/jitrules.jl
+++ b/src/rules/jitrules.jl
@@ -2,11 +2,11 @@ function setup_macro_wraps(
     forwardMode::Bool,
     N::Int,
     Width::Int,
-    base = nothing,
-    iterate = false;
-    func = true,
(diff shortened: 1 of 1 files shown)


wsmoses (changes requested):
So the error here is that the input should have been a ref and adding the ref here will cause other errors

SouthEndMusic:
So it's a downstream problem, not an `Enzyme.jl` problem? Or should the error have been caught earlier in `Enzyme.jl`?

wsmoses:
no it is an issue here most likely, but this is not the function going awry.

whatever code creates the arguments to the function provide invalid arguments [which you may need to dataflow back to find]

I'd recommend trting to make a minimal reproducer of your issue

PR #2301, opened 2025-02-08
DI: Reverse dependency structure

wsmoses:
@gdalle this will enable the DI ext to more properly touch internals if need be

### DIFF
--- a/ext/EnzymeDIExt/reverse_twoarg.jl
+++ b/ext/EnzymeDIExt/reverse_twoarg.jl
@@ -1,0 +1,150 @@
+## Pullback
+
+struct EnzymeReverseTwoArgPullbackPrep{TY} <: DI.PullbackPrep
+    ty_copy::TY
+end
+
+function DI.prepare_pullback(
+    f!::F,
+    y,
+    ::AutoEnzyme{<:Union{ReverseMode,Nothing}},
+    x,
+    ty::NTuple,
+    contexts::Vararg{DI.Context,C},
+) where {F,C}
+    return EnzymeReverseTwoArgPullbackPrep(map(copy, ty))
+end
+
+function DI.value_and_pullback(
+    f!::F,
+    y,
+    prep::EnzymeReverseTwoArgPullbackPrep,
+    backend::AutoEnzyme{<:Union{ReverseMode,Nothing}},
+    x::Number,
+    ty::NTuple{1},
+    contexts::Vararg{DI.Context,C},
+) where {F,C}
+    copyto!(only(prep.ty_copy), only(ty))
+    mode = reverse_noprimal(backend)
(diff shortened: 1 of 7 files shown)


ext/EnzymeDIExt/utils.jl
```diff
@@ -0,0 +1,130 @@
...
+
+function reverse_split_withprimal(backend::AutoEnzyme{<:ReverseMode})
+    return set_err(WithPrimal(Split(backend.mode)), backend)
+end
+
+function reverse_split_withprimal(backend::AutoEnzyme{Nothing})
+    return set_err(ReverseSplitWithPrimal, backend)
+end
+
+set_err(mode::Mode, ::AutoEnzyme{<:Any,Nothing}) = EnzymeCore.set_err_if_func_written(mode)
```
wsmoses, line 103: @gdalle as discussed on slack this probably should be an extension to the set_err_if_func_written function to take an ADmode, so likely we have this in an EnzymeCoreADTypes ext?

ext/EnzymeDIExt/utils.jl
```diff
@@ -0,0 +1,130 @@
...
+
+## Modes
+
+forward_noprimal(backend::AutoEnzyme{<:ForwardMode}) = NoPrimal(backend.mode)
+forward_noprimal(::AutoEnzyme{Nothing}) = Forward
+
+forward_withprimal(backend::AutoEnzyme{<:ForwardMode}) = WithPrimal(backend.mode)
+forward_withprimal(::AutoEnzyme{Nothing}) = ForwardWithPrimal
+
+reverse_noprimal(backend::AutoEnzyme{<:ReverseMode}) = NoPrimal(backend.mode)
```
wsmoses, line 89: @gdalle similarly here we can make an ADTypes ext func for get_mode_or_default(AutoEnzyme, defaultMode)

ext/EnzymeDIExt/reverse_twoarg.jl
```diff
@@ -0,0 +1,150 @@
...
+) where {F,C}
+    copyto!(only(prep.ty_copy), only(ty))
+    mode = reverse_noprimal(backend)
+    f!_and_df! = get_f_and_df(f!, backend, mode)
+    dx_sametype = convert(typeof(x), only(tx))
+    make_zero!(dx_sametype)
+    dy_sametype = convert(typeof(y), only(prep.ty_copy))
+    x_and_dx = Duplicated(x, dx_sametype)
+    y_and_dy = Duplicated(y, dy_sametype)
+    annotated_contexts = translate(backend, mode, Val(1), contexts...)
```
wsmoses, line 122: @gdalle I presume this can be moved into Enzyme.gradient! And have DI call that?

PR #2314, opened 2025-02-18
Unsafe wrap is not supported fully

vchuravy:
The added test currently fails with:

```
cannot find shadow for   %19 = call nonnull "enzyme_type"="{[-1]:Pointer}" {} addrspace(10)* @ijl_ptr_to_array({} addrspace(10)* noundef addrspacecast ({}* inttoptr (i64 139992576232976 to {}*) to {} addrspace(10)*), i64 %16, {} addrspace(10)* noundef nonnull %17, i32 noundef 0) #27, !dbg !75

Caused by:
Stacktrace:
 [1] #unsafe_wrap#115
   @ ./pointer.jl:110
 [2] unsafe_wrap
   @ ./pointer.jl:108
 [3] Container
```

### DIFF
--- a/test/array.jl
+++ b/test/array.jl
@@ -1,40 +1,89 @@
 using Enzyme, Test
 
 function ptrcopy(B, A)
-@static if VERSION < v"1.11"
-	Base.unsafe_copyto!(B, 1, A, 1, 2)
-else
-	Base.unsafe_copyto!(B.ref, A.ref, 2)
-end
-	nothing
+    @static if VERSION < v"1.11"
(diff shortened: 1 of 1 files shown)


vchuravy:
@wsmoses how do I best go about fixing this?

wsmoses:
I would like look at the rule we have for that and make sure we initialize/store the shadow.

note that I’m not sure we implemented this so far, partially due to a semantics question. If the function copies in reverse we need to do the reverse += copy, but if not it should have an empty reverse pass

PR #2316, opened 2025-02-20
Fix gc error

wsmoses:


wsmoses:
I have

```julia
using Enzyme

Enzyme.Compiler.DumpPostOpt[] = true 

struct ObjPair{A, B}
	x::A
	y::B
end

function makesum(pair)
	return pair.x[1]
end

function f(x)
	obj = ObjPair(x, 3)
	r = Base.inferencebarrier(makesum)(obj)
	return r::Float64
end


sprint() do io
    Enzyme.Compiler.enzyme_code_llvm(io, f, Active, Tuple{Duplicated{Vector{Float64}}})
end
```

as some test code, but I'm not sure how to make this into a good check, if @vchuravy you have any ideas

PR #2342, opened 2025-03-28
Handle recursive store for primitive types symbol/string

wsmoses:


wsmoses:
your fix might be more general, but also not completely solve the issue. Essentially from here we have this fallback case if we know the object is a jl_valuet and we just put a null pointer there. Here we hardcode adding string/symbol (but ensuring we actually do the null store), whereas yours has a more general 0 check but doesn't store (potentially still having the GC issue)

vchuravy:
So we should likely do both? Aka it makes no sense to try and store into a datatype without any fields, but we should also create a shadow store for those objects.

vchuravy:
Check #2263?

wsmoses:
@vchuravy we should discuss this next week. going to merge to unblock flux test infra

PR #2352, opened 2025-04-11
Add common error supertype

### LINKED ISSUE #2351: Define a common supertype `EnzymeError` for all exceptions and make it public?
Would you be open to a PR doing something like

```julia
"""
    EnzymeError

Common supertype for all Enzyme-specific exceptions.
"""
abstract type EnzymeError <: Base.Exception
public EnzymeError
```

and then making all your internal exceptions subtype it?

The goal would be to decrease maintenance burden on your end by catching some common errors in DI and providing more context with error hints. Error hints dispatch on the exception type, and I don't want to add them to every single error thrown by Julia code. That means I need a way to identify the exceptions that are thrown by Enzyme itself, and then append nice messages to them, like so:

```julia
function __init__()
    Base.Experimental.register_error_hint(EnzymeError) do io, exception
        if contains(exception.msg, "runtime activity")
            println(io, """
                It seems you need to activate runtime activity for Enzyme.
                You can do it by modifying the `mode` field of your `AutoEnzyme` backend object before calling DifferentiationInterface.
                Example: `AutoEnzyme(mode=Enzyme.set_runtime_activity(Enzyme.Forward))`.
            """)
        end
    end
end
```

Note that it would require your custom `showerror` methods to call `Base.Experimental.show_error_hints`. In any case, the modifications to Enzyme should be limited to the [`errors.jl`](https://github.com/EnzymeAD/Enzyme.jl/blob/main/src/errors.jl) file. 

See also:

- https://docs.julialang.org/en/v1/base/base/#Base.Experimental.register_error_hint 
- https://discourse.julialang.org/t/enzyme-with-const-on-a-vector-throws-an-error/121927/20

gdalle:
Fixes #2351 

Note that this uses an experimental feature of Julia, but I think Enzyme is so tightly linked to the compiler that this is the least of your problems

### DIFF
--- a/src/errors.jl
+++ b/src/errors.jl
@@ -16,7 +16,9 @@ struct EnzymeRuntimeException <: EnzymeError
 end
 
 function Base.showerror(io::IO, ece::EnzymeRuntimeException)
-    Base.Experimental.show_error_hints(io, ece)
+    if isdefined(Base.Experimental, :show_error_hints)
+        Base.Experimental.show_error_hints(io, ece)
+    end
     print(io, "Enzyme execution failed.\n")
     msg = Base.unsafe_string(ece.msg)
     print(io, msg, '\n')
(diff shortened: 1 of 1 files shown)


vchuravy:
Thanks!

PR #2373, opened 2025-04-30
Revert "Avoid blow up of function pre-amble on higer order AD (#2350)"

wsmoses:
This reverts commit 6059bb34e4ff5fe97476f43aac89a7b0ac492c3d.

Let's see if 1.11 CI comes back


wsmoses:
okay so surprisingly @vchuravy this did fix it.

I'm going to merge to unbreak 1.11 CI and we can fix in a follow up

PR #2309, opened 2025-02-16
feat: VJP utility based on `autodiff_thunk`

gdalle:
Fixes #1853 

Related:

- https://github.com/JuliaDiff/DifferentiationInterface.jl/discussions/721

PR #2389, opened 2025-05-12
Use sources for integration tests

vchuravy:


vchuravy:
Set's up the integration tests such that https://github.com/EnzymeAD/Enzyme.jl/pull/2309#issuecomment-2868907605 is not an issue, but highlights an issue with  Bijectors and 1.11

vchuravy:
cc: @mhauru

PR #2394, opened 2025-05-13
Add batched writebarrier test

wsmoses:


vchuravy:
It might make sense to add this as a lit test? Since we care about how the write barrier is emitted.

wsmoses:
yeah that's fair, julia side? we do have a lit test on the enzyme side for this (and we should await https://github.com/JuliaPackaging/Yggdrasil/pull/11190)

vchuravy:
https://github.com/EnzymeAD/Enzyme/pull/2313 ? 

I am confused that we are removing a write barrier instead of emitting one for N children?

wsmoses:
that was a fixup to the actual fix which was https://github.com/EnzymeAD/Enzyme/pull/2313, though I now realize that test clearlt didn't do anything.

The issue was for batch mode we emitted one write barrier for the whole batch, concat'ing all args from both batches into one instead of, write_barrier(shadowptr1, shadow_val1), write_barrier(shadowptr2, shadow_val2) -- which I did confirm on this code.

I guess that really indicates we ought have a lit test

PR #2406, opened 2025-05-22
Attributes for jl_typeof

### LINKED ISSUE #2405: No augmented forward pass found for `jl_typeof`
Yet another in the series of diffeq examples, I get this error on julia v1.11, (but it runs in v1.10)
```julia
using SciMLSensitivity, OrdinaryDiffEq, Zygote, Enzyme

function fiip(du, u, p, t)
    du[1] = dx = p[1] * u[1] - p[2] * u[1] * u[2]
    du[2] = dy = -p[3] * u[2] + p[4] * u[1] * u[2]
end

let
    p = [1.5, 1.0, 3.0, 1.0];
    u0 = [1.0; 1.0];
    prob = ODEProblem(fiip, u0, (0.0, 10.0), p)
    sol = solve(prob, Tsit5())
    loss(u0, p) = sum(solve(prob, Tsit5(), u0 = u0, p = p, saveat = 0.1))
    du = zeros(size(u0))
    dp = zeros(size(p))
    dp[1] = 1.0
    dloss = make_zero(loss)
    Enzyme.autodiff(Reverse, Const(loss), Active, Duplicated(u0, du), Duplicated(p, dp))
end
```
```julia
ERROR: 
No augmented forward pass found for jl_typeof
 at context:   %8 = call "enzyme_type"="{[-1]:Pointer}" {} addrspace(10)* ({} addrspace(10)*, ...) @jl_typeof({} addrspace(10)* nonnull %7) #24, !dbg !16

Stacktrace:
 [1] loss
   @ ./REPL[3]:0


Stacktrace:
 [1] loss
   @ ./REPL[3]:0 [inlined]
 [2] diffejulia_loss_40108_inner_1wrap
   @ ./REPL[3]:0
 [3] macro expansion
   @ ~/.julia/packages/Enzyme/hu9gq/src/compiler.jl:5463 [inlined]
 [4] enzyme_call
   @ ~/.julia/packages/Enzyme/hu9gq/src/compiler.jl:4997 [inlined]
 [5] CombinedAdjointThunk
   @ ~/.julia/packages/Enzyme/hu9gq/src/compiler.jl:4869 [inlined]
 [6] autodiff(::ReverseMode{…}, ::Const{…}, ::Type{…}, ::Duplicated{…}, ::Duplicated{…})
   @ Enzyme ~/.julia/packages/Enzyme/hu9gq/src/Enzyme.jl:504
 [7] top-level scope
   @ REPL[3]:11
```

MasonProtter:
Fixes #2405 for me locally

I'm not sure the best way to test this though. Could we possibly add the MWE from that MWE as a DiffEq integration test?

### DIFF
--- a/src/llvm/attributes.jl
+++ b/src/llvm/attributes.jl
@@ -1,4 +1,5 @@
 const nofreefns = Set{String}((
+    "jl_typeof",
     "julia.gc_loaded",
     "jl_egal__unboxed", "ijl_egal__unboxed",
     "jl_restore_excstack",
@@ -158,6 +159,7 @@ const nofreefns = Set{String}((
 ))
 
 const inactivefns = Set{String}((
+    "jl_typeof",
     "jl_egal__unboxed", "ijl_egal__unboxed",
     "ClientGetDevice",
     "BufferOnCPU",
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -76,6 +76,7 @@ include("typetree.jl")
 include("passes.jl")
 include("optimize.jl")
 include("make_zero.jl")
+include("runtime_calls.jl")
 
 include("rules.jl")
 include("rrules.jl")
@@ -3771,15 +3772,6 @@ end
     @test res[2][6] ≈ 6.0
 end
 
-@testset "jl_typeof" begin
-    # https://github.com/EnzymeAD/Enzyme.jl/issues/2405
-    function foo(x)
-        @ccall jl_typeof(Ref(x)::Ref{Float64})::Nothing
-        x + 1
-    end
-    @test autodiff(Reverse, foo, Active(1.0))[1][1] == 1.0
-end
-
 # TEST EXTENSIONS 
 using SpecialFunctions
 @testset "SpecialFunctions ext" begin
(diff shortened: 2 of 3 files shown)


vchuravy:
If you can reduce the test to not depend on DiffEq internals that would be better.

MasonProtter:
> If you can reduce the test to not depend on DiffEq internals that would be better.

No idea how to do that here unfortunately. Any suggestions? The stacktrace Enzyme gave didn't give too much info as to where / how it got tripped up on a `jltypeof`.

wsmoses:
honestly I'd just do a `ccall(:jl_typeof` or something, but yeah that code is auto emitted by julia's compiler codegen so its rather undefined if/when it will generate

MasonProtter:
Okay, I added a dummy `ccall` to the tests that I can confirm locally passes with this change, and fails without this change. Wasn't sure where exactly to put the test though, is there a better place for it than where I put it?

wsmoses:
what if we make a new test file called runtime_calls.jl and we can include it somewhere

wsmoses (approved):

PR #2415, opened 2025-05-26
Add warnings

wsmoses:
per a convo with @gbaraldi last week, it feels like adding a warning in advance is easier than telling folks word of mouth to try different setups in issues.

My hope would've been to just emit the warning when an error is thrown, but if something crashes (which to be clear should never happen) I'm not sure we have other recourse. Also it seems like folks just ignore the messages in errors like "set runtime activity" =/

### DIFF
--- a/ext/EnzymeDifferentiationInterfaceExt.jl
+++ b/ext/EnzymeDifferentiationInterfaceExt.jl
@@ -1,0 +1,10 @@
+module EnzymeDifferentiationInterfaceExt
+
+function __init__()
+    @warn """You are using DifferentiationInterface!"
+             DifferentiationInterface introduces interstitial wrappers that may limit the scope of input programs and add overhead."
+             This can cause derivatives to be slower, or fail to differentiate with default settings when they work with Enzyme directly (e.g. Enzyme.gradient instead of DI.gradient)."
+             If you find issues, please report at https://github.com/EnzymeAD/Enzyme.jl/issues/new and try Enzyme directly in the interim."""
+end
+
+end # module
--- a/src/init.jl
+++ b/src/init.jl
@@ -1,0 +1,9 @@
+
+function __init__()
+    if VERSION >= v"1.11.0"
+        @warn """You are using Julia v1.11 or later!"
+                 Julia 1.11 changes the default Array type to contain a triply-nested pointer, rather than a doubly nested pointer."
+                 This may cause programs (primal but especially derivatives) to be slower, or fail to differentiate with default settings when they previously worked on 1.10."
+                 If you find issues, please report at https://github.com/EnzymeAD/Enzyme.jl/issues/new and try Julia 1.10 in the interim."""
+    end
+end


wsmoses:
cc @gdalle re DI comment since I can't seem to add you as a reviewer

vchuravy:
Warnings during init are incredibly obnoxious. I agree with @gdalle that the most likely consequences is people just not using Enzyme. 

The 1.11 warning we could add to the readme

wsmoses:
Can we somehow register it to run during Julia’s custom segfault / etc error handler?

I agree and would’ve preferred it not run when necessary, but didn’t know a way to handle the segfault case otherwise

wsmoses:
Alternatively is there a better place to put it (this is just a copy of where Oceananigans has their 1.11 warning)

wsmoses:
My other worry is that if folks already ignore warnings in the errors we throw, they’ll definitely ignore them in a readme they won’t open

vchuravy:
Perhaps as an error hint? There is some experimental facility for that, but I haven't worked with that yet.

> My other worry is that if folks already ignore warnings in the errors we throw, they’ll definitely ignore them in a readme they won’t open

Those people are also the ones ignoring a global warning.

wsmoses:
> Perhaps as an error hint? There is some experimental facility for that, but I haven't worked with that yet.
> 

Like the low level systems thing I would do is set up a signal handler, which iirc julia does for itself (though not gc I suppose)?

> > My other worry is that if folks already ignore warnings in the errors we throw, they’ll definitely ignore them in a readme they won’t open
> 
> Those people are also the ones ignoring a global warning.

Yeah that's fair, but I really doubt people look at the readme.

wsmoses:
Another question is how do we warn people about slow perf. We won't throw an error, so we can't tack things onto that

gdalle:
Regardless of how you want to warn people, I'd much rather want to see where this slow perf is so that DI resolves it whenever possible. Do you have examples?

vchuravy:
Also sometimes people are okay with things being slow if they are more convenient. So I am much less concerned about that.

wsmoses:
I get that, though also people may only try things once and decide that it doesn't work for them (when in reality they were using a bad configuration) ... and given the fact that I've seen so many people hit the same runtime_activity error that comes from the triple pointer + broadcasting and/or closures that equally impacts differentiability and perf (and often mistake it as a "Segfault"), I want to try to give users a way to get the best configuration from the start

gdalle:
> people may only try things once and decide that it doesn't work for them (when in reality they were using a bad configuration)

But then again there are plenty of people who are empowered to try Enzyme _thanks to DI_. Forcing a warning at every load of Enzyme might just convince them, along with everyone else, that Enzyme is not robust enough for their use case.

Given how widely DI has been adopted, I don't think it is fair to call it "the bad configuration" anymore. What we can do, however, is improve error messages in more delicate touches ⬇️ 

> I've seen so many people hit the same runtime_activity error 

The latest version of DI includes a rather friendly and visible error message for that one. 

<details>

```julia
julia> using DifferentiationInterface

julia> using Enzyme: Enzyme

julia> function g(active_var, constant_var, cond)
           if cond
               return active_var
           else
               return constant_var
           end
       end;

julia> function h(active_var, constant_var, cond)
           return [g(active_var, constant_var, cond), g(active_var, constant_var, cond)]
       end;

julia> pushforward(
           h, AutoEnzyme(; mode=Enzyme.Forward), [1.0], ([1.0],), Constant([1.0]), Constant(true)
       )
ERROR: 
If you are using Enzyme by selecting the `AutoEnzyme` object from ADTypes, you may want to try setting the `mode` option as follows:

        AutoEnzyme(; mode=Enzyme.set_runtime_activity(Enzyme.Forward))
        AutoEnzyme(; mode=Enzyme.set_runtime_activity(Enzyme.Reverse))

This hint appears because DifferentiationInterface and Enzyme are both loaded. It does not necessarily imply that Enzyme is being called through DifferentiationInterface.

Constant memory is stored (or returned) to a differentiable variable.
As a result, Enzyme cannot provably ensure correctness and throws this error.
This might be due to the use of a constant variable as temporary storage for active memory (https://enzyme.mit.edu/julia/stable/faq/#Runtime-Activity).
If Enzyme should be able to prove this use non-differentable, open an issue!
To work around this issue, either:
 a) rewrite this variable to not be conditionally active (fastest, but requires a code change), or
 b) set the Enzyme mode to turn on runtime activity (e.g. autodiff(set_runtime_activity(Reverse), ...) ). This will maintain correctness, but may slightly reduce performance.
Mismatched activity for:   %value_phi = select i1 %.not, {} addrspace(10)* %1, {} addrspace(10)* %0, !dbg !15 const val: {} addrspace(10)* %1
 value=Unknown object of type Vector{Float64}
 llvalue={} addrspace(10)* %1
```

</details>

wsmoses:
my earlier comment was in reference to the triple pointer by 1.11 not a DI specific issue. my original comment on giving people an actionable way to debug performance in advance through a warning also had no reference to DI, but is something we generally need to figure out for potential bad configurations (be it Julia 1.11 array sadness, DI, or something else).

I was not calling DI a bad configuration, but referring to any bad configuration that could cause problems as something we want to create a warning for.

Also the "This hint appears because DifferentiationInterface and Enzyme are both loaded. It does not necessarily imply that Enzyme is being called through DifferentiationInterface." comment is kind of weird to have (also it might merit having the enzyme error at the start). 

I think we should get the backtrace version of the error in which would solve part of this PR and other issues for non-segfault errors.

The perf warning, and segfault error ofc still needs a solution

gdalle:
> Also the "This hint appears because DifferentiationInterface and Enzyme are both loaded. It does not necessarily imply that Enzyme is being called through DifferentiationInterface." comment is kind of weird to have

That was my take on [your request](https://github.com/JuliaDiff/DifferentiationInterface.jl/pull/788#discussion_r2082555528) to "make the error message more conservative because it will print even when DI is not used".

> (also it might merit having the enzyme error at the start).

That's defined in Enzyme, more specifically here:

https://github.com/EnzymeAD/Enzyme.jl/blob/bd607630f083b3c293a4a45eee45e6c001cd57d2/src/errors.jl#L166-L194

At the moment, `Base.Experimental.show_error_hints` is called before the Enzyme error is printed. You're free to move it after.

PR #2421, opened 2025-05-28
feat: add a stacked option to onehot

avik-pal:


avik-pal:
cc @wsmoses  for a review

ext/EnzymeGPUArraysCoreExt.jl
```diff
@@ -14,25 +14,29 @@ function Enzyme.zerosetfn!(x::AbstractGPUArray, i::Int, val)
...
     N = length(x)
-    ntuple(Val(N)) do i
+    ret = ntuple(Val(N)) do i
         Base.@_inline_meta
         res = zero(x)
         @allowscalar @inbounds res[i] = 1
         return res
     end
+    stacked isa Val{false} && return ret
+    return stack(ret)
```
wsmoses, line 27: we shold use a different implementation here. allocate a zero array of the whole size then just selectively set the one's in the identity (alternatively, is there already an eye or similar function we could use)

ext/EnzymeStaticArraysExt.jl
```diff
@@ -14,22 +14,26 @@ end
 
 @inline Enzyme.specialize_output(output, input::StaticArray) = convert(SArray, output)
 
-@inline function Enzyme.onehot(x::StaticArrays.SArray{S, T, N, L}) where {S, T, N, L}
-    ntuple(Val(L)) do i
+@inline function Enzyme.onehot(x::StaticArrays.SArray{S, T, N, L}; stacked::Union{Val{true}, Val{false}} = Val(false)) where {S, T, N, L}
```
wsmoses, line 17: I'm also debating if we should make this a proper non kw argument (essentially because while the outer function is marked inline, I don't know a away to make the inner kwfunc inline [and we really do want that here]

PR #2423, opened 2025-05-30
StrongZero + jll

wsmoses:


vchuravy (commented):
You use the term "correctness" which is tricky here, since both answers are correct, but one may be more helpful. For me, this sounds similar to admissibility issues with FiniteDiff. 

E.g. if one answer were more correct, than we should always use it ;)

Issue #2347, opened 2025-04-04
Precompilation fails on Julia 1.12 beta release


gdalle:
Now that 1.12 is available as a pre-release, I wanted to try it out with Enzyme and I stumbled upon this error. You're probably aware already but in case you're not, here it is:

```julia
(@v1.12) pkg> activate --temp
  Activating new project at `/var/folders/pt/myrxq2gs5hzcp79v5475_dh40000gp/T/jl_zOno71`

(jl_zOno71) pkg> add Enzyme
   Resolving package versions...
   Installed GPUCompiler ─ v1.3.2
  Installing artifacts ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 1/1
    Updating `/private/var/folders/pt/myrxq2gs5hzcp79v5475_dh40000gp/T/jl_zOno71/Project.toml`
  [7da242da] + Enzyme v0.13.35
    Updating `/private/var/folders/pt/myrxq2gs5hzcp79v5475_dh40000gp/T/jl_zOno71/Manifest.toml`
  [fa961155] + CEnum v0.5.0
  [7da242da] + Enzyme v0.13.35
  [f151be2c] + EnzymeCore v0.8.8
  [e2ba6199] + ExprTools v0.1.10
  [61eb1bfa] + GPUCompiler v1.3.2
  [692b3bcd] + JLLWrappers v1.7.0
  [929cbde3] + LLVM v9.2.0
  [d8793406] + ObjectFile v0.4.4
  [aea7be01] + PrecompileTools v1.3.0
  [21216c6a] + Preferences v1.4.3
  [189a3867] + Reexport v1.2.2
  [6c6a2e73] + Scratch v1.2.1
  [53d494c1] + StructIO v0.3.1
  [a759f4b9] + TimerOutputs v0.5.28
  [7cc45869] + Enzyme_jll v0.0.173+0
  [dad2f222] + LLVMExtra_jll v0.0.35+0
  [0dad84c5] + ArgTools v1.1.2
  [56f22d72] + Artifacts v1.11.0
  [2a0f44e3] + Base64 v1.11.0
  [ade2ca70] + Dates v1.11.0
  [f43a241f] + Downloads v1.6.0
  [7b1f6079] + FileWatching v1.11.0
  [b77e0a4c] + InteractiveUtils v1.11.0
  [ac6e5ff7] + JuliaSyntaxHighlighting v1.12.0
[...]

PR #2441, opened 2025-06-26
Start working towards Julia 1.12 compat

oscardssmith:
The main remaining blocker is that Enzyme.jl isn't using the new pass manager.


giordano:
For reference, this is related to #2347

oscardssmith:
@wsmoses is the 32 bit 1.10 segfault expected?

PR #2475, opened 2025-07-14
start newpm migration

oscardssmith:


src/compiler.jl
```diff
@@ -4373,7 +4375,16 @@ function GPUCompiler.compile_unhooked(output::Symbol, job::CompilerJob{<:EnzymeT
         permit_inlining!(f)
     end
 
-    LLVM.ModulePassManager() do pm
+    LLVM.@dispose pb=LLVM.NewPMPassBuilder() begin
```
wsmoses, line 4378: Let's start to begin with just the preservenvvm being moved over (and leave the others the same, to confirm)

vchuravy:
replaced by #2675

PR #2479, opened 2025-07-18
Implement stop to block gradient propagation

vchuravy:


lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -774,4 +774,14 @@ end
 
 Combined(mode::ReverseMode) = mode
 
+"""
+    stop(x)
+
+Detach the gradient of the value returned from `stop` from the gradient of `x`.
+"""
+@noinline function stop(x::T) where T
+    return Core.inferencebarrier(x)::T
```
vchuravy:
```suggestion
    return Base.inferencebarrier(x)::T
```

test/runtests.jl
```diff
@@ -88,6 +88,20 @@ include("typeunstable.jl")
 include("absint.jl")
 include("array.jl")
 
+import EnzymeCore: stop
+
+@test autodiff(Enzyme.Reverse, stop, Active(1.0)) == ((0.0,),)
+autodiff(Enzyme.Forward, stop, Duplicated(1.0, 2.0)) == (0.0,)
```
wsmoses, line 94: Cc @avik-pal who added something like this though perhaps with a different name

wsmoses: yeah we called it ignore_derivatives elsewhere

vchuravy:
Notebook, that shows how inactive didn't work  https://pluto.land/n/cwthy85w

vchuravy:
Decision from Enzyme meeting: This should be called `ignore_derivatives` and implemented in Enzyme proper.

PR #2497, opened 2025-07-31
fix nested codegen opt pipeline

wsmoses:


vchuravy:
What was the actual fix here? The code-diff is quite large

wsmoses:
we were running check_ir and related llvm transformations for ir out of codegen, but not from nested_codegen. the large diff is, unfortunately, as a result of refactoring the code to make a helper function that could be used in both scenarios without copying

PR #2498, opened 2025-07-31
fix precompile on 1.12

oscardssmith:
https://github.com/JuliaLang/julia/pull/58131 unifies the method table, so this is necessary to make Enzyme precompile on 1.12.

Should we do something to make sure CI succeeds in loading Enzyme on 1.12? (even if we don't expect tests to pass).


wsmoses:
yeah completely agreed

Issue #2518, opened 2025-08-18
Illegal loop cache type crash in `cacheForReverse`


penelopeysm:
Full error message: https://gist.github.com/penelopeysm/b106d30daf38a536913f69aa113532cc

This one also only crashes with reverse-mode on 1.11. It runs with either forward-mode and/or 1.10.

```julia
struct IL{T}
    idxs::T
end
struct VN{sym,T<:Union{typeof(identity),IL}}
    inner::T
    function VN{sym}(inner) where {sym}
        return new{sym,typeof(inner)}(inner)
    end
end
function Base.:(==)(x::VN{sym1}, y::VN{sym2}) where {sym1, sym2}
    return sym1 == sym2 && x.inner == y.inner
end
struct VI{Tmd,Tlogp}
    metadata::Tmd
    logp::Tlogp
end
struct TSVI{V}
    varinfo::V
end
struct MyMetadata{
    TIdcs<:Dict{<:VN,Int},
    TVal<:AbstractVector{<:Real},
}
    idcs::TIdcs
    vals::TVal
end
function g(vi, z)
    s = Vector{Float64}(undef, length(z))
    for i in eachindex(s)
        s[i] = vi.varinfo.metadata.s.vals[1]
    end
    return VI(vi.varinfo.metadata, 1.0)
end
params = [0.0]
function f(x)
    md_s = MyMetadata(
        Dict{VN,Int}(VN{:s}(IL((1,))) => 1),
        x,
    )
    vi = TSVI(VI((; s=md_s), 0.0))
    vi = g(vi, [1.5, 2.0])
    return vi.logp
end
f(params)

import Enzyme: Enzyme, Reverse, set_runtime_activity
Enzyme.gradient(set_runtime_activity(Reverse), f, params)
```

penelopeysm:
Minimised a bit more now. I'm not sure where to go from here, the next thing I tried was to replace `==` with `===` in the `Base.:(==)` method (https://github.com/JuliaLang/julia/blob/9615af0f269df4d371b8010e9507ed5bae86103b/base/Base.jl#L212) but replacing either of the `==` causes it to work.

```julia
struct E{T}
    e::T
end
struct D{sym,T<:E}
    d::T
end
function Base.:(==)(d1::D{sym1}, d2::D{sym2}) where {sym1,sym2}
    # changing either == to === makes it no longer crash
    return sym1 == sym2 && d1.d == d2.d
end
function f(x)  # x is unused now
[...]

PR #2533, opened 2025-08-27
Error handler fix

wsmoses:


wsmoses:
Fixes https://github.com/EnzymeAD/Enzyme.jl/issues/2518

PR #2543, opened 2025-08-30
1.11 assertion

wsmoses:


wsmoses:
@vchuravy @gbaraldi this looks like a bug in base julia?

specifically the assertion appears to occur outside of enzyme code?

PR #2545, opened 2025-08-30
add example for symmetric matrices

vchuravy:


docs/src/notebooks/structured.jl
```diff
@@ -0,0 +1,229 @@
...
+
+	function EnzymeRules.reverse(config, ::Const{typeof(Base.getindex)}, dret::Active, tape,
+	                 S::Duplicated{<:MySymmetric2}, i::Const, j::Const)
+		i = i.val
+		j = j.val
+		A = S.val
+		dA = S.dval
+		@inbounds if i == j
+        	dA.data[i, j] += dret.val
+    	elseif (A.uplo == 'U') == (i < j)
```
wsmoses, line 183: I'm not sure about the correctness of this generally

vchuravy: Yeah, which is why I haven't pushed this as a general rule for `Symmetric`, this notebook is more meant to explain to users what is happening.

Issue #2483, opened 2025-07-21
Error when installing Enzyme


StephenVavasis:
I received the following error message when installing Enzyme (with `] add Enzyme`):

```
LoadError: UndefVarError: `RNumber` not defined in `EnzymeCore`
```

My system is Windows 11, Julia version 1.11.6.  I was able to install Enzyme in a clean copy of Julia on the same system, so presumably something in my environment or previously downloaded packages is causing the issue.  But bisecting on my environment is a big undertaking, so maybe someone knowledgeable can suggest a shortcut to finding the problem.

In case it helps, here is the entire error message:
```
julia> err
PkgPrecompileError: The following 4 direct dependencies failed to precompile:

EnzymeLogExpFunctionsExt

Failed to precompile EnzymeLogExpFunctionsExt [66ee98d2-fe2d-5ae6-bd9b-9e5fe8b5b781] to "C:\\Users\\vavasis\\.julia\\compiled\\v1.11\\EnzymeLogExpFunctionsExt\\jl_1EBD.tmp".
ERROR: LoadError: UndefVarError: `RNumber` not defined in `EnzymeCore`
Stacktrace:
  [1] getproperty(x::Module, f::Symbol)
    @ Base .\Base.jl:42
  [2] top-level scope
    @ C:\Users\vavasis\.julia\packages\Enzyme\DL6fx\src\analyses\activity.jl:76
  [3] include(mod::Module, _path::String)
    @ Base .\Base.jl:562
  [4] include(x::String)
    @ Enzyme.Compiler C:\Users\vavasis\.julia\packages\Enzyme\DL6fx\src\compiler.jl:1
  [5] top-level scope
    @ C:\Users\vavasis\.julia\packages\Enzyme\DL6fx\src\compiler.jl:175
  [6] include(mod::Module, _path::String)
    @ Base .\Base.jl:562
  [7] include(x::String)
    @ Enzyme C:\Users\vavasis\.julia\packages\Enzyme\DL6fx\src\Enzyme.jl:1
  [8] top-level scope
    @ C:\Users\vavasis\.julia\packages\Enzyme\DL6fx\src\Enzyme.jl:127
  [9] include
    @ .\Base.jl:562 [inlined]
[...]

PR #2547, opened 2025-09-01
add support for ignore_derivatives

vchuravy:


src/Enzyme.jl
```diff
@@ -1560,6 +1560,14 @@ Returns true if within autodiff, otherwise false.
 """
 @inline EnzymeCore.within_autodiff() = false
 
+function ignore_derivatives(x::T) where T
+    if EnzymeCore.within_autodiff()
+        return ccall("extern __enzyme_ignore_derivatives", llvmcall, Any, (Any,), x) :: T
```
vchuravy: @wsmoses this is of course not ideal since it causes `x` to be boxed and unboxed.

src/Enzyme.jl
```diff
@@ -1560,6 +1560,14 @@ Returns true if within autodiff, otherwise false.
 """
 @inline EnzymeCore.within_autodiff() = false
```
vchuravy: This should be in EnzymeCore probably?

src/Enzyme.jl
wsmoses: is there a reason we can't have this be T, (T,) ?

vchuravy: One can, but it needs a generated function.

src/compiler.jl
```diff
@@ -2562,6 +2562,18 @@ function enzyme!(
...
+
+    if haskey(functions(mod), "__enzyme_ignore_derivatives")
+        ignore_derivatives = LLVM.functions(mod)["__enzyme_ignore_derivatives"]
+        for u in LLVM.uses(ignore_derivatives)
+            ci = LLVM.user(u)
+            @assert isa(ci, LLVM.CallInst)
+            LLVM.replace_uses!(ci, operands(ci)[2])
+            LLVM.remove!(ci)
+        end
+        LLVM.erase!(ignore_derivatives)
```
vchuravy: Something here is wrong, LLVM crashes after this has run.

wsmoses: It should be operands[1], right? The function arg is the last one

src/compiler.jl
```diff
@@ -2568,12 +2568,13 @@ function enzyme!(
             for u in LLVM.uses(fn)
                 ci = LLVM.user(u)
                 @assert isa(ci, LLVM.CallInst)
-                LLVM.replace_uses!(ci, operands(ci)[2])
-                LLVM.remove!(ci)
+                LLVM.replace_uses!(ci, operands(ci)[1])
+                LLVM.erase!(ci)
```
wsmoses, line 2621: You might need to collect the user instructions first since deleting the user may invalidate the iterator

lib/EnzymeCore/Project.toml
```diff
@@ -18,3 +19,4 @@ Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
 
 [weakdeps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
+Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
```
vchuravy: @giordano how bad is this as an idea?

I want to prevent the installation of EnzymeCore without a version of Enzyme that can handle the magical intrinsic call. 
Otherwise I think we can't implement `ignore_derivatives` in EnzymeCore, and must implement in Enzyme proper.

giordano: You only need this to be able to set the compat bound?  This might work.

vchuravy: Yeah! Registrator will hate this xD

wsmoses: Can't you upper bound the old enzyme to not work with the newer EnzymeCore?

giordano: At this point the new version of EnzymeCore could just be v0.9, since v0.8.9 was already breaking: https://github.com/EnzymeAD/Enzyme.jl/issues/2483#issuecomment-3240071589

wsmoses: ehh, yeah tht was definitely an error but it was only breaking between Enzyme <-> EnzymeCore, not between public facing features of EnzymeCore

wsmoses: [doing a breaking release of EnzymeCore is a massive pain that basically requires the whole ecosystem to update]

giordano: Does any other package require EnzymeCore?

vchuravy:
@giordano the integration tests are not using EnzymeCore from the repo?

.github/workflows/Integration.yml
```diff
@@ -56,12 +56,12 @@ jobs:
         with:
           version: ${{ matrix.version }}
       - uses: julia-actions/cache@v2
-      - uses: julia-actions/julia-buildpkg@v1
       - name: "Install Dependencies"
         run: |
             julia --color=yes --project=test/integration/${{ matrix.package }} --threads=auto --check-bounds=yes -O1 -e 'using Pkg; Pkg.develop([PackageSpec(; path) for path in (".", "lib/EnzymeCore")])'
         shell: bash
         if: ${{ matrix.version == '1.10' }}
+      - uses: julia-actions/julia-buildpkg@v1
```
giordano: I'm not sure this is much useful at all nowadays?

vchuravy: Yeah.

lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -620,6 +620,27 @@ Returns true if within autodiff, otherwise false.
...
+
+Behaves like the `identity` function, but disconnects the "shadow"
+associated with `x`. This has the effect of preventing any derivatives
+from being propagated through `x`.
+
+!!! compat "Enzyme 0.13.74"
+    Support for `ignore_derivatives` was added in Enzyme 0.13.74.
+"""
+@generated function ignore_derivatives(x::T) where {T}
+    name = "extern __enzyme_ignore_derivatives." * string(T)
```
wsmoses, line 634: does this need to be generated? I suppose so for having different names for different julia types

wsmoses:
the generate plots/benchmark CI also doesn't like the bump btw:
```
ERROR: LoadError: Unsatisfiable requirements detected for package EnzymeCore [f151be2c]:
 EnzymeCore [f151be2c] log:
 ├─possible versions are: 0.1.0 - 0.8.13 or uninstalled
 └─restricted to versions 0.8.14 - 0.8 by Enzyme [7da242da] — no versions left
   └─Enzyme [7da242da] log:
     ├─possible versions are: 0.13.81 or uninstalled
     └─Enzyme [7da242da] is fixed to version 0.13.81
````

lib/EnzymeCore/src/EnzymeCore.jl
vchuravy: Yeah that's the crux, we need different names since otherwise LLVM yells at us for having different types.

wsmoses: figured, in any case lgtm

PR #2551, opened 2025-09-04
Make sure that we pick up new definition of typetree

vchuravy:


src/typetree.jl
```diff
@@ -195,6 +195,16 @@ end
 
 const TypeTreeTable = IdDict{Any,Union{Nothing,TypeTree}}
 
+"""
+    typetree_total(job, T, ctx, dl, seen=TypeTreeTable())
+
+A wrapper around `typetree` that ensures the call happens in the correct world for GPUCompiler.
+Useful when using typetree from a generated function since typetree is user-extendable.
+"""
+function typetree_total(@nospecialize(job::GPUCompiler.CompilerJob), @nospecialize(T), ctx, dl, seen=TypeTreeTable())
```
wsmoses, line 204: For clarity could we call this in world instead of total?

PR #2552, opened 2025-09-05
Allow for mistmatch between native target and params for IPU

giordano:
Should fix https://github.com/JuliaIPU/IPUToolkit.jl/pull/73#issuecomment-3240056921

### DIFF
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -4272,7 +4272,10 @@ function GPUCompiler.compile_unhooked(output::Symbol, job::CompilerJob{<:EnzymeT
     primal_target = (job.config.target::EnzymeTarget).target
     primal_params = (job.config.params::EnzymeCompilerParams).params
     if primal_target isa GPUCompiler.NativeCompilerTarget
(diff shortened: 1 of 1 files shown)


src/compiler.jl
```diff
@@ -4272,7 +4272,10 @@ function GPUCompiler.compile_unhooked(output::Symbol, job::CompilerJob{<:EnzymeT
     primal_target = (job.config.target::EnzymeTarget).target
     primal_params = (job.config.params::EnzymeCompilerParams).params
     if primal_target isa GPUCompiler.NativeCompilerTarget
-        @assert primal_params isa PrimalCompilerParams 
+        if !(primal_params isa PrimalCompilerParams)
+            # XXX: This means mode is not propagated and rules are not applied for GPU code.
+            @safe_debug "NativeCompilerTarget without primal compiler params" primal_job
```
giordano: @vchuravy `primal_job` is not defined in this scope, should it be
```suggestion
            @safe_debug "NativeCompilerTarget without primal compiler params" primal_params
```
?

Issue #2397, opened 2025-05-16
[Julia 1.11] Broadcast fails with Const arguments


vchuravy:
Given a function like:

```
function lagrangian_elastic_rope(q, qdot, p)
         N = length(q)
         (;m, k, g) = p
         # kinetic energy is just 0.5*mass*velocity^2
         Eₖ = 0.5 * sum(m .* qdot.^2)
         # potential energy is the elastic energy of the springs plus the gravitational potential
         Eₚ = k[1] * q[1]^2
         for i in 2:N
                 Eₚ += k[i] * (q[i] - q[i-1])^2
         end
         Eₚ = 0.5*Eₚ - g * sum(m .* q) 
         return Eₖ - Eₚ # lagrangian = kinetic_energy - potential_energy
 end

function parameters_elastic_rope(N=3)
         # a helper to setup the parameters of the problem
     N = min(N, 5)
     # time domain
     t0   = 0
     tend = 50
     # initial condition
     q0    = [1.0, 2.0, 3.0, 4.0, 5.0]
     qdot0 = [0.2, 0.3, 0.7, 0.4, 0.5]
     q0, qdot0 = q0[1:N], qdot0[1:N]
     # parameters
     m = [1.1, 1.2, 1.3, 1.4, 1.5]
     k = [1.5, 1.3, 1.7, 1.9, 2.1]
     g = 1.
     m, k = m[1:N], k[1:N]
     params = (;m, k, g)
     return params, q0, qdot0, t0, tend
 end
p, q0, qdot0, t0, tend = parameters_elastic_rope(3)

lagrangian_elastic_rope(q0, qdot0, p)
```

I want to be able to take the derivative of `lagrangian_elastic_rope` solely w.r.t `q0`

```
dq = Enzyme.make_zero(q0)
autodiff(Reverse, lagrangian_elastic_rope, Active, Duplicated(q0, dq), Const(qdot0), Const(p))
dq
```

This used to work in 1.10, but fails in 1.11 with a:

```
ERROR: Constant memory is stored (or returned) to a differentiable variable.
As a result, Enzyme cannot provably ensure correctness and throws this error.
This might be due to the use of a constant variable as temporary storage for active memory (https://enzyme.mit.edu/julia/stable/faq/#Runtime-Activity).
If Enzyme should be able to prove this use non-differentable, open an issue!
To work around this issue, either:
 a) rewrite this variable to not be conditionally active (fastest, but requires a code change), or
[...]

Issue #2288, opened 2025-01-30
Enzyme on 1.11 needs more Cache than on 1.10 ?


yolhan83:
Hello, I encountered an issue in Julia 1.11. I hope it hasn't already been reported. Here is a minimal working example (MWE):
```
using Enzyme

function foo(x,y)
    y2 = y.^2
    return sum(x .+ y2[:])
end

x = rand(10)
y = rand(10)
dx = Enzyme.make_zero(x)
dy = Enzyme.make_zero(y)
Enzyme.autodiff(Enzyme.Reverse, foo, Enzyme.Duplicated(x,dx), Enzyme.Const(y))
```
This works fine on 1.10, however, on 1.11, I get 
```
LoadError: Constant memory is stored (or returned) to a differentiable variable.
As a result, Enzyme cannot provably ensure correctness and throws this error.
This might be due to the use of a constant variable as temporary storage for active memory (https://enzyme.mit.edu/julia/stable/faq/#Runtime-Activity).
If Enzyme should be able to prove this use non-differentable, open an issue!
To work around this issue, either:
 a) rewrite this variable to not be conditionally active (fastest, but requires a code change), or
 b) set the Enzyme mode to turn on runtime activity (e.g. autodiff(set_runtime_activity(Reverse), ...) ). This will maintain correctness, but may slightly reduce performance.
Mismatched activity for:   %38 = phi {} addrspace(10)* [ %29, %L90 ], [ %573, %guard_exit125 ] const val:   %573 = load {} addrspace(10)*, {} addrspace(10)* addrspace(11)* %572, align 8, !dbg !468, !tbaa !104, !alias.scope !40, !noalias !43, !dereferenceable_or_null !221, !align !369, !enzyme_type !68, !enzymejl_source_type_Memory\7BFloat64\7D !0, !enzymejl_byref_MUT_REF !0
 value=Unknown object of type Memory{Float64}
 llvalue=  %573 = load {} addrspace(10)*, {} addrspace(10)* addrspace(11)* %572, align 8, !dbg !468, !tbaa !104, !alias.scope !40, !noalias !43, !dereferenceable_or_null !221, !align !369, !enzyme_type !68, !enzymejl_source_type_Memory\7BFloat64\7D !0, !enzymejl_byref_MUT_REF !0

Stacktrace:
  [1] ==
    @ .\promotion.jl:639
  [2] !=
    @ .\operators.jl:277
  [3] _newindexer
[...]

PR #2555, opened 2025-09-11
Fixup memory instance

wsmoses:
mildly crazy to say the least, but should fix the broadcast runtime activity issues like https://github.com/EnzymeAD/Enzyme.jl/issues/2454

confirmed fix for https://github.com/EnzymeAD/Enzyme.jl/issues/2397
confirmed fix for https://github.com/EnzymeAD/Enzyme.jl/issues/2288

### DIFF
--- a/src/compiler/interpreter.jl
+++ b/src/compiler/interpreter.jl
@@ -324,20 +324,12 @@ end
 
 Core.Compiler.method_table(@nospecialize(interp::EnzymeInterpreter)) = interp.method_table
 
-function override_bc_materialize end
-
 function is_alwaysinline_func(@nospecialize(TT))::Bool
     isa(TT, DataType) || return false
     @static if VERSION ≥ v"1.11-"
     if TT.parameters[1] == typeof(Core.memoryref)
         return true
     end
-    end
-    if TT.parameters[1] == typeof(Base.materialize) &&  length(TT.parameters) == 2 && TT.parameters[2] <: Base.Broadcast.Broadcasted
-	return true
-    end
-    if TT.parameters[1] == typeof(override_bc_materialize)
-        return true
     end
     return false
 end
@@ -983,7 +975,10 @@ end
 
 Base.@propagate_inbounds @inline overload_broadcast_getindex(A, I) = @inbounds A[I]
 
-@inline function override_bc_materialize(bc, ElType::Type)
+struct OverrideBCMaterialize{ElType}
+end
+
+@inline function (::OverrideBCMaterialize{ElType})(bc) where ElType
     if bc.args isa Tuple{AbstractArray} && bc.f === Base.identity
         return copy(bc.args[1])
     end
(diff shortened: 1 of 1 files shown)


wsmoses:
tangentially @mhauru the bijector case here is broken for unclear reasons, can you isolate a reproducer?

src/compiler/interpreter.jl
```diff
@@ -868,56 +868,121 @@ end
...
     for i in 1:length(x.parameters)
-        if x.parameters[i] <: Number
+	push!(todo, (:(x[$i]), x.parameters[i]))
+    end
+    while length(todo) != 0
+	expr, ty = pop!(todo)
+        if ty <: Number || ty <: Base.RefValue
             continue
         end
+	if ty <: Base.Broadcast.Broadcasted{<:Base.Broadcast.DefaultArrayStyle, Nothing}
```
vchuravy, line 892: This seems very sketchy? What is this doing?

src/llvm/transforms.jl
```diff
@@ -2297,6 +2297,33 @@ function checkNoAssumeFalse(mod::LLVM.Module, shouldshow::Bool = false)
     end
 end
 
+function rewrite_generic_memory!(mod::LLVM.Module)
```
vchuravy: What is this function doing?

test/runtests.jl
```diff
@@ -2051,6 +2054,9 @@ end
     dA = zero(A); dB = zero(B); dR = fill!(similar(A), 1)
 
     autodiff(Reverse, foo_bc!, Const, Duplicated(A, dR), Duplicated(transpose(A), transpose(dA)), Duplicated(B, dB))
+
+    # no runtime activity required
+    autodiff(Forward, f, Duplicated(rand(3), rand(3)), Const(rand(3)))
```
giordano: As is, this is only checking that the function call doesn't error out badly, but it'd be useful to check also that the result is correct.

src/compiler/interpreter.jl
wsmoses: recursing through all arrays in a broadcast to check if they are all the same size

src/llvm/transforms.jl
wsmoses: the pointer to the memory instance is checked whether or not it is null, creating phis. this just replaces the load of the memory instance itself with the value

wsmoses:
> ncomfortable in hard coding behavior for types with user specified behavior like
> 
> `Base.Broadcast.Broadcasted{<:Base.Broadcast.DefaultArrayStyle, Nothing}`
> 
> What happ

there is an explicit check that one of the args of the broadcast is an array

src/llvm/transforms.jl
vchuravy: Can you add this as a comment? Is this safe under push!/pop!?

Is this fix independent of the broadcast changes above?

src/errors.jl
```diff
@@ -524,6 +524,20 @@ function julia_error(
...
+                    larg, off = get_base_and_offset(operands(cur)[1]; inst=ncur, inttoptr=true)
+		    if isa(larg, LLVM.ConstantInt) && off == sizeof(Int)
+			ptr = reinterpret(Ptr{Cvoid}, convert(UInt, larg))
+			obj = Base.unsafe_pointer_to_objref(ptr)
+                        if obj isa Memory && obj == typeof(obj).instance
+                            return make_batched(ncur, prevbb)
+                        end
+		    end
+                end
+end
```
vchuravy: Please format the code

giordano:
> In a dynamic language like Julia it is not guaranteed that the implementation that was written in Base is the one that ends up being executed by the user. The hijack of the broadcast is in my opinion not sound. Leaving this aside, could you format your changes (and add some comments to the tricky bits) so that it is easier for me to follow your logic?

Just to provide an example of the above, with this PR, using the examples in https://github.com/EnzymeAD/Enzyme.jl/issues/2454#issue-3203158199
```julia-repl
julia> using Enzyme, FixedSizeArrays

julia> f(x, y) = sum(sin.(x) .^ 2 + cos.(y) .^ 2)

julia> gradient(Forward, f, rand(3), rand(3))
([0.9329777780991535, 0.6950736925275953, 0.9960966881297458], [-0.3250275243138287, -0.4228412074950716, -0.9999926119919079])

julia> gradient(Forward, f, FixedSizeVector(rand(3)), FixedSizeVector(rand(3)))
ERROR: Constant memory is stored (or returned) to a differentiable variable.
[...]
```
So this PR would fix broadcasting of `Array`s but not a custom container like `FixedSizeArray`, because [it uses a custom broadcast style](https://github.com/JuliaArrays/FixedSizeArrays.jl/blob/0587d03a31844ac284d15a8142f8f9236b5b9a3d/src/FixedSizeArray.jl#L215-L237)

wsmoses:
yeah that's qutie intentional, this should not touch broadcasting of other arrays (e.g. gpu arrays). still necessary/helpful here, but the general case requires actually going through the giant amount of memory nonsense from https://github.com/EnzymeAD/Enzyme.jl/issues/2562

vchuravy:
FixedSizeArray should be the old case since  the outer container is an immutable and thus one reference less

gdalle:
Just FYI @wsmoses this PR broke the custom error hint I had added to DI for runtime activity errors, for reasons that I explain in https://github.com/JuliaDiff/DifferentiationInterface.jl/pull/852. No one is at fault here except the weird design of error hints

PR #2556, opened 2025-09-11
Fix compile time bug

wsmoses:
Before
CUDA CI on 1.10 : Reverse Kernel |    3      3  13m28.7s
CUDA CI on 1.11 : Reverse Kernel |    3      3  15m27.7s

After
CUDA CI on 1.10: Reverse Kernel |    3      3  2m27.1s
CUDA CI on 1.11 : Reverse Kernel |    3      3  4m00.7s

Obviously still work to be had, but a lot less stupid at least

### DIFF
--- a/src/compiler/interpreter.jl
+++ b/src/compiler/interpreter.jl
@@ -138,14 +138,26 @@ struct EnzymeInterpreter{T} <: AbstractInterpreter
     handler::T
 end
 
-
+const SigCache = Dict{Tuple, Dict{UInt, Base.IdSet{Type}}}()
 function get_rule_signatures(f, TT, world)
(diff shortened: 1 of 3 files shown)


src/compiler/interpreter.jl
```diff
@@ -138,14 +138,26 @@ struct EnzymeInterpreter{T} <: AbstractInterpreter
     handler::T
 end
 
-
+const SigCache = Dict{Tuple, Dict{UInt, Base.IdSet{Type}}}()
```
vchuravy, line 141: Could this be part of the interp?

wsmoses: no becuase the bottleneck is computing the cache_key and equivalent to call the constructor of the interp

PR #2559, opened 2025-09-11
docs: improve runtime activity FAQ and fix MWE

gdalle:
The function in the MWE was missing an `end`, so I made it executable to prevent this kind of bug (I didn't go for a full doctest because of the exception stack trace)

### DIFF
--- a/docs/src/faq.md
+++ b/docs/src/faq.md
@@ -313,7 +313,7 @@ function h(cond, active_var, constant_var)
 end
 
 try  #hide
-Enzyme.autodiff(Forward, h, Const(cond), Duplicated(x, dx), Const(y))
+Enzyme.autodiff(Forward, h, Const(condition), Duplicated(x, dx), Const(y))
 catch err; showerror(stderr, err); end  #hide
 ```
 
@@ -330,7 +330,7 @@ Enabling runtime activity does therefore, come with a sharp edge, which is that 
 Generally, the preferred solution to these type of activity unstable codes should be to make your variables all activity-stable (e.g. always containing differentiable memory or always containing non-differentiable memory). However, with care, Enzyme does support "Runtime Activity" as a way to differentiate these programs without having to modify your code. One can enable runtime activity for your code by changing the mode, such as
 
 ```@example runtime
-dout, out = Enzyme.autodiff(Enzyme.set_runtime_activity(ForwardWithPrimal), g, Const(cond), Duplicated(x, dx), Const(y))
+dout, out = Enzyme.autodiff(Enzyme.set_runtime_activity(ForwardWithPrimal), g, Const(condition), Duplicated(x, dx), Const(y))
 ```
 
 However, care must be taken to check derivative aliasing afterwards:


gdalle:
Running into https://github.com/JuliaDocs/Documenter.jl/issues/1420, will fix

docs/src/faq.md
```diff
@@ -284,26 +284,37 @@ When computing the derivative of mutable variables, Enzyme also needs additional
...
+        return active_var
+    else
+        return constant_var
+    end
 end
 
-Enzyme.autodiff(Forward, g, Const(condition), Duplicated(x, dx), Const(y))
+x, dx = [1.0], [2.0];
+y = [3.0];
+cond = false;  # return the constant variable
```
wsmoses: let's still call it condition to not confuse the name of the function parameter from the input to the call

docs/src/faq.md
```diff
@@ -312,14 +323,20 @@ If one created a new zero'd copy of each return from `g`, this would mean that t
...
-```julia
-Enzyme.autodiff(set_runtime_activity(Forward), h, Const(condition), Duplicated(x, dx), Const(y))
+```@example runtime
+dout, out = Enzyme.autodiff(Enzyme.set_runtime_activity(ForwardWithPrimal), g, Const(cond), Duplicated(x, dx), Const(y))
+```
+
+However, care must be taken to check derivative aliasing afterwards:
+
+```@example runtime
+dout === out  # if true, the actual derivative is zero
```
wsmoses: add the comment "and pointer-like"

gdalle: I wanted to ask: what is the precise definition of "pointer-like" in Julia terms, and what does it mean for two objects to be aliased in that way? I thought that it meant either `a === b` or `pointer(a) == pointer(b)`, as your docs suggest, but I found out that these conditions are not equivalent, and that the latter can only be checked for `Array`s and a few other concrete types. See https://github.com/JuliaDiff/DifferentiationInterface.jl/issues/666#issuecomment-3279224954 for details and https://github.com/JuliaDiff/DifferentiationInterface.jl/pull/850 for my current attempt in DI.

wsmoses: whether the layout of the julia object in memory is a pointer

gdalle: Okay, then what is the proper way to test aliasing for Enzyme's runtime activity feature in general? Because `===` seems wrong based on my tests and `pointer` works for a very limited set of types

wsmoses: not sure offhand julia-side, I'm guessing there's some feature for determining type layout

gdalle: I tagged you on a Slack exchange about this, hopefully it helps. If I don't have that information I can't really implement the proper check for runtime activity inside DI (nor can any other Enzyme user outside of DI for that matter)

wsmoses: in any case can you add the comment here and we can merge

wsmoses: gentle bump @gdalle

vchuravy:
```suggestion
dout === out  # if true, the actual derivative is zero and pointer-like
```

docs/src/faq.md
```diff
@@ -284,26 +284,37 @@ When computing the derivative of mutable variables, Enzyme also needs additional
...
+    end
 end
 
-Enzyme.autodiff(Forward, g, Const(condition), Duplicated(x, dx), Const(y))
+x, dx = [1.0], [2.0];
+y = [3.0];
+cond = false;  # return the constant variable
+try  #hide
+Enzyme.autodiff(Forward, g, Const(cond), Duplicated(x, dx), Const(y))
+catch err; showerror(stderr, err); end  #hide
```
vchuravy:
```suggestion
x, dx = [1.0], [2.0];
y = [3.0];
condition = false;  # return the constant variable
try  #hide
Enzyme.autodiff(Forward, g, Const(condition), Duplicated(x, dx), Const(y))
catch err; showerror(stderr, err); end  #hide
```

docs/src/faq.md
wsmoses: not quite, its "if true and pointer-like, the actual derivative is zero"

docs/src/faq.md
```diff
@@ -312,14 +323,20 @@ If one created a new zero'd copy of each return from `g`, this would mean that t
...
-```julia
-Enzyme.autodiff(set_runtime_activity(Forward), h, Const(condition), Duplicated(x, dx), Const(y))
+```@example runtime
+dout, out = Enzyme.autodiff(Enzyme.set_runtime_activity(ForwardWithPrimal), g, Const(cond), Duplicated(x, dx), Const(y))
+```
+
+However, care must be taken to check derivative aliasing afterwards:
+
+```@example runtime
+dout === out  # if true, the actual derivative is zero and pointer-like
```
vchuravy:
```suggestion
dout === out  # if true and pointer-like, the actual derivative is zero 
```

docs/src/faq.md
vchuravy: One way would be `Base.mightalias`, but it is allowed to have false positives.

gdalle: In this setting, false positives are as bad as false negatives: we don't want to replace a derivative with zero if we don't have to, but we also don't want to return the primal as a derivative.
To me it seems like users of Enzyme's runtime activity feature really have no way of performing the check you want them to perform, since even you two don't know how to. Is there a way out of this? Maybe Enzyme itself could return a flag that tells you whether or not each output is aliased with its shadow?

gdalle: For instance, in this case, the check `a === b` doesn't suffice to detect aliasing:

```julia
julia> x = [1.0];

julia> mutable struct S
           x::Vector{Float64}
       end

julia> a = S(x);

julia> b = S(x);

julia> a === b
false
```

wsmoses: `convert(LLVMType, julia_type; allow_boxed=true) isa LLVM.PointerType` ought work always but not an answer you want @gdalle

wsmoses: `Base.DataTypeFieldDesc` may also be helpful here

docs/src/faq.md
```diff
@@ -312,14 +323,20 @@ If one created a new zero'd copy of each return from `g`, this would mean that t
...
 
 Enabling runtime activity does therefore, come with a sharp edge, which is that if the computed derivative of a function is mutable, one must also check to see if the primal and shadow represent the same pointer, and if so the true derivative of the function is actually zero.
 
 Generally, the preferred solution to these type of activity unstable codes should be to make your variables all activity-stable (e.g. always containing differentiable memory or always containing non-differentiable memory). However, with care, Enzyme does support "Runtime Activity" as a way to differentiate these programs without having to modify your code. One can enable runtime activity for your code by changing the mode, such as
 
-```julia
-Enzyme.autodiff(set_runtime_activity(Forward), h, Const(condition), Duplicated(x, dx), Const(y))
+```@example runtime
+dout, out = Enzyme.autodiff(Enzyme.set_runtime_activity(ForwardWithPrimal), g, Const(cond), Duplicated(x, dx), Const(y))
```
giordano: https://github.com/EnzymeAD/Enzyme.jl/actions/runs/17741896546/job/50420085337#step:7:21
```
┌ Error: failed to run `@example` block in docs/src/faq.md:332-334
│ ```@example runtime
│ dout, out = Enzyme.autodiff(Enzyme.set_runtime_activity(ForwardWithPrimal), g, Const(cond), Duplicated(x, dx), Const(y))
│ ```
│   exception =
│    UndefVarError: `cond` not defined in `Main.__atexample__named__runtime`
│    Suggestion: check for spelling errors or missing imports.
│    Hint: a global variable of this name may be made accessible by importing LinearAlgebra in the current active module Main
```

docs/src/faq.md
gdalle: > convert(LLVMType, julia_type; allow_boxed=true) isa LLVM.PointerType ought work always but not an answer you want @gdalle

How am I supposed to use it?

```julia
julia> using LLVM

julia> convert(LLVMType, Vector{Float64}; allow_boxed=true)
ERROR: No LLVM context is active
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] context(; throw_error::Bool)
   @ LLVM ~/.julia/packages/LLVM/UFrs4/src/state.jl:22
 [3] context
   @ ~/.julia/packages/LLVM/UFrs4/src/state.jl:20 [inlined]
 [4] convert(::Type{LLVMType}, typ::Type; allow_boxed::Bool)
   @ LLVM.Interop ~/.julia/packages/LLVM/UFrs4/src/interop/base.jl:73
 [5] top-level scope
   @ REPL[25]:1
```

> Base.DataTypeFieldDesc may also be helpful here

I don't know what this does and it seems private?

```julia
help?> Base.DataTypeFieldDesc
  │ Warning
  │
  │  The following bindings may be internal; they may change or be removed in future versions:
  │
  │    •  Base.DataTypeFieldDesc

  No documentation found for private symbol.
```

Besides, users would also need a way to compare the locations to which these "pointer-like" objects point, in order to detect aliasing

gdalle: I'd be happy if Enzyme provided `Enzyme.is_pointerlike(x::Any)` and `Enzyme.are_aliased(x::T, y::T) where {T}` with the precise semantics that you want, to use in runtime activity correctness checking. Expecting the users to go through this level of sophistication seems doomed to fail

Have opened https://github.com/EnzymeAD/Enzyme.jl/issues/2587 to keep track so that we may merge this docs fix

PR #2563, opened 2025-09-11
Update macOS version in CI workflow

wsmoses:


giordano:
Uhm, this was changed in #2273 because jobs were timing out. Using x86_64 builds of Julia on apple silicon via Rosetta is very unsupported and error-prone, and also github won't allow it anymore at some point: https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/#notice-of-macos-x86_64-intel-architecture-deprecation.

I think the best path forward is to simply use the native architecture everywhere in all the jobs, I'll open a pull request to that effect in a few minutes.

PR #2582, opened 2025-09-16
Make 32-bit tests pass

wsmoses:


test/blas.jl
```diff
@@ -3,6 +3,8 @@ using EnzymeTestUtils
 using LinearAlgebra
 using Test
 
+# LLVM ERROR: out of memory LLVMWriteBitcodeToMemoryBuffer
+if sizeof(Int) != sizeof(Int64)
```
giordano, line 7: Alternatively
```suggestion
if Sys.WORD_SIZE != 64
```
This is very similar to the test you wrote ([`Sys.WORD_SIZE` is based on the size of `Int`](https://github.com/JuliaLang/julia/blob/7fa26ce89dc09596e865123afe37fe5a5cd4bd9a/base/sysinfo.jl#L105)), but I think the intent is more explicit

test/internal_rules.jl
```diff
@@ -776,6 +776,8 @@ function test_sparse(M, v, α, β)
     test_reverse(LinearAlgebra.mul!, Const, (C, Const), (M, Const), (v, Const), (α, Active), (β, Active))
 end
 
+# LLVM ERROR: out of memory _ZN22MyAttributorLegacyPass11runOnModuleERN4llvm6ModuleE at /home/wmoses/.julia/artifacts/61617502a39254b2b056f7aa53ec9b80726eec6b/lib/libEnzyme-16.so (unknown line)
+if sizeof(Int) == sizeof(Int64)
```
giordano, line 780:
```suggestion
if Sys.WORD_SIZE == 64
```

vchuravy:
replaced by #2622

PR #2585, opened 2025-09-17
refactor: move SparseArrays into an extension

avik-pal:


avik-pal:
Documenter CI seems to have started failing due to (happening Reactant side as well):

```
OpenSSL version mismatch. Built against 30000100, you have 3050003f
fatal: Could not read from remote repository.
```

https://github.com/EnzymeAD/Enzyme.jl/actions/runs/17784875656/job/50550622672?pr=2585#step:7:67. One solution is to pin OpenSSL (https://github.com/LuxDL/Lux.jl/commit/a4b61d95581e3010f3b20d3bdacc373aeed251cc#diff-6e0adb5e01dce01db395f52201532ad8fec022922489f34827886053953ce93fR67)

src/absint.jl
```diff
@@ -600,7 +602,7 @@ function abs_typeof(
             # add the extra poitner offset when loading here]. However for pointers constructed by ccall outside julia
             # to a julia object, which are not inline by type but appear so, like SparseArrays, this is a problem
             # and merits further investigation. x/ref https://github.com/EnzymeAD/Enzyme.jl/issues/2085
-            if !Base.allocatedinline(typ) && typ != SparseArrays.cholmod_dense_struct && typ != SparseArrays.cholmod_sparse_struct && typ != SparseArrays.cholmod_factor_struct
```
wsmoses, line 611: unfortunately presently this cannot be moved into an extension (it will not get picked up from generated function shenanigans). the longer term soltn i need to figure out is how to handle this specific edge case of a ccall return, but I haven't gotten to so yet [and it only came up for sparsearrays]

vchuravy: You would need to perform an invoke in world (probably invole_in_world_total) with the world being the job.world

PR #2601, opened 2025-09-21
[CI] Use `default` architecture instead of forcing `x64`

giordano:
This is in particular relevant on `macOS-latest`, where `default` would be `aarch64`: currently we're running the tests with x86-64 builds of Julia on Apple Silicon via Rosetta, which is slow, unsupported, and likely to fail.  Ref: https://github.com/EnzymeAD/Enzyme.jl/pull/2563#issuecomment-3316103867

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -48,13 +48,13 @@ jobs:
           - macOS-latest
           - windows-latest
         arch:
-          - x64
+          - default
         assertions:
           - false
         libEnzyme: [local, packaged]
         exclude:
           - os: windows-latest
-            arch: x64
+            arch: default
             libEnzyme: local
         include:
           - os: ubuntu-24.04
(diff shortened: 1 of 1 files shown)


giordano:
> which is slow

As a concrete example:
* [Julia 1.10 - macOS-latest - x64 - packaged libEnzyme - assertions=false - push](https://github.com/EnzymeAD/Enzyme.jl/actions/runs/17895609357/job/50881820072#logs) job on `main`: 22m 34s
* [Julia 1.10 - macOS-latest - default - packaged libEnzyme - assertions=false - pull_request](https://github.com/EnzymeAD/Enzyme.jl/actions/runs/17896350080/job/50883484596?pr=2601) job in this PR: 14m 43s.

wsmoses (approved):

wsmoses:
do you mind making the same change in https://github.com/EnzymeAD/Enzyme/blob/main/.github/workflows/enzyme-julia.yml

PR #2607, opened 2025-09-22
rewrite make_zero! and remake_zero! with ntuple over fields

vchuravy:


src/typeutils/make_zero.jl
```diff
@@ -572,13 +572,13 @@ end
         return nothing
     end
     push!(seen, prev)
-    for i in 1:nf
+    ntuple(Val(nf)) do i
```
wsmoses, line 575: honestly this feels like it should become a generated function (and we can also therefore conditionally create the seen dict if there is > 1 potentially aliasing piece of data)

vchuravy: `ntuple` is a generated function, just one that is limited in scope. I still prefer this over the headache that generated function bring with them otherwise.

I also don't think it would be unreasonable for a user to provide their own `make_zero` function / `remake_zero!` function.

wsmoses: yeah we definitely want user definable overrides. I'm just dreaming of a world where we don't need the explicit specialization for make_zero(Array) in addition to make_zero(Array, seen).

in any case we can figure this out later, this is already clearly a good change

vchuravy: Also sadly while we might be able to use a generated function, that would only allow us to guard against recursion in the type-domain and not recursion in the value domain :/

wsmoses: Perhaps but things like make_zero of a tuple of floats is sufficiently common that we really want it to use the right thing without the is check

PR #2608, opened 2025-09-22
add some benchmarks for remake_zero!

vchuravy:
(cherry picked from commit f19b0409352a2f3a9e646667709c98cc23ba59b1)

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -1,6 +1,6 @@
 # To run:
 # using PkgBenchmark, Enzyme
-# result = benchmarkpkg(KernelAbstractions)
+# result = benchmarkpkg(Enzyme)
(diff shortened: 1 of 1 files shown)


vchuravy:
Since CI for https://github.com/EnzymeAD/Enzyme.jl/pull/2607 will only pick it up from master. 

Any other ideas instead of just adding deeper and deeper nestings?

wsmoses:
so my thought on how to do this properly is we should have a function make_zero_internal or something which is the user definable one (or alternatively just make_zero(x, ::Union{Nothing, IdSet}))

then the generated function for make_zero(x) has a back edge from all the other internal utilities, and does a recursion through type space until it hits primitives like above, or guaranteed_const types. Types that are active don't need the id check, only duplicated

PR #2613, opened 2025-09-22
Add 1.12 to CI

vchuravy:


vchuravy:
Split the fix here out to #2654

wsmoses:
@gdalle the current DI version in CI here doesn't support 1.12. Can you open an independent PR bumping to a version which supports 1.12?

gdalle:
What do you mean by "it doesn't support Julia 1.12"?

gdalle:
There's no upper bound on Julia compat in DI

giordano:
It's JET, a dependency of DifferentiationInterfaceTest, which fails to precompile: https://github.com/EnzymeAD/Enzyme.jl/actions/runs/19831329281/job/56817898039?pr=2613#step:7:262

gdalle:
Oh right. I'm AFK right now but you can just update to the latest version of DI and DIT, that will do it

PR #2614, opened 2025-09-22
Mark x86 CI as xfail

wsmoses:


vchuravy (approved):

.github/workflows/CI.yml
```diff
@@ -62,6 +62,7 @@ jobs:
             libEnzyme: packaged
             version: '1.10'
             assertions: false
+            fail_ci_if_error: false
```
giordano: I don't think this does anything. On the runtest step you need to add [`continue-on-error`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepscontinue-on-error) with a condition matching this job

wsmoses: ah sigh ok

giordano (approved):

PR #2615, opened 2025-09-22
Fix ci

wsmoses:


test/runtests.jl
```diff
@@ -606,6 +606,7 @@ end
 end
 
 @testset "Simple Exception" begin
+    f_simple_exc(x, i) = ccall(:jl_, Cvoid, (Any,), x[i])
```
giordano: Does this need to use a function which prints, or can be any ccall? In the former case, can you at least add a comment to explain the motivation? It's hard to tell otherwise that the print is needed.

giordano: Likewise in https://github.com/EnzymeAD/Enzyme.jl/blob/724e6e48249fffbce826a40ff1da9e3a5d46b870/test/runtests.jl#L3430 (indentation is off BTW)

wsmoses: this should never actually get to the print though, as it should throw the error before the print?

But the actual issue is that it needs to be a function which uses the result in a differentiably active way (we mark jl_ as special and being active). That said in this case here we likely could get away with just
```
f_simple_exc(x, i) = x[i]
```

wsmoses: okay looking more closely at the other jl_ I think we may be able to get rid of it. If the input was Const, or otherwise necessary as a test of activity [jl_ is the simplest function which forces activity], then we would want to keep. but since the input is duplicated for that arg, actually we can get away with it

PR #2630, opened 2025-09-29
Foldl broadcast speedup

wsmoses:


wsmoses:
locally this results in a ~30x speedup so I'm going to go ahead and merge/tag.

@vchuravy we can talk about this over the week in post commit review

PR #2643, opened 2025-10-02
Add world to EnzymeContext

vchuravy:
- **access enzyme_context through gutils**
- **assert check the world we encode in enzyme_extrace_world**

### Update (2026-09-08)

Rebased onto main and squashed into one commit: the sixteen commits were fixups of each other, and main has moved on by 143 commits (custom rules are now compiled natively through `invoke_codegen!`, which takes the world explicitly; its calls into `nested_codegen!` now take the world from the context, as this PR intends). `rules/rules`, `rules/rule_inlining`, `rules/rrules`, `rules/ruleinvalidation`, `basic`, `errors` pass on Julia 1.12.

---

**2026-09-08:** review addressed in 4987dacd. `enzymejl_world` is removed entirely: nothing writes it, the ABI wrappers no longer copy it, and `enzyme_extract_world` is gone. `setup_global`/`unsafe_to_llvm` and every helper reaching them take an explicit `world::UInt` from the caller (rules: `enzyme_context(gutils).world`; codegen: the job's world; type-analysis handlers: `enzyme_context(TypeAnalyzer(data))`). Enzyme's C callbacks have no context argument, so `enzyme!` publishes the job's world in task-local storage for the duration of the C++ call and `callback_world()` reads it; that is the one remaining seam, to be removed once the C API hands the callbacks a context. `emit_methodinstance!` is back with an explicit world argument. `gradientutils.jl` is only included in module `Enzyme` now, so its LLVM helpers are imported explicitly (this was the `typeunstable` failure).


src/Enzyme.jl
```diff
@@ -133,6 +133,7 @@ Base.convert(::Type{API.CDerivativeMode}, ::ForwardMode) = API.DEM_ForwardMode
 function guess_activity end
 
 mutable struct EnzymeContext
+    world::UInt64
```
wsmoses, line 136: this should already be accessible everywhere gutils is available per enzyme_extract_world(fn) or something

vchuravy: This is meant as a proof of concept to demonstrate that yes we can get the world we are compiling from everywhere. 

Using a string attribute is fine, but is also a blocker for caching the IR for later use, so I wanted to see if we can use this to avoid adding ephemeral data to potentially cacheable inputs.

src/errors.jl
```diff
@@ -471,6 +471,8 @@ function julia_error(
             ) #=error=#
             world = enzyme_extract_world(f)
         end
+        # TODO: get world from TypeAnalyzer
+        # @assert world == enzyme_gutils_context(gutils).world
         throw(IllegalTypeAnalysisException(msg, mi, world, sval, ir, bt))
```
vchuravy: @wsmoses would you be okay with a change where `CustomErrorHandler` always receives `gutils`? We can go from `gutils` to `TypeAnalyzer` (if I saw that right).

wsmoses: Depending on the particular error thrown there may not be a gutils created.

But we can for sure audit all the current error types and see what we can do

wsmoses:
ok new jll is out

src/errors.jl
```diff
@@ -1711,15 +1713,19 @@ end
...
                 f,
                 false,
             ) #=error=#
-            world = enzyme_extract_world(f)
+            world = enzyme_context(gutils).world
         end
+        # TODO(vchuravy)
+        # what is data?
+        # Can we get world here?
+        @assert world == enzyme_context(gutils).world
```
vchuravy:
```suggestion
```

src/errors.jl
```diff
@@ -1175,22 +1175,20 @@ function julia_error(
         API.EnzymeStringFree(ip)
 
         mi = nothing
-        world = nothing
+        world = enzyme_typeanalyzer_context(data).world
```
vchuravy:
```suggestion
        world = enzyme_context(data).world
```

With a multiple dispatch on typeanalyzer

src/gradientutils.jl
```diff
@@ -366,3 +366,15 @@ function batch_call_same_with_inverted_arg_if_active!(
...
+    ptr = API.EnzymeGradientUtilsGetExternalContext(gutils)
+    @assert ptr != C_NULL
+    return unsafe_pointer_to_objref(ptr)::EnzymeContext
+end
+
+function enzyme_gutils_context(gutils::API.EnzymeGradientUtilsRef)
+    ptr = API.EnzymeGradientUtilsGetExternalContext(gutils)
+    @assert ptr != C_NULL
+    return unsafe_pointer_to_objref(ptr)::EnzymeContext
+end
```
vchuravy: Should be possible to remove

src/jlrt.jl
```diff
@@ -573,88 +573,6 @@ function emit_jltypeof!(B::LLVM.IRBuilder, @nospecialize(arg::LLVM.Value))::LLVM
     call!(B, FT, fn, [arg])
 end
 
-function emit_methodinstance!(B::LLVM.IRBuilder, @nospecialize(func), args::Vector{LLVM.Value})::LLVM.Value
```
wsmoses, line 576: don't remove this?

src/rules/activityrules.jl
```diff
@@ -1,5 +1,5 @@
 
-function julia_activity_rule(f::LLVM.Function, method_table)
+function julia_activity_rule(f::LLVM.Function, world, method_table)
```
wsmoses, line 2: add the explicit ::UInt on world

src/compiler.jl
```diff
@@ -1353,6 +1358,7 @@ function nested_codegen!(
     GPUCompiler.prepare_job!(job)
     otherMod, meta = GPUCompiler.emit_llvm(job)
     
+    # TODO: interp should be cached since it contains internal caches
```
wsmoses, line 1361: ?

wsmoses (approved):

src/compiler.jl
```diff
@@ -533,13 +538,14 @@ include("llvm/transforms.jl")
...
-    funcspec = my_methodinstance(mode == API.DEM_ForwardMode ? Forward : Reverse, typeof(f), tt, world)
-    nested_codegen!(enzyme_context, mode, mod, funcspec, world)
+function nested_codegen!(enzyme_context::EnzymeContext, mode::API.CDerivativeMode, mod::LLVM.Module, @nospecialize(f), @nospecialize(tt::Type))
+    funcspec = my_methodinstance(mode == API.DEM_ForwardMode ? Forward : Reverse, typeof(f), tt, enzyme_context.world)
+    nested_codegen!(enzyme_context, mode, mod, funcspec)
 end
 
 
 function prepare_llvm(interp, mod::LLVM.Module, job, meta)
+    # TODO: remove enzymejl_world
```
vchuravy: Remove enzymejl_world entirely

src/rules/activityrules.jl
vchuravy:
```suggestion
function julia_activity_rule(f::LLVM.Function, world::UInt, method_table)
```

src/compiler.jl
vchuravy: Done in 3f554b69. The writers, the wrapper copies and `enzyme_extract_world` are gone. `setup_global` decided a global's inactivity with `active_reg` in that world; it now uses the generated `active_reg_nothrow`, which needs no world and is regenerated when `inactive_type` changes. Errors raised from module-level passes (`nodecayed_phis!`, `sret_ty`, the internal-error handler) carry no world any more; `code_typed(err)` shows them in the current one.

src/errors.jl
vchuravy: Done, as `enzyme_context(TypeAnalyzer(data)).world`: `EnzymeTypeAnalyzerRef` and `EnzymeGradientUtilsRef` are both `Ptr{Cvoid}`, so a method on the ref type would silently accept a gutils pointer; wrapping in `TypeAnalyzer` is the dispatch.

src/gradientutils.jl
vchuravy: Removed.

src/jlrt.jl
vchuravy: Restored in 3f554b69, with an explicit `world::UInt` argument since the attribute it read from is gone.

src/compiler.jl
vchuravy: Stale note, dropped.

src/errors.jl
vchuravy: Gone with the squash.

src/Enzyme.jl
vchuravy: `enzyme_extract_world` is gone as of 3f554b69; the context is the only source of the world now.

src/utils.jl
```diff
@@ -104,7 +104,6 @@ export unsafe_to_ptr
 function setup_global(
         B::LLVM.IRBuilder,
         T_jlvalue::LLVM.StructType,
-        world::Union{UInt, Nothing},
```
vchuravy, line 107: This is wrong `setup_global` should get the world from the outside.

src/utils.jl
```diff
@@ -136,10 +135,12 @@ function setup_global(
...
         inactive = true
     end
-    if !inactive && world isa UInt
+    if !inactive
         legal, jTy, byref = Compiler.abs_typeof(gv, true)
         if legal
-            state = Enzyme.Compiler.active_reg(jTy, world)
+            # World-free: the generated `active_reg_nothrow` carries an edge to
+            # `EnzymeRules.inactive_type`, so it is regenerated when that changes.
+            state = Enzyme.Compiler.active_reg_nothrow(jTy)
```
vchuravy: Can't use this variant since we don't attach the edges. Instead we should be passing in world as a UInt from the outside.

src/utils.jl
vchuravy: Reverted to an explicit `world::UInt` in b369e3c0. `unsafe_to_llvm` and every helper that reaches it (`emit_allocobj!`, `emit_apply_type!`, `emit_tuple!`, `emit_jltypeof!`, `emit_error`, `makeInstanceOf`, `get_memory_*`, `box_inline_union!`, the conditional-execute callbacks, ...) now take the world from the caller: rules pass `enzyme_context(gutils).world`, codegen the job's world. The only place without a caller-supplied world is Enzyme's C callbacks (`julia_allocator`, `fixup_return`, `julia_sanitize`, and the error handler when it is handed a null GradientUtils): `enzyme!` publishes the job's world in task-local storage for the duration of the C++ call and `callback_world()` reads it. That is the one seam left; giving those callbacks a context argument in the C API removes it (CreateAllocation has ~20 callers in enzyme-core, some without a GradientUtils, so it would be the EnzymeLogic's ExternalContext that gets threaded).

src/utils.jl
vchuravy: Back to `active_reg(jTy, world)` with the world passed in (b369e3c0).

src/errors.jl
```diff
@@ -1036,6 +1043,7 @@ function julia_error(
     data::Ptr{Cvoid},
     data2::LLVM.API.LLVMValueRef,
     B::LLVM.API.LLVMBuilderRef,
+    context::Ptr{Cvoid} = C_NULL,
```
vchuravy, line 1046:
```suggestion
    context::Ptr{Cvoid},
```

src/errors.jl
```diff
@@ -1110,34 +1118,37 @@ function julia_error(
             end
 
             mi = nothing
-            world = nothing
+            # `data` is the `GradientUtils` when the error is raised while
+            # differentiating a function, and null from the batching, truncation
+            # and tracing entry points.
+            world = data == C_NULL ? enzyme_context(context).world :
+                enzyme_context(GradientUtils(API.EnzymeGradientUtilsRef(data))).world
```
vchuravy, line 1125: use `context` directly

src/errors.jl
```diff
@@ -1266,18 +1273,16 @@ function julia_error(
...
             mi, rt = enzyme_custom_extract_mi(
                 f,
                 false,
             ) #=error=#
-            world = enzyme_extract_world(f)
         end
 
         err = if mi !== nothing
-            EnzymeInternalError{Core.MethodInstance, UInt}(msg, irstr(), bt, mi, world)
+            EnzymeInternalError{Core.MethodInstance, Nothing}(msg, irstr(), bt, mi, nothing)
```
vchuravy, line 1285: This should take world again

vchuravy:
Opened #3553 as an alternative to this PR, stacked on #3552 (`EnzymeContext` as a `ScopedValue`).

Same destination — world in `EnzymeContext`, `enzymejl_world` gone — reached without the C++ handles. Because the context is bound to a scoped value in `compile_unhooked`, a callback reads `enzyme_world()` directly: no `EnzymeLogicSetExternalContext` round trip, no `EnzymeGradientUtilsGetExternalContext`/`EnzymeTypeAnalyzerGetLogic`, no task-local storage for the callbacks that are handed neither, and the rule handlers lose arguments rather than gaining a `gutils` to look up from. It also finishes the removal that this branch still has TODOs for: nothing writes the attribute, and `enzyme_extract_world` is gone.

Happy to close #3553 if you would rather land this one.

PR #2645, opened 2025-10-05
Add new test mode rewind

kshyatt:
This allows you to:
1. Test reverse mode as usual (against FD)
2. Run the AD in forward mode to generate output tangents
3. Use those output tangents to test reverse mode again (against FD) -- this implicitly tests the forward rule

This can be helpful when the tangents need a particular choice of gauge (e.g. the tangents for eigenvectors) -- FD will generate tangents with an arbitrary gauge.

### DIFF
--- a/lib/EnzymeTestUtils/src/test_rewind.jl
+++ b/lib/EnzymeTestUtils/src/test_rewind.jl
@@ -1,0 +1,114 @@
+"""
+    test_rewind(f, Activity, args...; kwargs...)
+
+Test `Enzyme.autodiff` of `f` in `Forward`-mode by backtracking using `Reverse`-mode,
+which itself is checked against finite differences. This mode can be useful when computing
+derivatives on functions such as matrix factorizations, where a particular choice of
+gauge is important and the finite-differences approach generates tangents in an arbitrary
+gauge. In effect, this plays the derivatives _forward_, then in _reverse_, "rewinding" the
+tape.
+
+`f` has all constraints of the same argument passed to `Enzyme.autodiff`, with additional
+constraints:
+- If it mutates one of its arguments, it _must_ return that argument.
+
+To use this test mode, `f` _must_ have both forward and reverse rules defined.
+
+# Arguments
+
+- `Activity`: the activity of the return value of `f`
+- `args`: Each entry is either an argument to `f`, an activity type accepted by `autodiff`,
+    or a tuple of the form `(arg, Activity)`, where `Activity` is the activity type of
+    `arg`. If the activity type specified requires a tangent, a random tangent will be
+    automatically generated.
+
+# Keywords
+
+- `rng::AbstractRNG`: The random number generator to use for generating random tangents.
+- `fdm=FiniteDifferences.central_fdm(5, 1)`: The finite differences method to use.
+- `fkwargs`: Keyword arguments to pass to `f`.
+- `rtol`: Relative tolerance for `isapprox`.
+- `atol`: Absolute tolerance for `isapprox`.
+- `testset_name`: Name to use for a testset in which all tests are evaluated.
+- `output_tangent`: Optional final tangent to provide at the beginning of the reverse-mode differentiation 
+
+# Examples
+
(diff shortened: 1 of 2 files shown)


lib/EnzymeTestUtils/src/test_rewind.jl
```diff
@@ -0,0 +1,114 @@
+"""
+    test_rewind(f, Activity, args...; kwargs...)
+
+Test `Enzyme.autodiff` of `f` in `Forward`-mode by backtracking using `Reverse`-mode,
+which itself is checked against finite differences. This mode can be useful when computing
+derivatives on functions such as matrix factorizations, where a particular choice of
+gauge is important and the finite-differences approach generates tangents in an arbitrary
+gauge. In effect, this plays the derivatives _forward_, then in _reverse_, "rewinding" the
```
vchuravy, line 8: I must admit this is the first time I encounter the term `gauge`.

vchuravy (commented):
Needs some tests/usage

Does it make sense to use `fdm` here? Or should we simply be testing for consistency between forward and reverse?

lib/EnzymeTestUtils/src/test_rewind.jl
kshyatt: Ah sorry it's physicist brain

kshyatt:
I used `fdm` here because the idea is to test at first that the reverse mode works at all, then test with the output tangents from forward mode. To put it another way, keeping the `fdm` allows for some "external" consistency check for my rules -- imagine I wrote a bad rule that just always returns 0 for all `dval`s. That would pass the consistency check of fwd/rvs -- but it would be wrong. So I think keeping the `fdm` sanity check is good.

lib/EnzymeTestUtils/src/test_rewind.jl
kshyatt: https://en.wikipedia.org/wiki/Gauge_fixing if you're interested. I'll try to rephrase

sethaxen:
In general I think forward-mode through an eigendecomposition will produce tangents "in the gauge" (I forget formally what the property is, something about the specific tangent space). But there's actually no guarantees in reverse-mode what kind of (co)-tangent you will have on the outputs. If downstream code interacts with the eigenvectors in a gauge-dependent manner, then your cotangents may not be "in the gauge" at all. So if you're implementing a rule, ideally you'd have one that works for all co-tangents; if that's not possible, then you'd need to require and document that downstream code interacts with the eigenvectors only in a gauge-invariant way. In that case, then I think this new testing method would work, but this seems highly specialized to this particular problem. If we think of the tangents as being constrained to a specific manifold, I can't actually remember off the top of my head if the manifold containing the co-tangent should _always_ be a superset of this specific manifold.

It would be nice to have a second motivating application for this before adding another function to the API.

In terms of design, a few notes:
- For mutating functions `test_forward` and `test_reverse` will mutate the original arguments (including the function itself if it changes its own state), so you'll need some way to avoid this.
- Why is the first `test_reverse` unnecessary? Seems like it could just be left to the user to also run `test_reverse` themselves.
- I wonder then if this could just be an option for `test_reverse`, e.g. an option like `output_tangent` set to a singleton `AutoFwd` would indicate that forward-mode should be used to generate the output-tangent
- As an aside, I just saw that #2588 added an `output_tangent` keyword. I think a more consistent API than adding the keyword `output_tangent` would be to allow `ret_activity` to be a tuple `(output_tangent, activity)`, like the other arguments; then my above proposal would be to allow a user to provide e.g. `(AutoFwd, Duplicated)` to trigger this behavior.

kshyatt:
Thanks for the comment! I agree this is pretty specialized -- it's just nearly all the functions I need to do AD with have this annoying gauge property. The problem isn't forward mode itself, it's the FD check (as FD generates tangents in a totally arbitrary gauge).

kshyatt:
I think for now this can be closed, as after some thought our forward sensitivities are really bad to test against FD and I wasn't testing what I thought I was here...

PR #2648, opened 2025-10-06
use ParallelTestRunner

vchuravy:


test/runtests.jl
```diff
@@ -12,4 +12,4 @@ function testfilter(test)
     return true
 end
 
-runtests(ARGS; testfilter)
+runtests(ARGS; testfilter, init_code=:(include("setup.jl")))
```
vchuravy: Eventually I would just write it above as

```
init_code = quote
    # content of setup.jl
end
```

giordano:
The `Info: Testing against` message is now being printed by _all_ test files.

vchuravy:
We can remove that, I think. It was just me making sure that a custom JLL would work on the test runners. I am always paranoid about that

PR #2650, opened 2025-10-07
Add GPL guards for CHOLMOD usage to make Enzyme.jl GPL-safe

ChrisRackauckas-Claude:
## Summary

This PR adds `@static if Base.USE_GPL_LIBS` guards around all CHOLMOD-related dispatches in Enzyme.jl to ensure license compatibility when Julia is built without GPL libraries (Suitesparse/CHOLMOD).

## Changes

- **src/analyses/activity.jl**: Added GPL guards around:
  - `ptreltype(::Type{SparseArrays.CHOLMOD.Dense{T}})` dispatch (lines 22-24)
  - `is_arrayorvararg_ty(::Type{SparseArrays.CHOLMOD.Dense{T}})` dispatch (lines 38-40)

- **src/absint.jl**: Added GPL guard around CHOLMOD struct type checks (lines 611-618):
  - Wrapped checks for `cholmod_dense_struct`, `cholmod_sparse_struct`, and `cholmod_factor_struct`
  - When GPL libs are not available, the code gracefully skips CHOLMOD-specific handling

## Test Plan

- [x] Package precompiles successfully with the changes
- [x] Package loads correctly: `using Enzyme` works without errors
- [x] Changes are backward compatible - when `Base.USE_GPL_LIBS` is true, behavior is unchanged

## Background

CHOLMOD is part of Suitesparse, which is licensed under GPL. When Julia is built with `USE_GPL_LIBS=0`, these libraries are not available, and code that references them should be guarded to maintain license compatibility.

### DIFF
--- a/src/absint.jl
+++ b/src/absint.jl
@@ -608,7 +608,12 @@ function abs_typeof(
             # add the extra poitner offset when loading here]. However for pointers constructed by ccall outside julia
             # to a julia object, which are not inline by type but appear so, like SparseArrays, this is a problem
             # and merits further investigation. x/ref https://github.com/EnzymeAD/Enzyme.jl/issues/2085
-            if !Base.allocatedinline(typ) && typ != SparseArrays.cholmod_dense_struct && typ != SparseArrays.cholmod_sparse_struct && typ != SparseArrays.cholmod_factor_struct
+            @static if Base.USE_GPL_LIBS
+                cholmod_exception = typ != SparseArrays.cholmod_dense_struct && typ != SparseArrays.cholmod_sparse_struct && typ != SparseArrays.cholmod_factor_struct
+            else
+                cholmod_exception = true
+            end
+            if !Base.allocatedinline(typ) && cholmod_exception
                 shouldLoad = false
                 offset %= sizeof(Int)
             else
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Hand checked, yes this is what I intended.

Otherwise `using Enzyme` fails on a Julia build without GPL. This makes it safe.

Any way we can get a patch out there quickly on this?

src/absint.jl
```diff
@@ -608,7 +608,12 @@ function abs_typeof(
             # add the extra poitner offset when loading here]. However for pointers constructed by ccall outside julia
             # to a julia object, which are not inline by type but appear so, like SparseArrays, this is a problem
             # and merits further investigation. x/ref https://github.com/EnzymeAD/Enzyme.jl/issues/2085
-            if !Base.allocatedinline(typ) && typ != SparseArrays.cholmod_dense_struct && typ != SparseArrays.cholmod_sparse_struct && typ != SparseArrays.cholmod_factor_struct
+            @static if Base.USE_GPL_LIBS
+                cholmod_exception = typ != SparseArrays.cholmod_dense_struct && typ != SparseArrays.cholmod_sparse_struct && typ != SparseArrays.cholmod_factor_struct
```
wsmoses, line 612: I would prefer to use isdefined

vchuravy: `USE_GPL_LIBS` is always defined.

wsmoses: ok sure, @vchuravy feel free to shepard this one in then

ChrisRackauckas:
I don't know how to test this in CI 😅 I don't think there are readily available GPL-free binaries.

oscardssmith:
We are hoping to add nogpl builds as installable from Juliaup but that has not yet happened yet, so IMO this should merge first and we can add tests later once CI supports this.

PR #2651, opened 2025-10-07
support nogpl builds

oscardssmith:
when building julia without GPL libs, CHOLMOD doesn't exist. For example `julia +pr59699`


ChrisRackauckas:
@oscardssmith I already created https://github.com/EnzymeAD/Enzyme.jl/pull/2650

oscardssmith:
oops, missed that.

PR #2657, opened 2025-10-08
Disable precompile on 1.12

vchuravy:


vchuravy:
I want the package to be loadable (even if it is not yet usable)

PR #2659, opened 2025-10-08
Add chunk size selection strategies

gdalle:
- [x] Add chunk size selection strategies to EnzymeCore
  - [x] `LargestChunk` (which replaces `chunk=nothing`) as discussed in https://github.com/SciML/ADTypes.jl/pull/124
  - [x] `AutoChunk` as discussed in #1542, #1545
  - [x] `FixedChunk{C}` (which replaces `chunk=Val(C)`)
- [x] Implement, test and document those strategies in Enzyme
  - [x] Shadow creation
  - [x] Forward gradient and Jacobian
  - [x] Reverse Jacobian


lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -797,4 +798,43 @@ end
...
+Select chunk size so that the corresponding array is processed in a single chunk.
+"""
+struct OneChunk <: ChunkStrategy end
+
+"""
+    AutoChunk()
+
+Select chunk size automatically based on internal Enzyme-specific heuristics.
+"""
+struct AutoChunk <: ChunkStrategy end
```
wsmoses, line 877: We should have a fixed chunk size selection here too

gdalle: That's already taken care of by passing `Val(C)`?

gdalle:
@wsmoses I think this is a good first step, and I'm not able to do the reverse-mode Jacobian fix on my own. Can someone help? Also, should we make `create_shadows` public and documented?

src/sugar.jl
```diff
@@ -502,10 +506,11 @@ end
 @inline specialize_output(output, input) = output
 
 """
-    gradient(::ForwardMode, f, x; shadows=onehot(x), chunk=nothing)
+    gradient(::ForwardMode, f, x, args...; chunk=nothing, shadows=create_shadows(chunk, x, args...))
 
-Compute the gradient of an array-input function `f` using forward mode. The
-optional keyword argument `shadow` is a vector of one-hot vectors of type `x`
+Compute the gradient of an array-input function `f` using forward mode.
+The optional keyword argument `chunk` optionally denotes the chunk size to use: it can be either `nothing`, `Val(C)` for some `C`, `SingleChunk()` or `AutoChunk()`.
```
gdalle:
```suggestion
The optional keyword argument `chunk` denotes the chunk size to use: it can be either `nothing`, `Val(C)` for some integer `C`, `SingleChunk()` or `AutoChunk()`.
```

src/sugar.jl
```diff
@@ -1173,7 +1180,7 @@ end
     jacobian(::ReverseMode, f, x)
 
 Compute the jacobian of a array-output function `f` using (potentially vector)
-reverse mode. The `chunk` argument optionally denotes the chunk size to use and
+reverse mode. The `chunk` argument optionally denotes the chunk size to use (it can be either `nothing` or `Val(C)` for some `C`) and
```
gdalle:
```suggestion
reverse mode. The `chunk` argument optionally denotes the chunk size to use (it can be either `nothing` or `Val(C)` for some integer `C`) and
```

lib/EnzymeCore/src/EnzymeCore.jl
wsmoses: If we're going to have a chunkstrategy, we might as well be explicit rather than passing in a val (we can still convert the val to FixedChunk or w/e til next bump and mark deprecated

gdalle: I added it, and also added `SmallestChunk` to eventually replace the current `chunk = nothing`

gdalle:
@wsmoses this is good to go, I fixed the reverse mode too

lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -797,4 +798,92 @@ end
...
+struct FixedChunk{C} <: ChunkStrategy end
+
+"""
+    AutoChunk()
+
+Select chunk size automatically based on internal Enzyme-specific heuristics, which are subject to change.
+"""
+struct AutoChunk <: ChunkStrategy end
+
+const DEFAULT_CHUNK_SIZE = 16
```
wsmoses, line 879: for type reasons, should we make this Val(16), otherwise it wont necessarily be a constant within pick_chunksize?

src/sugar.jl
```diff
@@ -446,7 +450,7 @@ end
             push!(exprs, :(nothing))
         elseif ty <: AbstractFloat
             push!(exprs, :(nothing))
-        elseif ChunkTy == Nothing || ChunkTy == Val{1}
+        elseif ChunkTy == Nothing || ChunkTy == Val{1} || ChunkTy == SmallestChunk || ChunkTy == FixedChunk{1}
```
wsmoses: can we rewrite this to only have a chunkstrategy input?

src/sugar.jl
```diff
@@ -502,10 +506,11 @@ end
 @inline specialize_output(output, input) = output
 
 """
-    gradient(::ForwardMode, f, x; shadows=onehot(x), chunk=nothing)
+    gradient(::ForwardMode, f, x, args...; chunk=nothing, shadows=create_shadows(chunk, x, args...))
```
wsmoses: we should change chunk=nothing, to the relevant correct explicit default.

we should also not support val/nothing inside of here and isntead add a deprecated method (or perhaps first check in the expr) if its one of the legacy methods and mark as deprecated

src/sugar.jl
```diff
@@ -1172,9 +1179,9 @@ end
     jacobian(::ReverseMode, f, x; n_outs=nothing, chunk=nothing)
     jacobian(::ReverseMode, f, x)
 
-Compute the jacobian of a array-output function `f` using (potentially vector)
-reverse mode. The `chunk` argument optionally denotes the chunk size to use and
-`n_outs` optionally denotes the shape of the array returned by `f` (e.g `size(f(x))`).
+Compute the jacobian of a array-output function `f` using (potentially vector) reverse mode.
+The optional keyword argument `chunk` denotes the chunk size to use: it can be either `nothing`, `Val(C)` for some integer `C`, or a subtype of [`EnzymeCore.ChunkStrategy`](@ref EnzymeCore.ChunkStrategy).
```
wsmoses: val/nothing should be marked deprecated

lib/EnzymeCore/src/EnzymeCore.jl
gdalle: I don't think so because

1. The constant is correctly inferred when returned as a `Val` in a function

```julia
julia> const C = 3
3

julia> f() = Val(C)

julia> @code_warntype f()
MethodInstance for f()
  from f() @ Main REPL[2]:1
Arguments
  #self#::Core.Const(Main.f)
Body::Val{3}
1 ─ %1 = Main.Val::Core.Const(Val)
│   %2 = (%1)(Main.C)::Core.Const(Val{3}())
└──      return %2
```

2. It doesn't matter anyway because `pick_chunksize(::AutoChunk, a)` will be type-unstable since it depends on the array size:

```julia
julia> g(n) = Val(min(n, C))
g (generic function with 1 method)

julia> @code_warntype g(4)
MethodInstance for g(::Int64)
  from g(n) @ Main REPL[4]:1
Arguments
  #self#::Core.Const(Main.g)
  n::Int64
Body::Val
1 ─ %1 = Main.Val::Core.Const(Val)
│   %2 = Main.min::Core.Const(min)
│   %3 = (%2)(n, Main.C)::Int64
│   %4 = (%1)(%3)::Val
└──      return %4

File: src/sugar.jl (hunk 5 above)
Comment (gdalle) [reply src/sugar.jl]: done in the latest commit

File: src/sugar.jl (hunk 6 above)
Comment (gdalle) [reply src/sugar.jl]: Done in the latest commit.

An issue with the current code is that the deprecation warning will only be visible at the first function call, since that is the only time where the generating function is actually generated:

```julia
julia> jacobian(Forward, copy, ones(2); chunk=nothing)
┌ Warning: The `chunk=nothing` configuration will be deprecated in a future release. Please use `chunk=SmallestChunk()` instead.
│   caller = #s719#135 at sugar.jl:461 [inlined]
└ @ Core ~/Documents/GitHub/Julia/Enzyme.jl/src/sugar.jl:461
┌ Warning: The `chunk=nothing` configuration will be deprecated in a future release. Please use `chunk=SmallestChunk()` instead.
│   caller = #s717#137 at sugar.jl:621 [inlined]
└ @ Core ~/Documents/GitHub/Julia/Enzyme.jl/src/sugar.jl:621
([1.0 0.0; 0.0 1.0],)

julia> jacobian(Forward, copy, ones(2); chunk=nothing)
([1.0 0.0; 0.0 1.0],)
```

Not sure whether that's an issue or not

src/sugar.jl
gdalle: done in the latest commit

src/sugar.jl
gdalle: If we move it to a permanent warning at every call, we should probably add tests for it too

gdalle:
Do you want this to error? 

```julia
jacobian(Forward, copy, ones(2); chunk=FixedSize{3}())
```

Currently Enzyme doesn't mind when the chunk size is larger than the input, I'm not sure what the expected behavior there is (and it's hard to deduce from the code). For comparison:

```julia
julia> cfg = ForwardDiff.JacobianConfig(copy, ones(2), ForwardDiff.Chunk{3}());

julia> ForwardDiff.jacobian(copy, ones(2), cfg, Val(true))
ERROR: ArgumentError: chunk size cannot be greater than ForwardDiff.structural_length(x) (3 > 2)
```

wsmoses:
I'm okay for it not erring at the moment, but we can always add it later

lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -797,4 +798,92 @@ end
...
+abstract type ChunkStrategy end
+
+"""
+    SmallestChunk()
+
+Select chunk size equal to 1, so that the corresponding array is processed in as many chunks as it has elements.
+
+!!! tip
+    In the current Enzyme interface, this strategy is equivalent to setting `chunk = nothing`.
+"""
```
wsmoses, line 856: can we somehow set this SmallestChunk = FixedChunk(1) or something? otherwise this feels redundant.

also chunk = nothing rn I think actually does largestchunk atm (but I may be misremembering)

gdalle:
@wsmoses I think the last aspect to settle is https://github.com/EnzymeAD/Enzyme.jl/pull/2659#discussion_r2479329980, but we can also change it in postprod since it's only about when to display deprecation warnings

lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -797,4 +798,68 @@ end
...
+
+"""
+    FixedChunk{C}()
+
+Select chunk size equal to a fixed integer `C`.
+
+!!! tip
+    In the current Enzyme interface, this chunk strategy is equivalent to setting `chunk = Val(C)`.
+"""
+struct FixedChunk{C} <: ChunkStrategy end
```
wsmoses, line 868: for ease of use can we add a constructor FixedChunk(c::Int) = FixedChunk{c}()

wsmoses (approved):

lib/EnzymeCore/src/EnzymeCore.jl
gdalle: Is it really a good idea? The constructor `FixedChunk{C}()` explicitly reminds people that `C` is in the type domain, and so that it might be type-unstable to construct it with an arbitrary runtime value. If we define `FixedChunk(C) = FixedChunk{C}()`, we hide that a little bit, that's why I didn't go for it. Ultimately it's your call though

wsmoses: Val(x) already has this, I think it's fine here

gdalle: Added it in the latest commit

gdalle:
Good to go, if CI passes!

wsmoses (approved):
lgtm but I'm going to wait for @vchuravy to get back from vacation (ending this week iirc) to see if he has any api thoughts before merging

gdalle:
@vchuravy what do you think?

gdalle:
Gentle bump, is this worth pursuing?

gdalle:
Hey there, I was celebrating the birthday of https://github.com/SciML/ADTypes.jl/pull/124 and I was wondering whether this is still worth pursuing @wsmoses?

gdalle:
Gonna try a random @vchuravy tag for good measure

ChrisRackauckas:
@wsmoses the thing you were waiting on me solving for a bit https://github.com/SciML/Optimization.jl/pull/1274 https://github.com/SciML/Optimization.jl/pull/1319 actually was having issues due to this not being supported very well in DI, which is due to this PR being missing. I am going to hack around it to get things working for now, but this should really get looked at here so the full solution can be in DI.

PR #2660, opened 2025-10-08
[CI] Build documentation on Julia v1.11

giordano:
Until v1.12 is properly supported by Enzyme.jl, we need to stick to v1.11 for building the docs.


giordano:
I guess this was missed in #2656, but the motivation is the same.

PR #2669, opened 2025-10-14
fix unsafe_to_pointer definition for opaque pointers

vchuravy:
- **remove unecessary import**
- **fix unsafe_to_pointer definition for opaque pointers**


vchuravy:
I am splitting my PRs to land work early, while I keep  working on the rest.

PR #2671, opened 2025-10-14
Add single 1.12 CI job with assertions enabled

vchuravy:
In contrast to #2613 this only adds one job for now. I propose that we merge this so that we can have a canary when things are breaking (more).

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -72,6 +72,11 @@ jobs:
             libEnzyme: packaged
             version: '1.11'
             assertions: true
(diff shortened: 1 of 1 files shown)


giordano:
> we can have a canary when things are breaking (more).

Not easy to actively see when there are more breaking changes when tests are already failing, is it?

vchuravy:
> Not easy to actively see when there are more breaking changes when tests are already failing, is it?

It's more to show that the 1.12 PRs I am making are making headway.

giordano (approved):
Ok, I see 👍

PR #2673, opened 2025-10-17
Update DynamicPPL version in Project.toml

wsmoses:


giordano:
The extension isn't really tested, is it?

penelopeysm:
> The extension isn't really tested, is it?

No, also it's type piracy.

This compat bound is making DynamicPPL CI fail on a number of PRs. I tested with Enzyme 0.13.86 manually and the error has already been fixed in some recent patch. So the compat needs to be fixed or I'll have to start removing Enzyme from our CI benchmarks

wsmoses:
It should be testable once
https://github.com/EnzymeAD/Enzyme.jl/pull/1813 goes through (which
relatedly, gently ping)

@penelopeysm per our convo on slack where this PR was created to unblock
you I’m going to go ahead and merge and figure out how do to the “version
compat through up to 0.50 or something” in a follow up PR

PR #2674, opened 2025-10-18
First pass at easyrules

wsmoses:


test/rules/easyrules.jl
```diff
@@ -0,0 +1,82 @@
...
+    @test autodiff(ForwardWithPrimal, mysin, Duplicated(2.0f0, 1.2f0))[1] ≈ (mysin(2.0f0), 1.2f0 * exp(2.0f0))
+    @test calls[] == 3
+
+    @test autodiff(ForwardWithPrimal, mysin, BatchDuplicated(2.0f0, (1.2f0, 1.4f0)))[1] ≈ (mysin(2.0f0), (1.2f0 * exp(2.0f0), 1.4f0 * exp(2.0f0)))
+    @test calls[] == 4
+end
+
+function mymul(x, y)
+    return sin(x)
+end
```
giordano, line 97: Is `sin(x)` right?

wsmoses:
@vchuravy do you know why the benchmarking ci fails?

vchuravy:
So the core of the idea here is that the user symbolically provides the Jacobian? 

I wonder if we could take the idea a bit further.

```
EnzymeRules.@easy_scalar_rule(
    mytup(x::Float64, y::Float64),
    (myexp(x), Enzyme.Const),
    (Enzyme.Const, 0.123456),
)
```

```
using Symbolics

@variables x y
J = [exp(x)  0
     0    1.23456]
```

So could we go from a Jacobian specified with Symbolics (instead of quasi_quotation) to forward-rules?

lib/EnzymeCore/src/easyrules.jl
```diff
@@ -0,0 +1,727 @@
...
+                $(gensetup...)
+                return $genres
+            end
+        end
+    end
+end
+
+"""
+    @easy_rule(f(x₁::T1, x₂::T2, ...),
+                 @setup(statement₁, statement₂, ...),
```
vchuravy, line 672: Why does the `setup` need to be a new macro?

lib/EnzymeCore/src/easyrules.jl
```diff
@@ -0,0 +1,727 @@
...
+    end
+
+    actives = Expr[]
+    for ann_name in input_names
+        push!(actives, Expr(:if, Expr(:call, <:, esc(ann_name), EnzymeCore.Const), nothing, :(Expr(:(.), Symbol($(String(ann_name))), :(:dval)))))
+    end
+
+    N = length(inputs)
+
+    @strip_linenos quote
```
vchuravy, line 360: Can we add a test that an error from the rule has a backtrace that points to the rule?

src/rules/customrules.jl
```diff
@@ -1,3 +1,100 @@
+import LinearAlgebra
+
+@inline add_fwd(prev, post) = recursive_add(prev, post)
```
vchuravy, line 3: This is currently not used.

vchuravy:
Would be nice to support callable structs

src/rules/customrules.jl
wsmoses: yes it is

wsmoses:
```

    if partial <: Number || dx isa Number
        if prev !== Nothing
            return quote
                Base.@_inline_meta
                EnzymeCore.EnzymeRules.add_fwd(prev, EnzymeCore.EnzymeRules.multiply_fwd_into(nothing, partial, dx))
            end
        end
        return quote
            Base.@_inline_meta
            partial * dx
        end
    end

    @assert partial <: AbstractArray
    @assert dx <: AbstractArray
    N = ndims(partial)
    M = ndims(dx)

    if N == M
        if prev !== Nothing
            return quote
                Base.@_inline_meta
                EnzymeCore.EnzymeRules.add_fwd(prev, EnzymeCore.EnzymeRules.multiply_fwd_into(nothing, partial, dx))
            end
        end
```

test/rules/easyrules.jl
```diff
@@ -0,0 +1,218 @@
...
+    @test calls[] == 6
+end
+
+function mymul(x, y)
+    return x * y
+end
+
+EnzymeRules.@easy_rule(
+    mymul(x::Float64, y::Float64),
+    (myexp(x), Enzyme.Const)
```
vchuravy: Could we use `NoTangent()` from ChainRulesCore?

test/rules/easyrules.jl
```diff
@@ -0,0 +1,218 @@
...
+
+const calls = Ref(0)
+
+function myexp(x)
+    global calls
+    calls[] += 1
+    return exp(x)
+end
+
+EnzymeRules.@easy_rule(
```
vchuravy, line 21:
```suggestion
EnzymeRules.@rule(
```

lib/EnzymeCore/src/easyrules.jl
```diff
@@ -0,0 +1,727 @@
...
+             (∂f₂_∂x₁, ∂f₂_∂x₂, ...),
+             ...)
+```
+
+If a specific argument has no partial derivative, then all corresponding argument values can instead be marked `Enzyme.Const`. For example, consider the case where `config` has no derivative.
+
+```julia
+@easy_rule(f(config, x, ...),
+             @setup(nothing),
+             (Enzyme.Const, ∂f₁_∂x, ...),
```
wsmoses: how about `@Constant`

src/rules/customrules.jl
vchuravy: But `add_fwd` in `src/rules/customrules.jl` is not the same as `EnzymeCore.EnzymeRules.add_fwd`?

PR #2679, opened 2025-10-22
Don't run `threads` tests, they're driven by `multi-threads`

giordano:


vchuravy:
Should swap it to a custom test worker after https://github.com/JuliaTesting/ParallelTestRunner.jl/issues/62

giordano:
> We will still run them single-threaded right?

Yes, `multi-threads` runs with 1 and 2 threads: https://github.com/EnzymeAD/Enzyme.jl/blob/d2137d423a334d9eeee7572c17ccc80c8f5963c6/test/multi-threads.jl#L3-L8 At the moment we're basically running the 1 thread twice.

Issue #2664, opened 2025-10-10
Possible typo `LLVM` instead of `LVM` in `lower_convention`?


ymardoukhi:
https://github.com/EnzymeAD/Enzyme.jl/blob/f360f6bc57f06a986edb57b3527b48b2a0d27a3d/src/compiler.jl#L4404

I hit this error when dealing with `@views`.

```julia
ERROR: UndefVarError: `LVM` not defined in `Enzyme.Compiler`
Suggestion: check for spelling errors or missing imports.
Stacktrace:
  [1] (::Enzyme.Compiler.var"#359#366"{LLVM.Module, LLVM.Function})(io::IOBuffer)
    @ Enzyme.Compiler ~/.julia/dev/Enzyme/src/compiler.jl:4402
  [2] sprint(::Function; context::Nothing, sizehint::Int64)
    @ Base ./strings/io.jl:114
  [3] sprint(::Function)
    @ Base ./strings/io.jl:107
  [4] lower_convention(functy::Type, mod::LLVM.Module, entry_f::LLVM.Function, actualRetType::Type, RetActivity::Type, TT::Union{…}, run_enzyme::Bool)
    @ Enzyme.Compiler ~/.julia/dev/Enzyme/src/compiler.jl:4400
  [5] compile_unhooked(output::Symbol, job::GPUCompiler.CompilerJob{…})
    @ Enzyme.Compiler ~/.julia/dev/Enzyme/src/compiler.jl:4713
  [6] compile(target::Symbol, job::GPUCompiler.CompilerJob; kwargs::@Kwargs{})
    @ GPUCompiler ~/.julia/packages/GPUCompiler/Gp8bZ/src/driver.jl:67
  [7] compile
    @ ~/.julia/packages/GPUCompiler/Gp8bZ/src/driver.jl:55 [inlined]
  [8] _thunk(job::GPUCompiler.CompilerJob{…}, postopt::Bool)
    @ Enzyme.Compiler ~/.julia/dev/Enzyme/src/compiler.jl:5946
  [9] _thunk
    @ ~/.julia/dev/Enzyme/src/compiler.jl:5944 [inlined]
 [10] cached_compilation
    @ ~/.julia/dev/Enzyme/src/compiler.jl:6001 [inlined]
[...]

PR #2684, opened 2025-10-24
Test embedded bitcode path

ymardoukhi:
Test added for capturing `Broken function` exception and its corresponding `stdout` message regarding `Called function is not the same type as the call!` due to automatic inlining of `ccall` function.

See issue #2664

### DIFF
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -513,6 +513,9 @@ function prepare_llvm(interp, mod::LLVM.Module, job, meta)
             attributes,
             StringAttribute("enzymejl_rt", string(convert(UInt, unsafe_to_pointer(RT)))),
         )
+        if EnzymeRules.has_easy_rule_from_sig(Interpreter.simplify_kw(mi.specTypes); job.world)
+            push!(attributes, LLVM.StringAttribute("enzyme_LocalReadOnlyOrThrow"))
+        end
         if returnRoots
             attr = StringAttribute("enzymejl_returnRoots", "")
             push!(parameter_attributes(llvmfn, 2), attr)
@@ -1008,6 +1011,7 @@ end
             Duplicated,
             nothing,
             run_enzyme,
+            world
         )
         if cur
             state.primalf = llvmfn
@@ -3663,6 +3667,7 @@ function lower_convention(
     @nospecialize(RetActivity::Type),
     @nospecialize(TT::Union{Type, Nothing}),
     run_enzyme::Bool,
+    world::UInt
 )
     entry_ft = LLVM.function_type(entry_f)
 
@@ -4222,7 +4227,9 @@ function lower_convention(
         attributes,
         StringAttribute("enzymejl_rt", string(convert(UInt, unsafe_to_pointer(rt)))),
     )
-
+    if EnzymeRules.has_easy_rule_from_sig(Interpreter.simplify_kw(mi.specTypes); world)
+        push!(attributes, LLVM.StringAttribute("enzyme_LocalReadOnlyOrThrow"))
+    end
     for prev in collect(function_attributes(entry_f))
         if kind(prev) == kind(StringAttribute("enzyme_ta_norecur"))
             push!(attributes, prev)
@@ -4720,6 +4727,7 @@ function GPUCompiler.compile_unhooked(output::Symbol, job::CompilerJob{<:EnzymeT
             job.config.params.rt,
             TT,
             params.run_enzyme,
+            job.world
         )
     end
 
@@ -6475,3 +6483,4 @@ end
 include("compiler/reflection.jl")
 
 end
+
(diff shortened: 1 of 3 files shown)


src/compiler.jl
giordano, line 1: There are loads of unrelated formatting changes which make it hard to review the relevant changes.

ymardoukhi: Will reformat with runic

src/compiler.jl
```diff
@@ -513,9 +513,6 @@ function prepare_llvm(interp, mod::LLVM.Module, job, meta)
             attributes,
             StringAttribute("enzymejl_rt", string(convert(UInt, unsafe_to_pointer(RT)))),
         )
-        if EnzymeRules.has_easy_rule_from_sig(Interpreter.simplify_kw(mi.specTypes); job.world)
```
wsmoses, line 516: this should _not_ have been deleted

src/compiler.jl
wsmoses: generally this should only be a one line change to the file, can you udpate the patch to just do that?

ymardoukhi: I will. I am using a different global formatter that affected the file. Reverting the changes.

test/embedded_bitcode.jl
giordano, line 1: I seem to remember LLVM IR isn't very portable across major versions, I presume that'd be a potential problem here when testing different Julia (and llvm) versions?

src/compiler.jl
giordano: @ymardoukhi are you working on an old version of the repo? You reverted the changes in #2674.

src/compiler.jl
```diff
@@ -6483,3 +6483,4 @@ end
 include("compiler/reflection.jl")
 
 end
+
```
giordano:
```suggestion
```

src/compiler.jl
ymardoukhi: I honestly have no clue why this got removed! But added it back.

test/embedded_bitcode.jl
```diff
@@ -0,0 +1,97 @@
...
+    expected_grad_a = (nothing, [1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0])
+    grad_a = gradient(Reverse, func_ccall, Const(0.0), a)
+    @test expected_grad_a == grad_a
+
+
+    errstream = joinpath(tempdir(), "stdout.txt")
+    err_llvmir = nothing
+    b = @view a[1:5]
+
+    @show errstream
```
giordano: Debugging print?

ymardoukhi: Have it removed.

test/embedded_bitcode.jl
ymardoukhi: I agree with you, and I also didn't feel this was the best test case.

The only way I could reproduce this error was with `ccall` inside a wrapper function. On the contrary, if I use`llvmcall`, I could get the gradient correctly. When I inspected the LLVM IR when passing a `view` of an array, Julia inlines the wrapper function with weakly typed subarray pointers, hence the mismatch at the callsite and the inner `ccall`.

Also, have a look at this [comment](https://github.com/EnzymeAD/Enzyme.jl/issues/2664#issuecomment-3426175672) in the original issue.

test/Project.toml
```diff
@@ -29,5 +31,6 @@ Enzyme = {path = ".."}
 EnzymeTestUtils = {path = "../lib/EnzymeTestUtils"}
 
 [compat]
+Clang_jll = "16.0.6"
```
wsmoses: this can't be put here, it breaks julia version compat [hence ci failures]

test/embedded_bitcode.jl
```diff
@@ -0,0 +1,93 @@
...
+    attributes #0 = { mustprogress nofree nosync nounwind willreturn }
+    attributes #1 = { mustprogress nocallback nofree nosync nounwind readnone speculatable willreturn }
+    attributes #2 = { noinline }
+"""
+
+
+tmp_dir = tempdir()
+tmp_so_file = joinpath(tmp_dir, "func.so")
+run(
+    pipeline(
```
wsmoses, line 44: is this even necessary here?

if you're starting from llvm anyways, why not just use llvmcall?

ymardoukhi: Surprisingly, I won't hit that execution path when using `llvmcall`. Have a look at this comment here
https://github.com/EnzymeAD/Enzyme.jl/issues/2664#issuecomment-3426175672

The wrapper function around `llvmcall` handles the `SubArray` input with no issues, whereas the same wrapper function around `ccall` will fail to compute the gradient. I stopped short of investigating the core issue; the only clue I got from the emitted LLVM IR is that Julia inlines the wrapper function, which leads to a mismatch between the types at the callsite.

vchuravy:
Could we use a small C library instead of using LLVM IR bitcode?

ymardoukhi:
> Could we use a small C library instead of using LLVM IR bitcode?

Great point! Will consider this.

PR #2690, opened 2025-10-25
Update to ParallelTestRunner v2

giordano:


.github/workflows/CI.yml
giordano, line 1: @icweaver I feel like you may be interested in the changes in this file: you can easily run tests in verbose mode in CI with
```yaml
      - uses: julia-actions/julia-runtest@v1
        with:
          test_args: '--verbose'
```
The other changes in this PR are for using multiple jobs on macOS (we need to see if we have enough memory for running them though)

test/runtests.jl
```diff
@@ -1,19 +1,34 @@
...
+    if Sys.iswindows()
+        delete!(testsuite, "ext/specialfunctions")
     end
-    if test == "threads"
-        # We run the "threads" tests via the "multi-threads" driver.
-        return false
+
+    # Add threads tests to be run with multiple Julia threads (will be
+    # configured in `test_worker`).
+    testsuite["threads/2"] = :(include($(joinpath(@__DIR__, "threads.jl"))))
```
giordano: As far as I understand, it's not possible to run just the custom `threads/2` testset from the command line, because when you pass a custom list of testsets as test arguments then `filter_tests!(testsuite, args)` would be false and this line wouldn't be stepped over.  However the user can do
```julia
using Pkg
Pkg.test("Enzyme"; julia_args=`--threads=2`, test_args=`threads`)
```

vchuravy: You should add this before the filter_tests

giordano: Ah, good point!  Done

.github/workflows/CI.yml
icweaver: Oh nice, appreciate the tip!

test/runtests.jl
```diff
@@ -1,19 +1,33 @@
...
+# Add threads tests to be run with multiple Julia threads (will be configured in
+# `test_worker`).
+testsuite["threads/2"] = :(include($(joinpath(@__DIR__, "threads.jl"))))
+
+# Parse arguments
+args = parse_args(ARGS)
+
+if filter_tests!(testsuite, args)
+    # Skip GPU-specific tests by default.
+    delete!(testsuite, "metal")
```
wsmoses, line 16: we're running these at least in gpu ci right?

giordano: This was skipped already before.

wsmoses: not the cuda one tho?

giordano: As far as I understand, the CUDA one is run at https://github.com/EnzymeAD/Enzyme.jl/blob/2cd186bee0195dc5eb3eebc6a599ce56676e4e0f/.buildkite/pipeline.yml#L25

test/runtests.jl
```diff
@@ -1,19 +1,33 @@
 import Enzyme
 import Enzyme_jll
-using ParallelTestRunner: runtests
+using ParallelTestRunner: addworker, filter_tests!, find_tests, parse_args, runtests
 
-function test_filter(test)
-    if test ∈ ("metal", "cuda", "amdgpu")
```
giordano, line 6: @wsmoses the cuda one was skipped here

PR #2692, opened 2025-10-26
EnzymeTestUtils: `GPUArray` support

wsmoses:


kshyatt:
Shouldn't this also support AMDGPU arrays since it's all `AbstractGPUArray` behind the scenes? Title is too modest, I think

kshyatt:
FWIW I'm still getting scalar indexing errors in `jvp` in FD, I think this might need an update?

PR #2698, opened 2025-10-27
Warn about julia 1.12

wsmoses:
given the number of issues/confusion, we should tell people it's expected for now with a link


wsmoses:
I'm not sure we have a good single entry point, unfortunately. Best would be the autodiff generated function (assuming we dont have issues with other utilities, which we do).

Is there a better reference issue (and/or should we merge issues into a different one)?

PR #2701, opened 2025-10-27
1.12: Fix absint inlining

wsmoses:


vchuravy (approved):
Looks fine, but do we need a branch? E.g. is this something that changed twice in 1.12?

wsmoses:
talking with oscar, no we dont -- we just had an intermediate fix

PR #2713, opened 2025-10-28
Rewrite optimize pipeline for NewPM

vchuravy:


vchuravy:
@wsmoses this now shows the same failure as before, but is just the rewrite of `optimize!`

src/compiler/optimize.jl
```diff
@@ -22,277 +22,123 @@ end
...
+        add!(pb, NewPMAAManager()) do aam
+            add!(aam, ScopedNoAliasAA())
+            add!(aam, TypeBasedAA())
+            add!(aam, BasicAA())
         end
-        add!(pm, ModulePass("GCInvariantVerifier", gc_invariant_verifier))
-    end
-end
+        add!(pb, NewPMModulePassManager()) do mpm
+            add!(mpm, Addr13NoAliasPass())
```
wsmoses, line 38: it looks like you didn't separate this out into different module runs, just for fun could we try that?

wsmoses: at minimum we could therefore binary seach the difference

src/compiler/optimize.jl
```diff
@@ -22,273 +22,151 @@ end
...
-        scoped_no_alias_aa!(pm)
-        type_based_alias_analysis!(pm)
-        basic_alias_analysis!(pm)
-        cpu_features_tm!(pm, tm)
-
-        LLVM.API.LLVMAddGlobalOptimizerPass(pm) # Exxtra
-        gvn!(pm) # Exxtra
-        LLVM.run!(pm, mod)
+    @dispose pb = NewPMPassBuilder() begin
+        add!(pb, NewPMAAManager()) do aam
```
wsmoses, line 168: lets see if we can just do this one, for example, then the prev module pipeline, etc, til we figure out the failure

vchuravy:
It turns out that https://github.com/EnzymeAD/Enzyme.jl/blob/107b32760d41618c461b41121d04f59c7abfea9f/src/compiler/optimize.jl#L441-L458 was running the middle pipeline twice. I figured out that you only run a few extra passes to get the test over the hill

wsmoses:
oh interesting.

that said, can we keep exact equivalent for the current new pm change, and then we can do pass pipeline tuning as a separate follow up?

wsmoses:
that way itll be easier to debug (e.g. git blame/figure out) if anything else goes awry/differently that wasnt caught by the CI tests

PR #2719, opened 2025-10-30
Error: improve no type error message

wsmoses:


wsmoses:
Closes https://github.com/EnzymeAD/Enzyme.jl/issues/2715

PR #2725, opened 2025-10-31
Add easy_rule for matrix det

kshyatt:
Something very odd going on with some of this.

- `test_forward` fails for the `Matrix{ComplexF64}` even though it's (secretly) all real valued
- `test_reverse` seems to be running the rule on each element of the matrix argument individually @sethaxen any ideas why?

### DIFF
--- a/src/internal_rules.jl
+++ b/src/internal_rules.jl
@@ -923,11 +923,11 @@ function EnzymeRules.reverse(
 end
 
 function cofactor(A)
-    cofA     = zeros(eltype(A), size(A))
+    cofA     = similar(A)
     minorAij = similar(A, size(A, 1) - 1, size(A, 2) - 1)
     for i in 1:size(A, 1), j in 1:size(A, 2)
         fill!(minorAij, zero(eltype(A)))
-        
+
         # build minor matrix
         for k in 1:size(A, 1), l in 1:size(A, 2)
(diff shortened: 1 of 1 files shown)


src/internal_rules.jl
```diff
@@ -922,11 +922,26 @@ function EnzymeRules.reverse(
...
+        # build minor matrix
+        for k in 1:size(A, 1), l in 1:size(A, 2)
+            if !(k == i || l == j)
+                ki = k < i ? k : k - 1
+                li = l < j ? l : l - 1
+                @inbounds minorAij[ki, li] = A[k, l]
+            end
+        end
+        @inbounds cofA[i, j] = (-1)^(i - 1 + j - 1) * det(minorAij)
+        minorAij .= zero(eltype(A))
```
wsmoses: use fill!(minorAij, zero) instead?

wsmoses:
> * `test_forward` fails for the `Matrix{ComplexF64}` even though it's (secretly) all real valued

So in this case it was actually testing complex [e.g. the random seeds had imag components], now fixed the complex failures from internal side

> * `test_reverse` seems to be running the rule on each element of the matrix argument individually @sethaxen any ideas why?

so you were acidentally doing for x in [...] because of a trailing comma

Issue #2528, opened 2025-08-26
Reverse Integration CI


wsmoses:
Now that we have some more CI resources, lets get reverse CI setup for some integration tests. We don't have infinite resources, but let's try to get some useful things going.

Specifically my thoughts here are to have CPU reverse integration CI, on 32-core machines. Each integration can have one of these runners [which do need to complete in some reasonable time, tbd]. These won't be considered blocking for PRs (unless they can spit out MWE's for examples of the failure), but can be used to early catch failures (and hopefully prevent any issues before the pop up, and provide some nicer notions of stability).

- [x] Lux @avik-pal could you take a look? (done in #2569) 
- [ ] Flux @CarloLucibello, @mcabbott could you take a look?
- [ ] Turing @mhauru could you take a look?
- [x] SciML @ChrisRackauckas could you take a look? (done in #2593)
- [ ] Comrade @ptiede could you take a look? [assuming theres some interesting CPU-only tests]
- [x] DifferentiationInterface @gdalle could you take a look? (done in #2531) 
- [ ] Molly @jgreener64 could you take a look? [assuming theres some interesting CPU-only tests] (#2728)
- [ ] Oceananigans @glwagner could you take a look?
- [ ] @MilesCranmer I remember you mentioning something else
- [ ] Chmy/friends @luraess could you take a look? [assuming theres some interesting CPU-only tests]
- [x] KernelAbstractions (#2709)
- [ ] MPI (#518)

gdalle:
Very good news! How do you want to do this, one `yml` file for all packages? GitHub actions or Buildkite? What is your idea of "reasonable time"?

wsmoses:
github actions. Up to either, but currently we have one yml for all integration tests. not sure about reasonable time yet, but it should for sure complete before the rest of Enzyme proper CI, and not hog all the CI VMs

wsmoses:
also cc @giordano

giordano:
One question is: do you want to just execute the full downstream test suite for each package (which would be comprehensive but may be very compute intensive and run loads of unrelated tests), or just have smaller test scripts to be included here which represent relevant workloads?

wsmoses:
I think we want to have smaller test scripts that are isolated to just enzyme-related tests [and to the extent possible, direct calls to Enzyme fns].

This is coming from two different constraints:
* obviously the downstream package devs want it to be as high-level/comprehensive as possible, to make it easier for them
[...]

PR #2728, opened 2025-11-01
Molly integration tests

jgreener64:
This took a while as Molly was in a state of flux, but with the release of v0.23.0 integration tests can be added as discussed in https://github.com/EnzymeAD/Enzyme.jl/issues/2528.

These tests pass on Julia 1.11 but error on 1.10 with a forward mode error that I can minimise next week. 1.12 also fails but I think that is being dealt with elsewhere.

Some of the tests are not at all minimal, but I am happy to be tagged to create minimal examples if they break. Having complex examples is beneficial as I do find that they break regularly.

### DIFF
--- a/.github/workflows/Integration.yml
+++ b/.github/workflows/Integration.yml
@@ -53,6 +53,10 @@ jobs:
           - SciML
           - KernelAbstractions
           - Molly
+        exclude:
+          - version: '1.10'
+            os: linux-x86-n2-32
+            package: Molly
     steps:
       - uses: actions/checkout@v5
       - uses: julia-actions/setup-julia@v2


vchuravy (approved):

.github/workflows/Integration.yml
```diff
@@ -52,6 +52,11 @@ jobs:
           - Lux
           - SciML
           - KernelAbstractions
+          - Molly
+        exclude:
+          - version: '1.10'
+            os: linux-x86-n2-32
```
giordano: It's unnecessary to specify also the `os` (and that's slightly fragile in case we change runners)
```suggestion
```

giordano: Uhm, my github-actions-fu failed me, apparently it was necessary after all 🤔 I removed the commit with my suggestion

giordano: Wait, no, my suggestion was right!  I misread the exclusion, I thought we wanted to exclude 1.11, and I saw the 1.11 job running, but we're actually excluding v1.10!

PR #2729, opened 2025-11-03
Update rules.jl with better docs for forward

kshyatt:
Fixed some typos and added a little more info about `RT`

### DIFF
--- a/lib/EnzymeCore/src/rules.jl
+++ b/lib/EnzymeCore/src/rules.jl
@@ -14,11 +14,17 @@ import Base: unwrapva, isvarargtype, unwrap_unionall, rewrap_unionall
 """
     forward(fwdconfig, func::Annotation{typeof(f)}, RT::Type{<:Annotation}, args::Annotation...)
 
-Calculate the forward derivative. The first argument is a [`FwdConfig](@ref) object
+Calculate the forward derivative. The first argument is a [`FwdConfig`](@ref) object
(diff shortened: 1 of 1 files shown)


wsmoses (approved):

lib/EnzymeCore/src/rules.jl
```diff
@@ -14,11 +14,17 @@ import Base: unwrapva, isvarargtype, unwrap_unionall, rewrap_unionall
...
-Either wrapped in a [`Const`](@ref)), or a [`Duplicated`](@ref) if it is a closure.
-The third argument is the return type annotation, and all other arguments are the annotated function arguments.
+The second argument `func` is the callable to which the rule applies,
+either wrapped in a [`Const`](@ref)), or a [`Duplicated`](@ref) if it is a closure.
+The third argument is the return type annotation, and all other arguments are the annotated arguments
+to the function `f`.
+
+Valid types for `RT` are:
+  - [`Duplicated`](@ref)
+  - [`DuplicatedNoNeed`](@ref)
```
giordano, line 26: I think this isn't documented and actually broke the docs

wsmoses: really, it should be?

wsmoses: it might be a namespace thing, lemme try to fix

PR #2730, opened 2025-11-03
Update return type annotations in rules.jl

wsmoses:
cc @kshyatt


giordano (approved):
Docs building is now successful, so it was indeed a scoping issue

PR #2731, opened 2025-11-03
fix: MethodError takes in a Tuple

avik-pal:


wsmoses (approved):

wsmoses:
also concurrently, @avik-pal would it be worthwhile adding (cpu) lux integration tests to enzyme too?

avik-pal:
We added those https://github.com/EnzymeAD/Enzyme.jl/actions/runs/19047933170/job/54400735705?pr=2731 already

wsmoses:
ah right, they're just consistently greenthat I forgot xD

PR #2732, opened 2025-11-04
Updates for rules docs

kshyatt:
More back ticks, links, and typo fixes


lib/EnzymeCore/src/rules.jl
```diff
@@ -175,19 +177,22 @@ end
...
 """
 function augmented_primal end
 
 """
     reverse(::RevConfig, func::Annotation{typeof(f)}, dret::Active, tape, args::Annotation...)
     reverse(::RevConfig, func::Annotation{typeof(f)}, ::Type{<:Annotation), tape, args::Annotation...)
 
-Takes gradient of derivative, activity annotation, and tape. If there is an active return dret is passed
-as Active{T} with the derivative of the active return val. Otherwise dret is passed as Type{Duplicated{T}}, etc.
+Takes gradient of derivative, activity annotation, and tape. If there is an active return, `dret`` is passed
```
wsmoses: too many backticks?

kshyatt: Oh true good catch

PR #2734, opened 2025-11-04
Bump DynamicPPL compat

penelopeysm:
@wsmoses I think this should work / keep us out of trouble for a while more. We have 0.39 in progress already so the bound needed an update anyway.


Project.toml
```diff
@@ -43,7 +43,7 @@ ADTypes = "1"
 BFloat16s = "0.2, 0.3, 0.4, 0.5, 0.6"
 CEnum = "0.4, 0.5"
 ChainRulesCore = "1"
-DynamicPPL = "0.35, 0.36, 0.37, 0.38"
+DynamicPPL = "0.35 - 0.99"
```
vchuravy: Oof, I am not a fan of that. This will cause me headaches down the road and requires careful manual handling of the compaz in the registry

vchuravy: This makes it feel like the extension belongs more into Dynamic PPL rather than Enzyme

penelopeysm: I am not completely a fan of this either, and I've pointed out already a couple of times that this extension is type piracy. Personally I would rather the extension be removed

penelopeysm: After discussion limited this just up to 0.39!

vchuravy (approved):

Issue #2706, opened 2025-10-28
Duplicated constants with the same value in the LLVM IR could be merged?


ymardoukhi:
When new thunks are generated, Enzyme fails to merge the constants with the same value in the extended module.

I expect to merge those constants with the same value. If constants with different values are encountered an error must be raised.

```julia
using Enzyme, Clang_jll, Libdl

const FUNC_LLVM_IR = raw"""
	; ModuleID = '<stdin>'
	source_filename = "<string>"
	target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
	target triple = "x86_64-conda-linux-gnu"
	
	@.const.array.data.5 = hidden unnamed_addr constant [48 x i8] c"\00\00\00\00\00\00\00\00\00\00\00\00\00\00\D0?\00\00\00\00\00\00\D0?\00\00\00\00\00\00\E8?\00\00\00\00\00\00\E8?\00\00\00\00\00\00\F0?", align 8
	
	; Function Attrs: argmemonly nofree norecurse nosync nounwind
	define i32 @func(double* noalias nocapture writeonly %retptr, { i8*, i32, i8*, i8*, i32 }** noalias nocapture readnone %excinfo, double %arg.t, i8* nocapture readnone %arg.arr.0, i8* nocapture readnone %arg.arr.1, i64 %arg.arr.2, i64 %arg.arr.3, double* nocapture readonly %arg.arr.4, i64 %arg.arr.5.0, i64 %arg.arr.6.0) local_unnamed_addr #0 {
	B0.endif.endif:
	  %.449 = fcmp oeq double %arg.t, 1.000000e+00
	  %.241 = fcmp ult double %arg.t, 0.000000e+00
	  %.338 = fcmp uge double %arg.t, 2.500000e-01
	  %not.or.cond = or i1 %.241, %.338
	  %_ind.5.1 = sext i1 %not.or.cond to i64
	  %_ind.4.1 = select i1 %.449, i64 2, i64 %_ind.5.1
	  %.241.1 = fcmp oge double %arg.t, 2.500000e-01
	  %.338.1 = fcmp olt double %arg.t, 7.500000e-01
	  %or.cond6 = and i1 %.241.1, %.338.1
	  %_ind.5.1.1 = select i1 %or.cond6, i64 1, i64 %_ind.4.1
[...]

PR #2735, opened 2025-11-04
fix: duplicate global constants

ymardoukhi:
See issue #2706 

This PR should resolve the issue when global constants with `private` linkage get `external` linkage if the LLVM IR is called via `ccall` interface. See the same issue above for detailed examination.


src/compiler/orcv2.jl
```diff
@@ -258,6 +286,13 @@ function add!(mod)
         replace_uses!(f, ptr)
         Compiler.eraseInst(mod, f)
     end
+
+    # rename non-special global constants that
+    # have external linkage (modified by the ccall
```
wsmoses: alternate solution, can we just mark all defined symbols during the lto import phase as internal?

we should already do so for functions

wsmoses: x/ref https://github.com/EnzymeAD/Enzyme.jl/blob/c009fa593821cbd9017450a6721259f7dbc6ebf1/src/compiler/validation.jl#L596

ymardoukhi: Worked like a charm! All the tests passed locally. This also [explains](https://julialang.slack.com/archives/C01J3R4FHB3/p1762264252690679) why something like

```
@.const.array.data.5 = private unnamed_addr constant [48 x i8] c"\00\00\00\00\00\00\00\00\00\00\00\00\00\00\D0?\00\00\00\00\00\00\D0?\00\00\00\00\00\00\E8?\00\00\00\00\00\00\E8?\00\00\00\00\00\00\F0?", align 8
```

turns into

```
@.const.array.data.5 = dso_local unnamed_addr constant [48 x i8] c"\00\00\00\00\00\00\00\00\00\00\00\00\00\00\D0?\00\00\00\00\00\00\D0?\00\00\00\00\00\00\E8?\00\00\00\00\00\00\E8?\00\00\00\00\00\00\F0?", align 8
```

What was the initial thought behind changing the linkage to external in the first place?

vchuravy:
So one challenge here is that we are hitting different language semantics. Julia code doesn't use LLVM globals as much, except for compilation unit local operations. 

In C/C++ that is not the case and I have to answer the question if a global variable with the same name ought to be unifed or not. This is AD independent and is normally handled by the dynamic linker, here we are consuming LLVM IR from before the dynamic linker phase and I think we must met the semantics.

ymardoukhi:
> So one challenge here is that we are hitting different language semantics. Julia code doesn't use LLVM globals as much, except for compilation unit local operations.
> 
> In C/C++ that is not the case and I have to answer the question if a global variable with the same name ought to be unifed or not. This is AD independent and is normally handled by the dynamic linker, here we are consuming LLVM IR from before the dynamic linker phase and I think we must met the semantics.

That clarifies the intent. Thanks!

But what confused me earlier was that no matter how I defined the linkage in my LLVM IR, it was overwritten when the `llvmbc` was loaded from the shared object. I think it must be the user's sole responsibility to make sure how the globals are handled? If I used `Base.llvmcall`, I could see that the LLVM IR remained unaltered.

wsmoses:
> > So one challenge here is that we are hitting different language semantics. Julia code doesn't use LLVM globals as much, except for compilation unit local operations.
> > In C/C++ that is not the case and I have to answer the question if a global variable with the same name ought to be unifed or not. This is AD independent and is normally handled by the dynamic linker, here we are consuming LLVM IR from before the dynamic linker phase and I think we must met the semantics.
> 
> That clarifies the intent. Thanks!
> 
> But what confused me earlier was that no matter how I defined the linkage in my LLVM IR, it was overwritten when the `llvmbc` was loaded from the shared object. I think it must be the user's sole responsibility to make sure how the globals are handled? If I used `Base.llvmcall`, I could see that the LLVM IR remained unaltered.

yeah so I think probably the thing to do is to mark it as external, but remove initialization assuming it will be initialized by the actual library itself [therefore preserving the one definiton, and not overwriting]. alternatively we can link it weakly/etc

PR #2740, opened 2025-11-05
Implement fix_ptr_lookup for raw fname

vchuravy:


.github/workflows/Integration.yml
```diff
@@ -53,6 +53,7 @@ jobs:
           - SciML
           - KernelAbstractions
           - Molly
+          - MPI
```
wsmoses, line 56: presumably if isolating this, we don't need integration test in this PR?

wsmoses (approved):
otherwise lgtm

PR #2742, opened 2025-11-05
Test MPI.Recv/MPI.Send

vchuravy:
Needs an Enzyme_jll bump at least


vchuravy:
When I am doing stacked branches like that, reading main is better than merging main

wsmoses:
yeah that was my bad, I didn't realize it was a stacked branch

PR #2745, opened 2025-11-05
Add MPI extension with Allreduce! forward rule

lcw:


ext/EnzymeMPIExt.jl
```diff
@@ -0,0 +1,37 @@
...
+
+function EnzymeRules.forward(config, ::Const{typeof(MPI.Allreduce!)}, rt, v, op::Const, comm::Const)
+    op = op.val
+    comm = comm.val
+
+    if !(op == MPI.SUM || op == +)
+        error("Forward mode MPI.Allreduce! is only implemented for MPI.SUM.")
+    end
+
+    if EnzymeRules.needs_primal(config)
```
wsmoses, line 16: The needs primal only is relevant to returning the result, if something is mutated it still needs to happen unless the argument is marked noneed

wsmoses (approved):

ext/EnzymeMPIExt.jl
```diff
@@ -0,0 +1,37 @@
+module EnzymeMPIExt
+
+using MPI
+using Enzyme
+
+import Enzyme.EnzymeCore: EnzymeRules
+
+function EnzymeRules.forward(config, ::Const{typeof(MPI.Allreduce!)}, rt, v, op::Const, comm::Const)
```
wsmoses, line 8: for sake of understanding, is there a reason why this wasnt handled already llvm-side?

vchuravy:
@lcw and I also realized that this may not be needed for the application we look at.
(We may only need adjoint MPI_Allreduce)

ext/EnzymeMPIExt.jl
vchuravy: On Julia 1.10

```
  %14 = call i32 @PMPI_Allreduce(i64 %6, i64 %bitcast_coercion, i32 noundef 1, i32 %9, i32 %11, i32 %13) #12 [ "jl_roots"({} addrspace(10)* %2, {} addrspace(10)* %1, {} addrspace(10)* addrspacecast ({}* inttoptr (i64 140676541099488 to {}*) to {} addrspace(10)*), {} addrspace(10)* %0) ], !dbg !25
0x3bb3f008
Unhandled MPI FUNCTION
UNREACHABLE executed at /workspace/srcdir/Enzyme/enzyme/Enzyme/CallDerivatives.cpp:2211!

[63249] signal (6.-6): Aborted
in expression starting at REPL[15]:1
unknown function (ip: 0x7ff25629894c)
gsignal at /usr/lib/libc.so.6 (unknown line)
abort at /usr/lib/libc.so.6 (unknown line)
_ZN4llvm25llvm_unreachable_internalEPKcS1_j at /home/vchuravy/.julia/juliaup/julia-1.10.10+0.x64.linux.gnu/bin/../lib/julia/libLLVM-15jl.so (unknown line)
handleMPI at /workspace/srcdir/Enzyme/enzyme/Enzyme/CallDerivatives.cpp:2211
handleKnownCallDerivatives at /workspace/srcdir/Enzyme/enzyme/Enzyme/CallDerivatives.cpp:2249
visitCallInst at /workspace/srcdir/Enzyme/enzyme/Enzyme/AdjointGenerator.h:6402
visit at /opt/x86_64-linux-gnu/x86_64-linux-gnu/sys-root/usr/local/include/llvm/IR/InstVisitor.h:111 [inlined]
CreateForwardDiff at /workspace/srcdir/Enzyme/enzyme/Enzyme/EnzymeLogic.cpp:5062
EnzymeCreateForwardDiff at /workspace/srcdir/Enzyme/enzyme/Enzyme/CApi.cpp:661
EnzymeCreateForwardDiff at /home/vchuravy/.julia/packages/Enzyme/rsnI8/src/api.jl:342
```

wsmoses: okay we should fix that needs https://github.com/EnzymeAD/Enzyme/pull/2530

we may also need callingconv stuff like
https://github.com/EnzymeAD/Enzyme/blob/2e6f771cf4570d2800a67e9da9298c0d612d5020/enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp#L5076 and perhaps elsewhere

ext/EnzymeMPIExt.jl
```diff
@@ -0,0 +1,37 @@
...
+    if EnzymeRules.needs_primal(config)
+        MPI.Allreduce!(v.val, op, comm)
+    end
+
+    if EnzymeRules.width(config) == 1
+        MPI.Allreduce!(v.dval, op, comm)
+    else
+        # would be nice to use MPI non-blocking collectives
+        foreach(v.dval) do dval
+            MPI.Allreduce!(dval, op, comm)
```
michel2323, line 25: Are the `dval`s really non-contiguous? If so, I'd think about a derived datatype for `v.dval` to get this into one `Allreduce` and get the benefits of vector-mode.

michel2323 (approved):

ext/EnzymeMPIExt.jl
vchuravy: Yes the dvals are all independent memory allocations.

The structure here is a `NTuple{N, Vector{Float64}}`, I have been wanting a variant where we use continous memory, but we can only do that from 1.11 and we likely can't gurantuee it.

I was thinking about derived types as well, but I have not yet managed to convince MPI to understand a `Vector{Float64}`

Issue #2733, opened 2025-11-04
Problem generating thunk for function


kshyatt:
I'm trying to compare the results of running the reverse pass for two functions. I need to do this because FD doesn't work for my tests for one of them (long story).

I'm working off of `test_reverse` in `EnzymeTestUtils`.

Here's what I have so far:

```julia
function test_pullbacks_match(rng, f!, f, A, args, Δargs, alg=nothing)                                    
    ΔA = randn(rng, eltype(A), size(A)...)                                                                
    copy_activities = (Duplicated(A, copy(ΔA)), Duplicated(copy(args[1]), copy(Δargs[1])), Duplicated(copy(args[2]), copy(Δargs[2])))
    inplace_activities = (Duplicated(A, copy(ΔA)), Duplicated(copy(args[1]), copy(Δargs[1])), Duplicated(copy(args[2]), copy(Δargs[2])))

    mode        = EnzymeTestUtils.set_runtime_activity(ReverseSplitWithPrimal, false)                     
    copy_act    = Const(f)                                                                                
    inplace_act = Const(f!)
    forward_copy, reverse_copy = autodiff_thunk(                                                          
        mode, typeof(copy_act), Duplicated, map(typeof, copy_activities)...                               
    )
    forward_inplace, reverse_inplace = autodiff_thunk(
        mode, typeof(inplace_act), Duplicated, map(typeof, inplace_activities)...                         
    )
    dx_copy_ad    = only(reverse_copy(copy_act, copy_activities..., nothing))
    dx_inplace_ad = only(reverse_copy(inplace_act, inplace_activities..., nothing))                       
    # check all returned derivatives between copy & inplace 
    for (i, (copy_act_i, inplace_act_i)) in enumerate(zip(copy_activities, inplace_activities))           
        msg_deriv = "shadow derivative for argument $(i - 1) should match between copy and inplace"       
        EnzymeTestUtils.test_approx(copy_act_i.dval, inplace_act_i.dval, msg_deriv)
    end 
    return
end 
```
`f` and `f!` are essentially
`f!(A, args) = qr!(A, args)` (stores the result in `args`) and `f(A, args) = f!(copy(A), args)`.

But when I try to run this I get the following:
```Julia
  Got exception outside of a @test
  MethodError: no method matching thunk(::Expr)
  The function `thunk` exists, but no method is defined for this combination of argument types.
  Closest candidates are:
[...]

PR #2768, opened 2025-11-10
fix primal return type methoderror

oscardssmith:
mostly fixes https://github.com/EnzymeAD/Enzyme.jl/issues/2733 (but hits the following error because I'm metaprogramming badly somehow)
```
ERROR: UndefVarError: `ft` not defined in `Enzyme.Compiler`
Suggestion: check for spelling errors or missing imports.
Stacktrace:
 [1] primal_return_type(mode::ForwardMode{false, FFIABI, false, false, false}, ft::Type{typeof(identity)}, tt::Type{Tuple{}})
   @ Enzyme.Compiler ~/.julia/dev/Enzyme/src/typeutils/inference.jl:139
 [2] autodiff(mode::CMode, f::FA, args::Vararg{Annotation, Nargs}) where {FA<:Annotation, CMode<:Mode, Nargs}
   @ Enzyme ~/.julia/dev/Enzyme/src/Enzyme.jl:556
 [3] autodiff(mode::CMode, f::F, args::Vararg{Annotation, Nargs}) where {F, CMode<:Mode, Nargs}
   @ Enzyme ~/.julia/dev/Enzyme/src/Enzyme.jl:534
 [4] top-level scope
   @ REPL[3]:1
 ```

### DIFF
--- a/src/typeutils/inference.jl
+++ b/src/typeutils/inference.jl
@@ -109,18 +109,20 @@ function primal_return_type_generator(world::UInt, source, self, @nospecialize(m
    
     mi = my_methodinstance(mode, ft, tt, world, min_world, max_world)
 
-    slotnames = Core.svec(Symbol("#self#"), :mode, :ft, :tt)
-    stub = Core.GeneratedFunctionStub(
-        primal_return_type,
-        slotnames,
-        Core.svec(),
-    )
-    mi === nothing && return stub(world, source, :(throw(MethodError(ft, tt, $world))))
-
-    result = primal_return_type_world(mode, world, mi)
-    code = Any[Core.Compiler.ReturnNode(result)]
-    # create an empty CodeInfo to return the result
-    ci = create_fresh_codeinfo(primal_return_type, source, world, slotnames, code)
+    if mi === nothing
+        result = Union{}
+        code = Any[Expr(:call, Base.throw, :(MethodError(ft, tt))),
+        Core.Compiler.ReturnNode(result)]
+        # create an empty CodeInfo to return the result
+        slotnames = Core.svec(Symbol("#self#"), :ft, :tt)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by vchuravy** [comment]
This seems wrong to me, do see #2663 for the change that likely broke the behavior here.

In particular: https://github.com/EnzymeAD/Enzyme.jl/pull/2663/files#diff-a5ddd0c06b3228d600577533c56e26161c6952c6b5c89b4befb5fc6b66ffce45L104-L117

File: src/typeutils/inference.jl [hunk 1]
```diff
@@ -109,27 +109,30 @@ function primal_return_type_generator(world::UInt, source, self, @nospecialize(m
...
-    code = Any[Core.Compiler.ReturnNode(result)]
-    # create an empty CodeInfo to return the result
-    ci = create_fresh_codeinfo(primal_return_type, source, world, slotnames, code)
+    if mi === nothing
+        result = Union{}
+        code = Any[Expr(:call, Base.throw, :(MethodError(ft, tt))),
+        Core.Compiler.ReturnNode(result)]
+        # create an empty CodeInfo to return the result
+        slotnames = Core.svec(Symbol("#self#"), :ft, :tt)
+        ci = create_fresh_codeinfo(primal_return_type, source, world, slotnames, code)
```
vchuravy, line 118: No we used `stub` successfully here before.

oscardssmith:
The thing that doesn't make sense to me here with the old code is that you were stubbing `primal_return_type` when you want to just throw a `MethodError`? Also, why is `GeneratedFunctionStub` a thing we want to be using here at all?

vchuravy:
We are writing a generator for a generated function, so we can't throw a MethodError right here and there. We want to throw a method error from the generated function.

> Also, why is GeneratedFunctionStub a thing we want to be using here at all?

Because that's the supported interface?

vchuravy:
Correct fix in #2771

PR #2772, opened 2025-11-11
Implement our own LookupError

vchuravy:


oscardssmith:
why?

vchuravy:
Because you can't throw a method error using only the type of the function?

wsmoses (commented):
could we add a test with this to make sure it doesn't decay again?

wsmoses (approved):

oscardssmith:
@vchuravy ah, that makes sense then.

PR #2784, opened 2025-11-14
CI Integration Example SpeedyWeather.jl

maximilian-gelbrecht:
@vchuravy suggested I add an example using SpeedyWeather.jl to the integration tests. 

The test is pretty much the same thing we do in the paper. It's a sensitivity analysis of a single grid point and it checks that this runs without errors and the gradient makes physical sense, so the gradient is localised around the selected grid point and small far away from it.


vchuravy:
Thanks Max!

vchuravy:
@giordano can we make the filtering more aggressive and only run CI on the integration test being changed?

giordano:
We could generate the matrix dynamically, but only if `test/integration/` files only have been modified, otherwise I think we want to run all the tests, right?

giordano:
BTW, this is missing adding SpeedyWeather to https://github.com/EnzymeAD/Enzyme.jl/blob/ab96c82667b3999ac624dfff6eaabdbc376d40cc/.github/workflows/Integration.yml#L47-L56

maximilian-gelbrecht:
Oh, yes. I just added it.

test/integration/SpeedyWeather/runtests.jl
```diff
@@ -0,0 +1,64 @@
...
+seed_point = 443    # seed point
+d_diag.grid.temp_grid[seed_point, 8] = 1
+
+# Sensitivity Analysis of Temperature at a single grid point (one-hot seed)
+autodiff(Enzyme.Reverse, checkpointed_timesteps!, Const, Duplicated(progn, d_progn), Duplicated(diagn, d_diag), Duplicated(model, d_model), Const(N), Const(checkpoint_scheme))
+
+vor_grid = transform(d_progn.vor[:, :, 2], model.spectral_transform)
+
+# nonzero
+@test sum(abs.(vor_grid)) > 0
```
giordano:
```suggestion
@test sum(abs, vor_grid) > 0
```

test/integration/SpeedyWeather/Project.toml
```diff
@@ -0,0 +1,4 @@
+[deps]
+Checkpointing = "eb46d486-4f9c-4c3d-b445-a617f2a2f1ca"
+Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
+SpeedyWeather = "9e226e20-d153-4fed-8a5b-493def4f21a9"
```
giordano, line 3:
```suggestion
[deps]
Checkpointing = "eb46d486-4f9c-4c3d-b445-a617f2a2f1ca"
Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
SpeedyWeather = "9e226e20-d153-4fed-8a5b-493def4f21a9"

[sources]
Enzyme = { path = "../../.." }

[compat]
Checkpointing = "0.11"
SpeedyWeather = "=0.17.4"
```

giordano:
Besides the fact `Test` isn't loaded, the jobs are timing out.  Any way to have something lighter to test?  All other jobs finish in less than 20 minutes (most of them are a lot faster)

maximilian-gelbrecht:
Yeah, in our own CI a very similar test takes about 35 minutes. I hoped it's a bit faster now after recent Enzyme patches, but it doesn't seem to be the case. 

If 20 min is the limit (didn't know that), then I can turn off some of the parametrizations and see if that's sufficient. I am out of time to really experiment with that for today though.

giordano:
Timeout is 45 minutes https://github.com/EnzymeAD/Enzyme.jl/blob/ab96c82667b3999ac624dfff6eaabdbc376d40cc/.github/workflows/Integration.yml#L32 but all other jobs take less than half of that.

maximilian-gelbrecht:
I mean the gradient compile time is what it is currently 🤷
On my laptop and my HPC it's a bit faster, but in the GitHub Actions CI it's really quite slow.

maximilian-gelbrecht:
I updated the example to mirror more closely the one we already use in your CI in Speedy. It should be faster now, and there it always finishes < 40 min. 

However, it also seems that Enzymev0.13.105 broke something for us, as we are getting fails from our CI tests on this since Friday. So this example here will also currently fail.

wsmoses:
well thats certainly a reason for getting the integration tests in!

wsmoses:
that said the erring part is a bit too complex/baked in, would you be able to construct a MWE of?

wsmoses:
okay I think I fixed the custom rule issue immediately here (and a potential PR for more information on the 1.11: https://github.com/EnzymeAD/Enzyme/pull/2582).

However, on a debug julia I pretty much always hit GC errors. @vchuravy I think we really ought debug and fix that MWE speedweather GC error issue before we can land this

maximilian-gelbrecht:
Thanks for your effort on this already. I already had a debugging session today with this, but it's quite nasty to boil this down to a proper MWE. I know which intermediate function from our code causes the error, but each individual function it calls is fine.  This is the [function](https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/97d2656bc15de17d7db8473caf20ad062d40e066/src/dynamics/tendencies.jl#L775-811) for our simplest model, and here's an [example call with Enzyme](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/929/files#diff-83ecb425653adae5d9fc26a03955444d85596ff7dcbada2a643a7299791a43bcR153), but in the next days I'll try to condense it to a proper MWE, it's just not that easy to boil this down so quickly.

maximilian-gelbrecht:
I continue to struggle a bit to come up with a MWE that doesn't use our data structures. 

Below is the simplest function in our code that produces the error. Each of the individual functions I can differentiate just fine, it's the interplay between them that seemingly causes an error here. 

```Julia 
using Enzyme, SpeedyWeather

function transform_debug!(
    diagn::DiagnosticVariables,   
    progn::PrognosticVariables,
    lf::Integer,
    model::Barotropic;
    kwargs...
)    
    # retrieve data
    (; vor_grid, u_grid, v_grid ) = diagn.grid
    (; scratch_memory) = diagn.dynamics  

    vor = get_step(progn.vor, lf)   # relative vorticity at leapfrog step lf
    U = diagn.dynamics.a            # reuse work arrays for velocities in spectral
    V = diagn.dynamics.b            # reuse work arrays for velocities in spectral
                                    # U = u*coslat, V=v*coslat
    S = model.spectral_transform

    # COMPUTE FUNCTIONS, WHEN At least 3 are active -> GC error 
    # that's  a spherical harmonics transforms (KA kernel + FFT)
    transform!(vor_grid, vor, scratch_memory, S)    # get vorticity on grid from spectral vor
    
    # that's a spatial derivative computed in spectral space (KA kernel) 
    SpeedyWeather.UV_from_vor!(U, V, vor, S)
    
    # that's a spherical harmonics transforms and some rescaling (KA kernel + FFT + KA kernel)
    transform!(u_grid, U, scratch_memory, S, unscale_coslat=true)
    #transform!(v_grid, V, scratch_memory, S, unscale_coslat=true)

    return nothing
end

spectral_grid = SpectralGrid(trunc=9, nlayers=1)
model = BarotropicModel(; spectral_grid)
simulation = initialize!(model)
progn, diagn, model = SpeedyWeather.unpack(simulation)

lf2 = 2 

dprogn = make_zero(progn)
ddiagn = make_zero(diagn)

autodiff(Reverse, transform_debug!, Const, Duplicated(diagn, ddiagn), Duplicated(progn, dprogn), Const(lf2), Const(model))
```
which causes 

```
[29800] signal (11.2): Segmentation fault: 11
in expression starting at REPL[12]:1
_ZN4llvm5Value11setNameImplERKNS_5TwineE at /Users/max/.julia/juliaup/julia-1.10.10+0.aarch64.apple.darwin14/lib/julia/libLLVM.dylib (unknown line)
Allocations: 66874844 (Pool: 66783213; Big: 91631); GC: 92
Segmentation fault: 11
```

and was fine before with Enzyme <0.13.105

maximilian-gelbrecht:
Is #2818 related to this here? Just let me know if I should invest a bit more time for finding a really proper MWE, or if that's already it.

wsmoses:
No that's unrelated and an issue from an in progress PR which doesn't occur on main atm

maximilian-gelbrecht:
The whole issue just confuses me. 

Now, I found that just leaving away the `kwargs...` in the function signature and the `@boundscheck` in the function makes it work. The individual functions are all fine to differentiate as well, and the custom rules that are used for the `_fourier!` function also pass their tests and are fine to differentiate in other contexts. 

So in the following script, using some of SpeedyWeather's numerics I just present the just slightly different versions where the last one that includes an unused `kwargs...` and a `@boundscheck` causes a SegFault. 

Julia 1.10, Enzyme v0.13.108, SpeedyWeather `main`
```Julia
using Enzyme, SpeedyWeather
 
# this works 
function transform_debug!(
    vor_grid, u_grid, scratch_memory, U, 
    vor, 
    S
)    
    # catch incorrect sizes early
    @boundscheck SpeedyWeather.ismatching(S, vor_grid) || throw(DimensionMismatch(S, vor_grid))
    @boundscheck SpeedyWeather.ismatching(S, vor) || throw(DimensionMismatch(S, vor))

    g_north = scratch_memory.north 
    g_south = scratch_memory.south 

    # this is a FFT, for which we have a custom rule defined 
    SpeedyWeather.SpeedyTransforms._fourier!(vor_grid, g_north, g_south, S)

    # catch incorrect sizes early
    @boundscheck SpeedyWeather.ismatching(S, u_grid) || throw(DimensionMismatch(S, u_grid))
    @boundscheck SpeedyWeather.ismatching(S, U) || throw(DimensionMismatch(S, U))

    g_north = scratch_memory.north 
    g_south = scratch_memory.south 

    SpeedyWeather.SpeedyTransforms._fourier!(u_grid, g_north, g_south, S)

    return nothing
    
end

# this works 
function transform_debug_with_kwargs_no_checks!(
    vor_grid, u_grid, scratch_memory, U, 
    vor, 
    S; 
    kwargs... # now has an unused kwargs...
)   
    # but no @boundschecks 
    g_north = scratch_memory.north 
    g_south = scratch_memory.south 

    # this is a FFT, for which we have a custom rule defined 
    SpeedyWeather.SpeedyTransforms._fourier!(vor_grid, g_north, g_south, S)

    g_north = scratch_memory.north 
    g_south = scratch_memory.south 

    SpeedyWeather.SpeedyTransforms._fourier!(u_grid, g_north, g_south, S)

    return nothing
end

# this gives a segfault 
function transform_debug_with_kwargs!(
    vor_grid, u_grid, scratch_memory, U, 
    vor, 
    S; 
    kwargs... # still has an unused kwargs...
)   
    # boundschecks and kwarg together cause an error here now 
    @boundscheck SpeedyWeather.ismatching(S, vor_grid) || throw(DimensionMismatch(S, vor_grid))
    @boundscheck SpeedyWeather.ismatching(S, vor) || throw(DimensionMismatch(S, vor))

    # that's  a spherical harmonics transforms (KA kernel + FFT)
    g_north = scratch_memory.north 
    g_south = scratch_memory.south 

    # this is a FFT for which we have a custom rule defined
    SpeedyWeather.SpeedyTransforms._fourier!(vor_grid, g_north, g_south, S)

    @boundscheck SpeedyWeather.ismatching(S, u_grid) || throw(DimensionMismatch(S, u_grid))
    @boundscheck SpeedyWeather.ismatching(S, U) || throw(DimensionMismatch(S, U))

    g_north = scratch_memory.north 
    g_south = scratch_memory.south 

    SpeedyWeather.SpeedyTransforms._fourier!(u_grid, g_north, g_south, S)

    return nothing
end

spectral_grid = SpectralGrid(trunc=9, nlayers=1)
S = SpectralTransform(spectral_grid)
#simulation = initialize!(model)

# these are RingGrids.Fields 
vor_grid = rand(spectral_grid.grid, 1)
u_grid = rand(spectral_grid.grid, 1)
v_grid = rand(spectral_grid.grid, 1)
scratch_memory = deepcopy(S.scratch_memory)

# these are LowerTriangularArrays.LowerTriangularArray
vor = rand(ComplexF32, spectral_grid.spectrum, 1)  
U =  rand(ComplexF32, spectral_grid.spectrum, 1)  

# this works 
autodiff(Reverse, transform_debug!, Const, Duplicated(vor_grid, make_zero(vor_grid)), Duplicated(u_grid, make_zero(u_grid)), Duplicated(scratch_memory, make_zero(scratch_memory)), Duplicated(U, make_zero(U)), Duplicated(vor, make_zero(vor)), Const(S))

# this works as well 
autodiff(Reverse, transform_debug_with_kwargs_no_checks!, Const, Duplicated(vor_grid, make_zero(vor_grid)), Duplicated(u_grid, make_zero(u_grid)), Duplicated(scratch_memory, make_zero(scratch_memory)), Duplicated(U, make_zero(U)), Duplicated(vor, make_zero(vor)), Const(S))

# this gives a SegFault
autodiff(Reverse, transform_debug_with_kwargs!, Const, Duplicated(vor_grid, make_zero(vor_grid)), Duplicated(u_grid, make_zero(u_grid)), Duplicated(scratch_memory, make_zero(scratch_memory)), Duplicated(U, make_zero(U)), Duplicated(vor, make_zero(vor)), Const(S))
```

maximilian-gelbrecht:
I've updated the example above and I am honestly just really confused by it. This was fine before, but now seemingly some interaction between our custom rules for the FFT, a `@boundscheck` and an unused `kwargs...` causes a SegFault, but also just in a configuration like that, with calling `_fourier!` twice. In other contexts all of this is fine. 

The custom rule is actually implemented [here](https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/main/SpeedyTransforms/ext/SpeedyTransformsEnzymeExt.jl)

maximilian-gelbrecht:
I was looking at this again just now. Whatever the exact problem was, with Enzyme v0.13.121 I don't get the error anymore and the example above doesn't give a SegFault anymore 🎉 .

I'll sit down and revise the example as soon as I find some time for it. 

In parallel we'll also give Reactant a shot, but that's still a bit in the future.

wsmoses:
that actually makes sense since that PR did resolve a segfault related issue

PR #2785, opened 2025-11-14
[CI] Dynamically generate list of packages to test

giordano:
Very incomplete, but this is an initial proof-of-concept of https://github.com/EnzymeAD/Enzyme.jl/pull/2784#issuecomment-3533218382, still need to do the filtering.

### DIFF
--- a/.github/workflows/Integration.yml
+++ b/.github/workflows/Integration.yml
@@ -32,21 +32,21 @@ jobs:
     name: Generate list of packages for integration tests
     runs-on: ubuntu-latest
     outputs:
-      matrix: ${{ steps.set-matrix.outputs.matrix }}
+      packages: ${{ steps.set-packages.outputs.packages }}
     steps:
       - uses: actions/checkout@v5
+      - uses: julia-actions/setup-julia@v2
       - name: Set packages
         id: set-packages
+        shell: julia --color=yes {0}
         run: |
-          set -e
-          shopt -s globstar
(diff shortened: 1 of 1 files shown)


.github/workflows/Integration.yml
```diff
@@ -28,7 +28,28 @@ concurrency:
...
+      packages: ${{ steps.set-packages.outputs.packages }}
+    steps:
+      - uses: actions/checkout@v5
+      - uses: julia-actions/setup-julia@v2
+      - name: Set packages
+        id: set-packages
+        shell: julia --color=yes {0}
+        run: |
+          all_pkgs = readdir("test/integration")
+          filtered_pkgs = all_pkgs
```
giordano, line 44: Ok, this is now working, but here we need to define a criterion for filtering the packages, probably something like what we do in Yggdrasil (but in Yggdrasil we only allow for squash-merge, which makes diff-ing on `main` easier)

wsmoses:
I would actually recommend not doing this, running the integration tests for all enzyme code has caught [at least for me] many bugs already

giordano:
As I said in https://github.com/EnzymeAD/Enzyme.jl/pull/2784#issuecomment-3533218382, the idea is to do this only when integration only tests are changed. It doesn't make much sense to run all the integration tests when only one new is added.

wsmoses:
ah got it, sorry missed that

PR #2791, opened 2025-11-15
Relax array type restriction inside `gradient!`

gdalle:
The `<:Array` constraints is preventing the use of `Enzyme.gradient!` on many different types of arrays, including Reactant's

### DIFF
--- a/src/sugar.jl
+++ b/src/sugar.jl
@@ -405,7 +405,7 @@ gradient!(ReverseWithPrimal, dx, f, [2.0, 3.0])
     dx::X,
     f::F,
     x::X,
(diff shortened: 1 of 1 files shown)


gdalle:
Unfortunately we don't have a good way to recognize mutable arrays. However, if you pass an SArray this will error anyway, so it is not too permissive

gdalle:
We're running into one of Julia's main weaknesses, which is the lack of traits and interfaces. If you want broad compatibility, you sometimes have to overpromise a little. Allowing `AbstractArray` in a dispatch is not quite the same as saying "this will work on any array whatsoever" in my view

wsmoses:
yeah it's just that giving a "function not defined for gradient!" is a nicer error message, but we can fix that up in post.

We can incidentally loosen it up even more so to not even restrict as an array

PR #2810, opened 2025-11-25
Ensure relocatable global variables

wsmoses:


src/compiler/orcv2.jl
```diff
@@ -30,9 +30,11 @@ end
 
 const jit = Ref{CompilerInstance}()
 const tm = Ref{TargetMachine}() # for opt pipeline
+const dylib = Ref{JITDylib}()
 
 get_tm() = tm[]
 get_jit() = jit[].jit
+get_dylib() = dylib[]
```
vchuravy: I would just define it as:

`get_dylib() = JITDylib(lljit)`

vchuravy: Where is this actually used?

wsmoses: it was used [I tried to do the llvm.define when we added to our list, but that led to issues], so now just doing a pass over ir doing the rewrite.

removed since

wsmoses:
@vchuravy interestingly [on a debug build] I can't repro the ci failures. any luck on your side?

src/compiler/orcv2.jl
```diff
@@ -173,6 +173,40 @@ function add_trampoline!(jd, (lljit, lctm, ism), entry, target)
...
+        ptr = LLVM.const_inttoptr(ptr, LLVM.PointerType(LLVM.function_type(f)))
+        replace_uses!(f, ptr)
+        Compiler.eraseInst(mod, f)
+    end
+    for g in collect(globals(mod))
+	if !startswith(LLVM.name(g), "ejl_inserted\$")
+           continue
+        end
+	_, ogname, load1, initaddr = split(LLVM.name(g), "\$")
+
```
wsmoses, line 193: @vchuravy here we have the original name of the global variable that julia emitted it as [as well as the initial address/etc]. If there's any way we can lookup a portable version of this from the name, we have our relocatable code. E.g. if I could cglobal(:jl_global_7353) [where just the name was consistent but the runtime pointer differed, we'd be done]

wsmoses:
@vchuravy @gbaraldi @oscardssmith are there any jl_global's passed as a ptr{ptr{jlvaluet}} or ptr{jlaluet} that are different to load dependeing on what thread you're on?

src/compiler/orcv2.jl
vchuravy: No, we can't rely on names. Names are currently session local and in the [future compilation local](https://github.com/JuliaLang/julia/pull/60031)

On the Julia side this rougly looks like:

We are building a list of pointers (GOT like) that will be initialized during the loading process, we create a side table of `Julia Object` => `Index in GOT`, which will be serialized during the saving of a package image, upon loading we deserialize that table (basically an IDDict)  and populate the GOT.

Now this might be something GPUCompiler doesn't yet fully expose.

vchuravy:
> jl_global's passed as a ptr{ptr{jlvaluet}} or ptr{jlaluet} that are different to load dependeing on what thread you're on?

I don't understand the question? The only object that is different to load is the ptls

wsmoses:
Resolved ish, but a question I have is some jlvaluet are passed a s a pointer to the value others as a pointer to the pointer.

Sometimes these aren't initialized and the inner ptr is null and or the final value is nothing. I've reverse engineered that behavior and we don't load those such values but is good to confirm

src/compiler/orcv2.jl
wsmoses: Yeah if we can somehow get access to that dict directly we could replace this with the correct object per session.

We actually store all the objects right now in our own dict, but it then begs the question of how we keep that dict relocatable to different sessions

vchuravy:
So GPU Compiler currently does: https://github.com/JuliaGPU/GPUCompiler.jl/blob/531dccd906cb14b9ff7fc7951ab13243bda24433/src/jlgen.jl#L751-L774

But I think we would need https://github.com/JuliaLang/julia/blob/ee6bb20bc86ec6fc9905e2769293ce1da4c0c4cf/src/aotcompile.cpp#L112-L143 to get a mapping from GlobalVariable to Julia value.

vchuravy:
> Sometimes these aren't initialized and the inner ptr is null and or the final value is nothing. I've reverse engineered that behavior and we don't load those such values but is good to confirm

Could those be jl_binding_t? https://github.com/JuliaLang/julia/blob/ee6bb20bc86ec6fc9905e2769293ce1da4c0c4cf/src/aotcompile.cpp#L128

wsmoses:
No they are not and we explicitly check for those (which appear to only happen on 1.10)

wsmoses:
okay figured it out, we needed to add an atomic to the load

Issue #2813, opened 2025-11-25
TRSM support


vchuravy:
Julia 1.12 switched from `LAPACK.trtrs` to `BLAS.trsm` https://github.com/JuliaLang/LinearAlgebra.jl/pull/1194

Which has some support in Enzyme upstream https://github.com/EnzymeAD/Enzyme/blob/a13f632e2fbaed1f971de015f15e8f4b353e66cb/enzyme/Enzyme/BlasDerivatives.td#L705-L713 but is lacking gradient definitions.

cc: @michel2323 @amontoison

jlperla:
Any update on this?  There is a comment about Julia reverting that original change (https://github.com/JuliaLang/LinearAlgebra.jl/commit/0a79d737dffe7e245f7cf5906d1a5b19ceec26dc) but no back porting?

Is the idea to wait until a back port to Julia 1.12 instead of adding in BLAS.trsm support?

PR #2814, opened 2025-11-25
trsm is not actually implemented yet

vchuravy:
Partial fix to #2813

The crux is that Julia used a different function and while `trsm` is defined in Enzyme proper there aren't currently any rules attached https://github.com/EnzymeAD/Enzyme/blob/a13f632e2fbaed1f971de015f15e8f4b353e66cb/enzyme/Enzyme/BlasDerivatives.td#L705-L713

### DIFF
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -5074,7 +5074,7 @@ function GPUCompiler.compile_unhooked(output::Symbol, job::CompilerJob{<:EnzymeT
         "trmv",
         "syrk",
         "trmm",
-        "trsm",
+        # "trsm", Not actually implemented yet
         "potrf",
     )
     ForwardModeTypes = ("s", "d", "c", "z")


wsmoses:
incidentally that was actually reverted: https://github.com/JuliaLang/LinearAlgebra.jl/pull/1239

but apparently not backported yet

src/compiler.jl
```diff
@@ -5074,7 +5074,7 @@ function GPUCompiler.compile_unhooked(output::Symbol, job::CompilerJob{<:EnzymeT
         "trmv",
         "syrk",
         "trmm",
-        "trsm",
+        # "trsm", Not actually implemented yet
```
wsmoses, line 5077: sorry quick post commit review, can this be limited to 1.12+.

It will otherwise slow down other code by making it non-blas, if not differentiated

vchuravy: Oof, that's frustrating, shouldn't we ideally not do the blas replacement when a function is not being differentiated? Otherwise this problem occurs for many more functions?

wsmoses: it does indeed occur for more functions (which is why we should be somewhat sparing on the blas replacements).

The issue is that at the time of replacement, we haven't run activity analysis so we don't know what is needed and what isn't

PR #2824, opened 2025-12-01
Fix hessian sret type

wsmoses:
requires https://github.com/EnzymeAD/Enzyme/pull/2592


wsmoses:
@copilot isolate a MWE of whatever test is timing out

PR #2832, opened 2025-12-03
Perform sret structuring

wsmoses:


src/llvm/transforms.jl
```diff
@@ -476,6 +476,68 @@ function memcpy_alloca_to_loadstore(mod::LLVM.Module)
     end
 end
 
+# Split a memcpy into an sret with jlvaluet into individual load/stores
+function memcpy_sret_split!(mod::LLVM.Module)
+    dl = datalayout(mod)
+    ctx = context(mod)
+	    sretkind = LLVM.kind(if LLVM.version().major >= 12
```
giordano, line 483: Is the LLVM version check to be extra safe?  Technically, oldest supported version at this point is 15 (the one coming with Julia v1.10), so you could just inline the check here and below.

wsmoses: yeah we just use that everywhere else atm. I suppose we can just do the latter tho at this point

wsmoses:
https://godbolt.org/z/xM6dM5noE

wsmoses:
https://github.com/llvm/llvm-project/blob/2d98a366b79e9cc657a3cf0ee3203f046eab457a/llvm/lib/Analysis/MemoryDependenceAnalysis.cpp#L73 -.-

PR #2863, opened 2025-12-26
Add Turing integration tests (and more Bijectors tests)

penelopeysm:
Merry Christmas! 🙂 

So, I've gotten the AD tests in Turing CI to work with Enzyme. This PR essentially contains a duplicate of those tests. #2528

I haven't timed this exact set of tests but I believe it should take around 15-20 minutes on a GitHub runner.

I would mention that Turing/DynamicPPL exclusively use Enzyme via DI. I could rewrite the first testset to use Enzyme directly without too much faff, but the second one would be very annoying to rewrite and would involve a lot of Turing internals, so I hope that this is okay.

Separately, Bijectors had some tests that used to fail, but this seems to have been fixed recently (I didn't investigate which patch fixed it), so I've added some of those tests here too.

### DIFF
--- a/test/integration/Bijectors/runtests.jl
+++ b/test/integration/Bijectors/runtests.jl
@@ -169,9 +169,14 @@ end
             ),
             (3, 3),
         ),
-        # NOTE(penelopeysm) This requires runtime activity, but forward-mode fails as this
-        # calls gemm! and runtime activity is not yet supported for BLAS calls.
-        sum_b_binv_test_case(Bijectors.PlanarLayer(3), (3, 3); runtime_activity = Reverse, broken = ((v"1.10" <= VERSION < v"1.11") ? Neither : Forward)),
+        # NOTE(penelopeysm) This requires runtime activity on 1.11 reverse-mode, and 1.11
+        # forward-mode fails as this calls gemm! and runtime activity is not yet supported
+        # for BLAS calls. 1.10 works fine without runtime activity.
+        sum_b_binv_test_case(
+            Bijectors.PlanarLayer(3), (3, 3);
+            runtime_activity = ((v"1.10" <= VERSION < v"1.11" ? Neither : Reverse)),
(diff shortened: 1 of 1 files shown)


test/integration/Turing/runtests.jl
```diff
@@ -0,0 +1,87 @@
...
+using Enzyme: Enzyme
+import ForwardDiff
+using StableRNGs: StableRNG
+using Test
+using Turing
+
+adtypes = (
+    AutoEnzyme(; mode = Enzyme.set_runtime_activity(Enzyme.Forward), function_annotation = Enzyme.Const),
+    AutoEnzyme(; mode = Enzyme.set_runtime_activity(Enzyme.Reverse), function_annotation = Enzyme.Const),
+)
```
penelopeysm, line 14: once https://github.com/TuringLang/DynamicPPL.jl/pull/1172 is in it shouldn't need Const.

wsmoses:
1.10 tests fail

penelopeysm:
Took a while to get through the sea of Gibbs sampler code. I've posted a minimiser in an issue, but I also think https://github.com/TuringLang/DynamicPPL.jl/pull/1188 _should_ fix it too. Will push a new commit when that's released!

penelopeysm:
Updated DynamicPPL version; tests on Julia 1.10 work locally for me now.

penelopeysm:
@wsmoses CI passed on the previous commit (at least for the Turing tests; I assume the Comrade tests are not mine to fix 🙂)

wsmoses:
its all green!

one minor thing, in the bijectors tests where we set runtime activity to true, can we first check test_throws EnzymeRuntimeActivityError for that case [that way we confirm its required]

penelopeysm:
Good idea, done that! I also updated the Turing tests since we changed it to not need a closure and hence doesn't need Const

penelopeysm:
I reckon that should all be fine now, tested locally too with both versions 😅

Issue #2873, opened 2026-01-05
Comrade integration correctness issue. 


ptiede:
A bunch of my tests just failed in 1.10. It looks like #2861 is the first instance when this occurred. The commit just looks like something changed in the Enzyme jll. It looks like it is actually a correctness issue in the gradient rather than a segfault.

```julia
  Expression: isapprox(dx, gf; atol = 1.0e-8, rtol = 1.0e-5)
   Evaluated: isapprox([-6499.944012770968, 2172.725237314325, 81934.0195753565, -78543.27840653463, 33120.11938050656, NaN, -276156.81689346454, -1.5723711950501625e6, 68058.33966013379, 34696.55460340206], [-6499.944012737727, 2172.7252373803003, 81934.01957529802, -78543.2784066267, 33120.11938050045, -1.608573813001449e6, -276156.8168933406, -1.5723711950521453e6, 68058.33966050706, 34696.55460314115]; atol = 1.0e-8, rtol = 1.0e-5)
``` 

notice the NaN.

ptiede:
This specific code has something to do with Bessel functions, so I don't know if that is the issue.

wsmoses:
oh er that's bad [also darn 1.12 calling conv changes which should be an NFC for 1.10 but I suppose here we are].

It's not something which is revertable or easily understandable from context, so an MWE is really helpful here.

ptiede:
On it!

ptiede:
@wsmoses @vchuravy, this is really strange, but I can only seem to trigger this when running my unit tests in total. When I run the code, isolated everything is fine. It is also kind of random when things fail. So essentially, I am stuck and don't know exactly how to get an MWE. 

This does seem to happen more often in a `@testset` environment, but beyond that, I am not sure where to look next.

ptiede:
Ok, I figured out one piece of this. I need to evaluate the gradient twice for this to show up consistently. 
Not a MWE, but this is as minimal as I can get it right now.

```julia
using Comrade, Distributions, FiniteDifferences, Enzyme, Test

function test_nonanalytic(intm, vis)
    skym = SkyModel(test_model, test_prior(), imagepixels(μas2rad(150.0), μas2rad(150.0), 256, 256))
    post = VLBIPosterior(skym, intm, vis)

    tpostf = asflat(post)
    x0 = rand(dimension(tpostf))
    dx1 = zero(x0)
[...]

PR #2876, opened 2026-01-06
Update TransformVariables version constraint in Comrade integration

ptiede:
Doesn't fix the correctness error in #2873  but should ensure the more recent version of Comrade is run


wsmoses:
well ironically it now passes

ptiede:
Sadly my tests still fail :'(

PR #2877, opened 2026-01-06
add dependabot for Julia

vchuravy:


.github/dependabot.yml
```diff
@@ -5,3 +5,7 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "weekly"
+  - package-ecosystem: "julia"
+    directory: "/" # Location of package manifests
```
giordano: You need to point dependabot also to the other environments
```suggestion
    # Location of Julia projects
    directories:
      - "/"
      - "lib/EnzymeCore"
      - "lib/EnzymeTestUtils"
```

.github/dependabot.yml
```diff
@@ -5,3 +5,7 @@ updates:
...
+    schedule:
+      interval: "weekly"
```
giordano, line 15:
```suggestion
      interval: "weekly"
    ignore:
      # Ignore package itself
      - dependency-name: "Enzyme"
```

PR #2878, opened 2026-01-06
Fix type unstable active zero

wsmoses:


src/utils.jl
```diff
@@ -486,7 +486,7 @@ end
 
 @inline function typed_fieldtype(@nospecialize(T::Type), i::Int)::Type
     if T <: GenericMemoryRef && i == 1 || T <: GenericMemory && i == 2
-        if T <: GenericMemoryRef && i == 1 && is_memory_ref_field2_an_offset(T)
```
gbaraldi, line 489: What is this checking?

wsmoses: this was an unrelated bugfix i found while minimizing but it's a helper function to determine for a given memory ref whether or not the additional field is an offset or a pointer to the data

gbaraldi: Can you get non concrete GenericMemoryRefs?

PR #2902, opened 2026-01-15
Restore alloca type

wsmoses:
Requires https://github.com/JuliaLang/julia/pull/60695 to be effective.

Fixes numerous 1.12 problems.

Essentially suppose we have a gc alloc of a complex64. LLVMAlloc opt will change this into an [16 x i8]. Then sroa will turn that into 16 individual byte-sized extract/insert -- which we cannot actually handle properly

### DIFF
--- a/src/llvm/attributes.jl
+++ b/src/llvm/attributes.jl
@@ -1,4 +1,5 @@
 const nofreefns = Set{String}((
+    "jl_genericmemory_copyto",
     "utf8proc_toupper",
     "ClientGetAddressableDevices",
     "ClientNumAddressableDevices",
--- a/src/typeutils/lltypes.jl
+++ b/src/typeutils/lltypes.jl
@@ -73,6 +73,37 @@ nfields(Type::LLVM.VectorType) = size(Type)
 nfields(Type::LLVM.ArrayType) = length(Type)
 nfields(Type::LLVM.PointerType) = 1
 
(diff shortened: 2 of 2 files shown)


src/compiler/optimize.jl
```diff
@@ -22,6 +22,7 @@ end
 
 EnzymeAttributorPass() = NewPMModulePass("enzyme_attributor", enzyme_attributor_pass!)
 ReinsertGCMarkerPass() = NewPMFunctionPass("reinsert_gcmarker", reinsert_gcmarker_pass!)
+RestoreAllocaType() = NewPMFunctionPass("restore_alloca_type", restore_alloca_type!)
```
wsmoses, line 25: @vchuravy can you check what obvious thing I'm doing wrong here?

oscardssmith:
With this on 1.12 backports does SciML work?

wsmoses:
Well we need to fix the "I'm silly and didn't register the pass correctly here", then we can test

gbaraldi:
Given https://github.com/JuliaLang/julia/pull/60699 is this still necessary? They will probably be both in the same release regardless

wsmoses:
Yes it's separately necessary to determine the type of the alloca

src/compiler/optimize.jl
vchuravy: `register!(pb, RestoreAllocaType())`

PR #2918, opened 2026-01-19
fix broadcast recursion

oscardssmith:
By compilerizing this we avoid potential Stackoverflows.
TODO: futurize for 1.12 support.


oscardssmith:
This isn't quite working yet.

PR #2923, opened 2026-01-23
Fix strip tracked pointers

wsmoses:


oscardssmith:
FYI this fixed
```
function make_sum_of_solution_u0(prob)
    return function (u0)
        return sum(SciMLBase.__init(prob, Euler(), dt=1.0).u)
    end
end
```
but not
```
function make_sum_of_solution_u0(prob)
    return function (u0)
        return sum(init(prob, Euler(), dt=1.0).u)
    end
end
```

wsmoses:
whats the error not fixed?

oscardssmith:
```
using OrdinaryDiffEqCore: perform_step!
julia> function make_sum_of_solution_u0(prob)
           return function (u0)
               integrator = SciMLBase.__init(prob, BS3())
               ( rand() < 0) && error()
               perform_step!(integrator, integrator.cache)
           return sum(integrator.u)
           end
       end
make_sum_of_solution_u0 (generic function with 1 method)

julia> du01 = DifferentiationInterface.gradient(
           make_sum_of_solution_u0(prob2), backend, [2.0, 1.0]
       )
ERROR: EnzymeNoTypeError: Enzyme cannot statically prove the type of a value being differentiated and risks a correctness error if it gets it wrong.
 Generally this shouldn't occur as Enzyme records type information from julia, but may be expected if you, for example, copy untyped data.
 or alternatively emit very large sized registers that exceed the maximum size of Enzyme's type analysis. If it seems reasonable to differentiate
 this code, open an issue! If the cause of the error is too large of a register, you can request Enzyme increase the size (https://enzyme.mit.edu/julia/dev/api/#Enzyme.API.maxtypeoffset!-Tuple{Any})
 or depth (https://enzyme.mit.edu/julia/dev/api/#Enzyme.API.maxtypedepth!-Tuple{Any}) of its type analysis.
 Alternatively, you can tell Enzyme to take its best guess from context with (https://enzyme.mit.edu/julia/dev/api/#Enzyme.API.looseTypeAnalysis!-Tuple{Any})
 All of these settings are global configurations that need to be set immediately after loading Enzyme, before any differentiation occurs.
 To toggle more information for debugging (needed for bug reports), set Enzyme.Compiler.VERBOSE_ERRORS[] = true (default false)
 Failure within method: MethodInstance for (::var"#make_sum_of_solution_u0##64#make_sum_of_solution_u0##65"{ODEProblem{…}})(::Vector{Float64})
Hint: catch this exception as `err` and call `code_typed(err)` to inspect the errornous code.

Stacktrace:
  [1] getproperty
    @ ~/.julia/packages/SciMLBase/c9FfR/src/integrator_interface.jl:523 [inlined]
  [2] #make_sum_of_solution_u0##64
    @ ./REPL[95]:5 [inlined]
  [3] diffejulia__make_sum_of_solution_u0__64_146578_inner_293wrap
    @ ./REPL[95]:0
  [4] macro expansion
    @ ~/.julia/dev/Enzyme/src/compiler.jl:6672 [inlined]
  [5] enzyme_call
    @ ~/.julia/dev/Enzyme/src/compiler.jl:6151 [inlined]
  [6] CombinedAdjointThunk
    @ ~/.julia/dev/Enzyme/src/compiler.jl:6035 [inlined]
  [7] autodiff
    @ ~/.julia/dev/Enzyme/src/Enzyme.jl:528 [inlined]
  [8] gradient
    @ ~/.julia/packages/DifferentiationInterface/MgcE4/ext/DifferentiationInterfaceEnzymeExt/reverse_onearg.jl:303 [inlined]
  [9] gradient(::var"#make_sum_of_solution_u0##64#make_sum_of_solution_u0##65"{ODEProblem{…}}, ::AutoEnzyme{Nothing, Duplicated}, ::Vector{Float64})
    @ DifferentiationInterface ~/.julia/packages/DifferentiationInterface/MgcE4/src/first_order/gradient.jl:63
 [10] top-level scope
    @ REPL[96]:1
```
removing either `( rand() < 0) && error()` or `perform_step!` fixes this.

wsmoses:
That is a separate error unrelated to this pr

PR #2934, opened 2026-01-27
Add easy_rule for BigFloat division

kshyatt:


kshyatt:
Added some more arithmetic as a lark. The easy rule formulae match what Symbolics gives (thank you @vchuravy for checking) so I am very confused by why `test_forward` is failing

test/rules/internal_rules/bigfloat.jl
```diff
@@ -3,18 +3,31 @@ using EnzymeTestUtils
...
     #test_reverse(+, Active, (a, Active), (b, Active))
     #test_reverse(-, Const, (a, Const), (b, Const))
     #test_reverse(-, Active, (a, Active), (b, Active))
 
-    test_forward(+, Const, (a, Const), (b, Const))
-    test_forward(+, Duplicated, (a, Duplicated), (b, Duplicated))
-    test_forward(-, Const, (a, Const), (b, Const))
-    test_forward(-, Duplicated, (a, Duplicated), (b, Duplicated))
+    for TR in (Const, Duplicated), TA in (Const, Duplicated), TB in (Const, Duplicated)
+        test_forward(+, TR, (a, TA), (b, TB))
```
wsmoses: Test forward and test reverse really should be avoided here as they're quite expensive. Instead can you just do the corresponding direct call to autodiff?

kshyatt: Yes!

test/rules/internal_rules/bigfloat.jl
```diff
@@ -5,29 +5,25 @@ using Test
 
 @testset "BigFloat arithmetic" begin
     a = rand(BigFloat)
+    da = rand(BigFloat)
     b = rand(BigFloat)
-    bf64 = rand(Float64) # for testing mixed methods
+    db = rand(BigFloat)
```
wsmoses: Can we just use non random numbers here and check against the equivalent constant?

kshyatt: Sure

kshyatt:
I think the CUDA stuff is unrelated, the windows fail might be pertinent though

test/runtests.jl
```diff
@@ -1,4 +1,4 @@
-import Enzyme
+#=import Enzyme
```
wsmoses: @kshyatt can you restore the normal testing setup, then I think this is good to go?

kshyatt: oui

wsmoses (approved):

PR #2935, opened 2026-01-27
Add integration test for MatrixAlgebraKit

kshyatt:
We have a bunch of Enzyme rules over there that absolutely love to break and generate weird errors

### DIFF
--- a/test/integration/MatrixAlgebraKit/Project.toml
+++ b/test/integration/MatrixAlgebraKit/Project.toml
@@ -1,0 +1,14 @@
+[deps]
+Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
+EnzymeCore = "f151be2c-9106-41f4-ab19-57ee4f262869"
+FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
+MatrixAlgebraKit = "6c742aac-3347-4629-af66-fc926824e5e4"
+StableRNGs = "860ef19b-820b-49d6-a774-d7a799459cd3"
+
+[sources]
+Enzyme = {path = "../../.."}
+EnzymeCore = {path = "../../../lib/EnzymeCore"}
+
+[compat]
+FiniteDifferences = "0.12.32"
+StableRNGs = "1.0.2"
--- a/test/integration/MatrixAlgebraKit/Project.toml
+++ b/test/integration/MatrixAlgebraKit/Project.toml
@@ -1,6 +1,7 @@
 [deps]
 Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
 EnzymeCore = "f151be2c-9106-41f4-ab19-57ee4f262869"
+EnzymeTestUtils = "12d8515a-0907-448a-8884-5fe00fdf1c5a"
 FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
 MatrixAlgebraKit = "6c742aac-3347-4629-af66-fc926824e5e4"
 StableRNGs = "860ef19b-820b-49d6-a774-d7a799459cd3"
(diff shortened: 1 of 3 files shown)


giordano:
Missed to push some files?

kshyatt:
Yep working on it now

wsmoses:
its still missing?

kshyatt:
Sorry yes will push soon! Got distracted with another thing I'm working on 😓

wsmoses:
bump @kshyatt

kshyatt:
We just had all our tests at MAK break (hmm...) so I'll try this again. Is there a way we could somehow use the existing `ParallelTestRunner` setup for MAK in these integration tests (e.g. `dev` MAK and just run its Enzyme tests)?

kshyatt:
@wsmoses Look ok now?

test/integration/MatrixAlgebraKit/Project.toml
```diff
@@ -0,0 +1,14 @@
...
+FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
+MatrixAlgebraKit = "6c742aac-3347-4629-af66-fc926824e5e4"
+StableRNGs = "860ef19b-820b-49d6-a774-d7a799459cd3"
+
+[sources]
+Enzyme = {path = "../../.."}
+EnzymeCore = {path = "../../../lib/EnzymeCore"}
+
+[compat]
+FiniteDifferences = "0.12.32"
```
wsmoses, line 14: you should pin the exact version of matrix algebra kit

test/integration/MatrixAlgebraKit/runtests.jl
```diff
@@ -0,0 +1,4 @@
+using Test, Pkg
+
+Pkg.develop("MatrixAlgebraKit")
```
wsmoses: this is possible, but really not desirable here since it isn't in any way actinoable from our side to debug/test if it fails -- what would be better is having some code which ideally directly calls enzyme

kshyatt: The reason I didn't do that right out of the gate is we have a LOT of code which directly calls enzyme and which various parts of have failed in the past 😅. But I'll see if I can whittle something down to start

wsmoses: yeah, imo here the reason is a maintainability/actionability thing -- such that if the test fails, can an enzyme person who has no idea what your code is, figure it out. Because if they can't and we let the test fail [and merge], the integration test isn't as useful

kshyatt: Sounds good! I'll try to bring over stuff that doesn't require all our setup logic for the input derivatives

wsmoses:
```
svd: Error During Test at /__w/Enzyme.jl/Enzyme.jl/test/integration/MatrixAlgebraKit/runtests.jl:66
  UndefVarError: `U` not defined
  Stacktrace:
   [1] macro expansion
     @ /__w/Enzyme.jl/Enzyme.jl/test/integration/MatrixAlgebraKit/runtests.jl:73 [inlined]
   [2] macro expansion
     @ /__w/_tool/julia/1.10.11/x64/share/julia/stdlib/v1.10/Test/src/Test.jl:1674 [inlined]
   [3] top-level scope
     @ /__w/Enzyme.jl/Enzyme.jl/test/integration/MatrixAlgebraKit/runtests.jl:1697
Test Summary: | Error  Total  Time
svd           |     1      1  0.9s
ERROR: LoadError: Some tests did not pass: 0 passed, 0 failed, 1 errored, 0 broken.
in expression starting at /__w/Enzyme.jl/Enzyme.jl/test/integration/MatrixAlgebraKit/runtests.jl:66
```

kshyatt:
@wsmoses I think this is finally ready

PR #2968, opened 2026-02-08
Add basic Flux integration tests using Enzyme

gamila-wisam:
### Summary

This PR adds a small set of integration tests for Flux models using Enzyme.jl, comparing
Enzyme gradients against Zygote gradients.

### Details

- Includes 3 simple models for initial testing:
  1. Dense layer
  2. Small Chain
  3. Small Conv layer
- Uses `check_approx` to compare Enzyme vs Zygote gradients.
- Flux trainable parameters are collected with `Flux.trainable` (replacing deprecated `Flux.params`).
- Designed to be lightweight for fast iteration; additional models can be added later.

### Testing

- The tests run successfully locally with `include("test/integration/Flux/runtests.jl")`.
- Full CI will run the standard Enzyme test suite, including these new tests.

### Motivation

Adds coverage for Flux models, ensuring Enzyme works correctly with common Flux layers.

### Related issue

References FluxML/Flux.jl#2644

### DIFF
--- a/test/integration/Flux/runtests.jl
+++ b/test/integration/Flux/runtests.jl
@@ -1,0 +1,63 @@
+using Enzyme
+using Flux
+using Zygote
+using Test
+using NNlib
+using StableRNGs
+using Random
+
+# generic loss function for any Flux model
+generic_loss_function(model, x, ps, st) = sum(abs2, first(model(x, ps, st)))
+
+# compute gradients using Enzyme
+function compute_enzyme_gradient(model, x, ps, st)
+    return Enzyme.gradient(
+        Enzyme.set_runtime_activity(Reverse),  
+        generic_loss_function,
+        Const(model),  
+        x,
+        ps,
+        Const(st),     
+    )[2:3]  
+end
+
+# compute gradients using Zygote
+function compute_zygote_gradient(model, x, ps, st)
+    _, dx, dps, _ = Zygote.gradient(generic_loss_function, model, x, ps, st)
+    return dx, dps
+end
+
+# compare Enzyme gradients with Zygote gradients
+function test_enzyme_gradients(model, x, ps, st)
+    dx, dps = compute_enzyme_gradient(model, x, ps, st)
+    dx_zygote, dps_zygote = compute_zygote_gradient(model, x, ps, st)
+
+    @test check_approx(dx, dx_zygote; atol = 1.0f-3, rtol = 1.0f-3)
+    @test check_approx(dps, dps_zygote; atol = 1.0f-3, rtol = 1.0f-3)
+end
+
+# small list of models to test
+const MODELS_LIST = [
+    # simple Dense layer
+    (Dense(2, 3), randn(Float32, 2, 4)),
+
+    # small Chain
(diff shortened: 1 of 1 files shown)


gamila-wisam:
The CI failure seems to be in SciML tests and is unrelated to these Flux integration tests. The new tests run successfully locally using :
include("test/integration/Flux/runtests.jl")

test/integration/Flux/runtests.jl
```diff
@@ -0,0 +1,63 @@
...
+    @test check_approx(dps, dps_zygote; atol = 1.0f-3, rtol = 1.0f-3)
+end
+
+# small list of models to test
+const MODELS_LIST = [
+    # simple Dense layer
+    (Dense(2, 3), randn(Float32, 2, 4)),
+
+    # small Chain
+    (Chain(Dense(2, 4, relu), Dense(4, 2)), randn(Float32, 2, 3)),
```
wsmoses, line 45: cc @CarloLucibello were there more models you wanted to test here, I know the flux ones have a bigger list iirc

test/integration/Flux/runtests.jl
```diff
@@ -0,0 +1,63 @@
...
+# compute gradients using Zygote
+function compute_zygote_gradient(model, x, ps, st)
+    _, dx, dps, _ = Zygote.gradient(generic_loss_function, model, x, ps, st)
+    return dx, dps
+end
+
+# compare Enzyme gradients with Zygote gradients
+function test_enzyme_gradients(model, x, ps, st)
+    dx, dps = compute_enzyme_gradient(model, x, ps, st)
+    dx_zygote, dps_zygote = compute_zygote_gradient(model, x, ps, st)
```
wsmoses, line 33: @gamila-wisam with zygote broken on 1.12, can you have this test against something other than zygote [otherwise we can't compare on 1.12+]

gamila-wisam: Sure, what about finite-differences gradients?

wsmoses: sure [as long as the models aren't so large that the time would be reasonable]

gamila-wisam: Yes I've considered that, I will try to ensure that runtime stays reasonable

test/integration/Flux/runtests.jl
```diff
@@ -0,0 +1,63 @@
...
+
+@testset "Enzyme Flux Integration" begin
+    for (i, (model, x)) in enumerate(MODELS_LIST)
+        @testset "[$i] $(nameof(typeof(model)))" begin
+            # set up parameters and state
+            ps = Flux.trainable(model)
+            st = nothing
+
+            # run the gradient test
+            test_enzyme_gradients(model, x, ps, st)
```
CarloLucibello, line 60: This approach won't work.
Flux is not Lux where the parameters are kept separate from the model. 
The gradient should be taken with respect to the model itself.

wsmoses:
@CarloLucibello can you take this over so flux is tested?

That way flux won't be broken by anything enzyme side (otherwise its untested and not assured)

PR #2972, opened 2026-02-12
Add inactive rules for timing, Printf, GC, and unicode case functions

michel2323:
Mark `Base.time_ns`, `Base.time`, `Printf.format`, and `Base.GC.enable` as inactive since they do not affect derivatives. Add `utf8proc_isupper` and `utf8proc_islower` to the nofree and inactive LLVM function sets, matching the existing treatment of `utf8proc_toupper`.


src/internal_rules/inactive.jl
```diff
@@ -107,6 +107,19 @@ function EnzymeRules.inactive(::typeof(Base.startswith), ::AbstractString, args.
 end
 
 Enzyme.EnzymeRules.inactive_noinl(::typeof(Core._compute_sparams), args...) = nothing
+function EnzymeRules.inactive(::typeof(Base.time_ns), args...)
+    return nothing
+end
+function EnzymeRules.inactive(::typeof(Base.time), args...)
```
wsmoses: Does inactive_noinl suffice ?

michel2323: I don't know. I saw that you use `inactive` for I/O functions. I wasn't sure whether these are. Is it ok like this? Only Printf is now `inactive`.

wsmoses: I'd do all as inactive_noinl unless we show it's necessary otherwise

michel2323: Done

PR #2979, opened 2026-02-22
Run runic formatter

abhro:
This PR happens because I was working on the docs and the editor kept complaining about whitespace. Since the pre-commit config makes runic a hook, I figured I'd just run it once to have the changes all done so that future runic changes are smaller and scoped to particular commits/PRs.

Also created the .git-blame-ignore-revs file so that we can ignore this commit in the context of git blame. If/when the PR gets merged, depending on the method of merging, the commit hash may need to be updated in the .git-blame-ignore-revs file to the actual hash in the history of 'main'.

Marking this PR as draft until CI passes because this was a lott of changes, and I didn't review all of it since it was made by a program and so _shouldn't_ break anything.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Run runic formatter
+a6237494bd9a8b40c44bc674bab581c66434b21e


vchuravy:
We haven't done this before since there are quite a few open PRs that all would have merge conflicts.  My hope is that we would get a continuous adoption of runic when changes are being made.

abhro:
Ah, that makes sense. I'll be closing this PR then. Thank you!

PR #3011, opened 2026-03-22
Fixup new callingconv to be newpm

wsmoses:


oscardssmith:
Have we turned on newpm for 1.10+ yet? That might be helpful in making sure newpm is reliable.

wsmoses:
yeah that's the default, but regardless this is just transitioning a manual call on an llvm fn to be a pass

PR #3024, opened 2026-03-27
Add ForwardModeSplit support

vchuravy:
Implements DEM_ForwardModeSplit (value 4 in CApi.h) which splits forward
mode AD into two separate LLVM functions — an augmented forward pass
(runs primal, stores tape) and a forward derivative pass (takes tape,
returns shadow) — mirroring ReverseModeSplit for the reverse case.

Changes:
- EnzymeCore: add ForwardModeSplit mode type, ForwardSplitNoPrimal /
  ForwardSplitWithPrimal constants, and ForwardSplitWidth/Modified helpers
- api.jl: DEM_ForwardModeSplit = 4, extend EnzymeCreateForwardDiff with
  aug parameter for split mode
- compiler.jl: ForwardModeSplitThunk struct; thunkbase branch calling
  EnzymeCreateAugmentedPrimal then EnzymeCreateForwardDiff(split);
  ABI wrapper and enzyme_call flags; fix AugmentedForwardThunk rettype to
  Const{actualRetType} so the primal-only augmented forward wrapper does
  not attempt to extract a shadow
- interpreter.jl, reflection.jl: treat DEM_ForwardModeSplit like
  DEM_ForwardMode (forward rules, Forward dispatch)
- Enzyme.jl: autodiff_thunk(::ForwardModeSplit, ...) returning
  (AugmentedForwardThunk, ForwardModeSplitThunk)
- errors.jl, customrules.jl, parallelrules.jl, activity.jl: extend all
  DEM_ForwardMode-only checks to also cover DEM_ForwardModeSplit
- test/forwardmodesplit.jl: new test file (42 tests)
- test/tests.jl: internal thunk tests for DEM_ForwardModeSplit

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/api.jl
+++ b/src/api.jl
@@ -228,7 +228,8 @@ end
     DEM_ForwardMode = 0,
     DEM_ReverseModePrimal = 1,
     DEM_ReverseModeGradient = 2,
-    DEM_ReverseModeCombined = 3
+    DEM_ReverseModeCombined = 3,
+    DEM_ForwardModeSplit = 4
 )
 
 # Create the derivative function itself.
@@ -335,10 +336,10 @@ function EnzymeCreateForwardDiff(
     additionalArg,
     typeInfo,
     uncacheable_args,
+    aug = C_NULL,
 )
     freeMemory = true
-    aug = C_NULL
-    subsequent_calls_may_write = false
+    subsequent_calls_may_write = mode == DEM_ForwardModeSplit
     ccall(
         (:EnzymeCreateForwardDiff, libEnzyme),
         LLVMValueRef,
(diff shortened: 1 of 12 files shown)


src/api.jl
```diff
@@ -335,10 +336,10 @@ function EnzymeCreateForwardDiff(
     additionalArg,
     typeInfo,
     uncacheable_args,
+    aug = C_NULL,
```
wsmoses, line 339: is there a jll change for this?

vchuravy: Why would there be? The C-API has already accepted this parameter and we are just exposing it to the user here?

PR #3044, opened 2026-04-23
Fix forward zero-set warning for non-float types

### LINKED ISSUE #1790: Fix "TODO forward zero-set of arraycopy used memset rather than runtime type "
https://github.com/EnzymeAD/Enzyme.jl/blob/8dde0343d7b4fdc92ecf4fe4fafd7e7df7cf1427/src/rules/llvmrules.jl#L335-L336

https://github.com/EnzymeAD/Enzyme.jl/blob/8dde0343d7b4fdc92ecf4fe4fafd7e7df7cf1427/src/rules/llvmrules.jl#L358-L359

I just got log-spammed with this:

```
┌ Warning: TODO forward zero-set of arraycopy used memset rather than runtime type 
│ Caused by:
│ Stacktrace:
│  [1] copy
│    @ ./array.jl:411
│  [2] unaliascopy
│    @ ./abstractarray.jl:1497
│  [3] unalias
│    @ ./abstractarray.jl:1481
│  [4] copyto!
│    @ ./abstractarray.jl:1067
│  [5] set!
│    @ ~/src/Chmy/src/Fields/field.jl:98
│  [6] set!
│    @ ~/src/Chmy/src/Fields/field.jl:0
└ @ Enzyme.Compiler ~/.julia/packages/GPUCompiler/y4cj1/src/utils.jl:59
```

vchuravy:
The Enzyme compiler previously emitted a warning during forward mode AD when doing an arraycopy or genericmemory_copy_slice on an inactive value that had an active return (necessitating zero-initialization of the shadow memory). When the precise runtime type of the underlying array couldn't be definitively determined at compile time (or if it was not a basic Float array), the compiler was falling back to emitting a raw memset to zero out the memory.

This patch introduces post_arraycopy_makezero and post_genericmemcpy_makezero which dynamically use Enzyme.make_zero! to zero initialize the arrays safely if they are not plain floats.

Co-authored-by: Antigravity <antigravity@google.com>

fixes #1790

### DIFF
--- a/src/rules/llvmrules.jl
+++ b/src/rules/llvmrules.jl
@@ -467,6 +467,11 @@ end
     return nothing
 end
 
+function post_arraycopy_makezero(B, callv, _, _)
+    emit_apply_generic!(B, LLVM.Value[unsafe_to_llvm(B, Enzyme.make_zero!), callv])
+    return nothing
+end
+
 function post_arraycopy_memset(B, callv, _, _)
     i8 = LLVM.IntType(8)
     algn = 0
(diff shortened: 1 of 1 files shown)


src/rules/llvmrules.jl
```diff
@@ -766,6 +764,11 @@ end
     return nothing
 end
 
+function post_genericmemcpy_makezero(B, callv, _, _)
+    emit_apply_generic!(B, LLVM.Value[unsafe_to_llvm(B, Enzyme.make_zero!), callv])
```
wsmoses: just make these the same fn? post_genericmemcpy_makezero and post_arraycopy_makezero

Issue #3041, opened 2026-04-22
Segmentation fault during GC


vchuravy:
Fairly small reproducer at https://gist.github.com/vchuravy/b7ab95cf515168b2655be760117b6cab

It should be able to reduce further since HagerZhang just calls `ϕdϕ` a bunch. 

```
Julia Version 1.11.9
Commit 53a02c0720c (2026-02-06 00:27 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 32 × AMD Ryzen 9 5950X 16-Core Processor
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, znver3)
Threads: 1 default, 0 interactive, 1 GC (on 32 virtual cores)
```

vchuravy:
This might have been an actual Julia bug. https://github.com/JuliaLang/julia/pull/61639 (or at least I found that one while looking into this one)

vchuravy:
Sadly reproduces even with the patch, but it seems to be a missing store to an `Array.mem` field, so potentially something that got fixed by https://github.com/JuliaLang/julia/pull/60924

vchuravy:
@wsmoses turning the attributor off turns the reproducer from a 90% likelihood to a 0/316 runs.

My current root cause is that we allocate an Array (likely the shadow) and we do not initalize the `Array.mem` to zero. 
GC then finds the Array, and is unhappy about the uninitialized memory. 

I am trying to find the transformation that causes this, but so far I had no luck, but it might be a combination of the allocator kind flags and attributor.

vchuravy:
Here is a Git repo that contains the history of all changes using `--print-changed` https://gist.github.com/vchuravy/aff6a80865e63d73079a0e6bc09100fb for all functions starting with `fwddiffe`

PR #3047, opened 2026-04-28
Turn RunAttributor into a compile-time preference

vchuravy:
Attributor yet again seems to be the cause of some illegal code transformation,
and I would like my last few days of investigating https://github.com/EnzymeAD/Enzyme.jl/issues/3041#issuecomment-4333046367 back.

This PR turns running it into an preference flag (so that folks who want it don't need to remember passing an environment flag),
but turns it off by default.

### DIFF
--- a/lib/EnzymeTestUtils/test/test_forward.jl
+++ b/lib/EnzymeTestUtils/test/test_forward.jl
@@ -116,7 +116,7 @@ end
                 atol = rtol = sqrt(eps(real(T)))
 
                 # Test that require the attributor on 1.11
-                if v"1.11.0" <= VERSION < v"1.12.0" && Enzyme.run_attributor()
+                if v"1.11.0" <= VERSION < v"1.12.0" && !Enzyme.run_attributor()
                     if Tx == Const &&(T == ComplexF32 || T == ComplexF64)
                         Ta == Duplicated || Ta == BatchDuplicated || continue
                     end
--- a/lib/EnzymeTestUtils/test/test_forward.jl
+++ b/lib/EnzymeTestUtils/test/test_forward.jl
@@ -115,9 +115,16 @@ end
                 a = randn(T)
                 atol = rtol = sqrt(eps(real(T)))
 
-	        @test !fails() do
-	        	test_forward(f_multiarg, Tret, (x, Tx), (a, Ta); atol, rtol)
-	        end
+                # Test that require the attributor on 1.11
+                if v"1.11.0" <= VERSION < v"1.12.0" && Enzyme.run_attributor()
+                    if Tx == Const &&(T == ComplexF32 || T == ComplexF64)
+                        Ta == Duplicated || Ta == BatchDuplicated || continue
+                    end
+                end
+
+                @test !fails() do
+                    test_forward(f_multiarg, Tret, (x, Tx), (a, Ta); atol, rtol)
+                end
             end
         end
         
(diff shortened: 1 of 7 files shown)


vchuravy:
*Sigh* it seems necessary for 1.10 test to pass.

src/preferences.jl
```diff
@@ -0,0 +1,25 @@
+using Preferences
+
+"""
+    run_attributor() -> Bool
+
+Return whether the Enzyme attributor pass is enabled. This is a compile-time
+preference that requires restarting Julia to take effect after being changed.
+
+The attributor pass can improve differentiation quality but may cause issues on
+Julia 1.12 and later. Defaults to `false`.
```
vchuravy:
```suggestion
!!! note
    The attributor pass can improve differentiation quality but may cause issues on
    Julia 1.11 and later. Defaults to `true` on Julia 1.10 and `false` on 1.11 and later.
```

src/preferences.jl
```diff
@@ -0,0 +1,25 @@
...
+
+See also [`set_run_attributor!`](@ref).
+"""
+run_attributor() = @load_preference("run_attributor", false)
```
vchuravy:
```suggestion
run_attributor() = @load_preference("run_attributor", VERSION < v"1.11")
```

lib/EnzymeTestUtils/test/test_forward.jl
```diff
@@ -115,9 +115,16 @@ end
                 a = randn(T)
                 atol = rtol = sqrt(eps(real(T)))
 
-	        @test !fails() do
-	        	test_forward(f_multiarg, Tret, (x, Tx), (a, Ta); atol, rtol)
-	        end
+                # Test that require the attributor on 1.11
+                if v"1.11.0" <= VERSION < v"1.12.0" && Enzyme.run_attributor()
```
vchuravy:
```suggestion
                if v"1.11.0" <= VERSION < v"1.12.0" && !Enzyme.run_attributor()
```

wsmoses:
is there a reason for doing this since we already have the ref option?

I agree this is nicer, but technically breaking

vchuravy:
> is there a reason for doing this since we already have the ref option?

Attributor is turned off already by default for 1.12 and as #3041 shows causes compilation issues that are really frustrating to debug in 1.11 (I spent ~4days trying to root cause that bug this week).

So yes as CI shows it sometimes does positive things, but it also continuously seems to create deeply frustrating bugs and I for one had enough of it.

wsmoses:
oh I meant more in reference to it being a preference vs the current ref [not wrt the default or not].

we should look at the remaining failure with 1.11 when disabled. part of why its okay to turn off is ive been slowly making an "attributor lite" pass for our specific needs, though not as aggressive

vchuravy:
> oh I meant more in reference to it being a preference vs the current ref [not wrt the default or not].

I think that's fine. We are promoting it from an internal interface, to something more official and documented.

src/compiler/optimize.jl
```diff
@@ -11,7 +11,8 @@ LLVM.@module_pass "enzyme-fixup-julia" FixupJuliaCallingConventionPass
 LLVM.@module_pass "enzyme-fixup-julia-sret" FixupJuliaCallingConventionSRetPass
 LLVM.@module_pass "enzyme-fixup-batched-julia" FixupBatchedJuliaCallingConventionPass
 
-const RunAttributor = Ref(VERSION < v"1.12")
+import ..Enzyme: run_attributor
+const RunAttributor = run_attributor()
```
wsmoses, line 15: I really worry about making this change as there is a lot of code which modifies the ref atm downstream

wsmoses: I also don't like the idea about this being a once-per-session setting

vchuravy: > I also don't like the idea about this being a once-per-session setting

But it already is, due to caching if you set it inside the session you will get inconclusive results.

With this at least you can setup two projects and change the setting there.

PR #3048, opened 2026-04-28
Manually work around unrecognized processor

### LINKED ISSUE #2908: v1.12: 'znver5' is not a recognized processor for this target (ignoring processor)
```
julia> using Enzyme
'znver5' is not a recognized processor for this target (ignoring processor)
'znver5' is not a recognized processor for this target (ignoring processor)
```

The fact that this warning is getting printed multiple times (and isn't actionable or relevant) is fairly suboptimal.

wsmoses:
fixes https://github.com/EnzymeAD/Enzyme.jl/issues/2908

### DIFF
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -51,19 +51,31 @@ using Preferences
 bitcode_replacement() = parse(Bool, @load_preference("bitcode_replacement", "true"))
 bitcode_replacement!(val) = @set_preferences!("bitcode_replacement" => string(val))
 
-function cpu_name()
-    name = ccall(:jl_get_cpu_name, String, ())
+function llvm_compatible_cpu_name(name::String)
     if Base.libllvm_version < v"19"
         if Sys.isapple() && Sys.ARCH === :aarch64
-            if name == "apple-m4"
-                return "apple-m3"
+            m = match(r"^apple-m(\d+)$", name)
+            if m !== nothing
+                v = parse(Int, m.captures[1])
+                if v >= 4
+                    return "apple-m3"
+                end
             end
         end
-        if name == "znver5"
-            return "znver4"
+        m = match(r"^znver(\d+)$", name)
+        if m !== nothing
+            v = parse(Int, m.captures[1])
+            if v >= 5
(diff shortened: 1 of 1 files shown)


oscardssmith (approved):
Horribly hacky but sure.

wsmoses:
yeah I mean is there a better fix?

wsmoses:
and why does the julia cpu runtime fn spit out a cpu which cannot be used?

src/compiler.jl
```diff
@@ -52,7 +52,18 @@ bitcode_replacement() = parse(Bool, @load_preference("bitcode_replacement", "tru
...
+            if name == "apple-m4"
+                return "apple-m3"
+            end
+        end
+        if name == "znver5"
+            return "znver4"
+        end
+    end
+    return name
 end
```
vchuravy: When @gbaraldi is back from vacation he will say something about https://github.com/JuliaLang/cpufeatures

But until then, can we move it to a seperate function "llvm_compatible_cpu_name"?

src/compiler.jl
```diff
@@ -52,7 +52,18 @@ bitcode_replacement() = parse(Bool, @load_preference("bitcode_replacement", "tru
 bitcode_replacement!(val) = @set_preferences!("bitcode_replacement" => string(val))
 
 function cpu_name()
-    ccall(:jl_get_cpu_name, String, ())
+    name = ccall(:jl_get_cpu_name, String, ())
+    if Base.libllvm_version < v"19"
+        if Sys.isapple() && Sys.ARCH === :aarch64
+            if name == "apple-m4"
```
vchuravy: Can we match the `-m4` string and check for greater or equal? So that m5 also get's detected as m3

And same for znver

oscardssmith: that only works if you assume that the cpu maker doesn't drop features (which is probably fair for apple and AMD since they have basic competence but is a bad idea for Intel)

vchuravy: We do detect features seperatly

vchuravy (approved):

Issue #3050, opened 2026-04-28
CUDA kernel support broken (presumably per CUDA 6)?


wsmoses:
@giordano @vchuravy @maleadt, presumably this happened with the CUDA 6 upgrade, since nothing else relevant merged here

see CI current: https://buildkite.com/julialang/enzyme-dot-jl/builds/6881/canvas?jid=019dd560-be8f-48b5-b897-388059688604&tab=output

```

warning: Linking two modules of different target triples: 'start' is 'x86_64-linux-gnu' whereas 'start' is 'nvptx64-nvidia-cuda'
--
 
warning: linking module flags 'Dwarf Version': IDs have conflicting values ('i32 4' from start with 'i32 2' from start)
mul_kernel: Error During Test at /var/lib/buildkite-agent/builds/gpuci-16/julialang/enzyme-dot-jl/test/cuda.jl:18
Got exception outside of a @test
LLVM error: Called function is not the same type as the call!
call void bitcast (void ({ i64, i32 }, { i8 addrspace(1)*, i64, [1 x i64], i64 }*)* @julia_mul_kernel_34046 to void ({ i64, i32 }, { i8 addrspace(1)*, i64, [1 x i64], i64 }, { i8 addrspace(1)*, i64, [1 x i64], i64 })*)({ i64, i32 } %state, { i8 addrspace(1)*, i64, [1 x i64], i64 } %0, { i8 addrspace(1)*, i64, [1 x i64], i64 } %1), !dbg !240
 
Stacktrace:
[1] verify(mod::LLVM.Module)
@ LLVM ~/.cache/julia-buildkite-plugin/depots/260b8bb4-e043-429b-b2f1-bbf9045c057a/packages/LLVM/hhpnm/src/analysis.jl:19
[2] macro expansion
@ ~/.cache/julia-buildkite-plugin/depots/260b8bb4-e043-429b-b2f1-bbf9045c057a/packages/GPUCompiler/Yuvf5/src/driver.jl:424 [inlined]
[3] emit_llvm(job::GPUCompiler.CompilerJob; kwargs::@Kwargs{})
@ GPUCompiler ~/.cache/julia-buildkite-plugin/depots/260b8bb4-e043-429b-b2f1-bbf9045c057a/packages/GPUCompiler/Yuvf5/src/utils.jl:116
[4] emit_llvm(job::GPUCompiler.CompilerJob)
[...]

PR #3052, opened 2026-04-28
Use CUDA@v5 in Buildkite tests

giordano:
Ref https://github.com/EnzymeAD/Enzyme.jl/issues/3050#issuecomment-4339204706


giordano:
I don't understand why CI isn't running at all here, don't know if it's GitHub constantly being on fire or what.  I'll try to open from a branch of this repo, instead of a fork, and see if that's of any help.

vchuravy:
That's a buildkite thing I believe it only runs on PRs from the same repo

PR #3053, opened 2026-04-28
Option to ignore const checks in EnzymeTestUtils

jlperla:
## Summary
  Adds an opt-in `ignore_const_checks::Bool=false` keyword to `EnzymeTestUtils.test_forward` and `test_reverse`. When `true`, the post-call  mutation assertion is skipped for `Const`-annotated arguments (and for  `test_reverse`, the FD-vs-AD derivative comparison is skipped on `Const` args  too).

## Motivation
Custom rules sometimes legitimately scribble on a `Const` scratch buffer
during AD without that mutation affecting the gradient — e.g. workspace
structs holding LAPACK workplace arrays we do not want/need to shadow. Today the only workaround is to
monkey-patch the helpers or hand-write the finite-differences checks.  This gives users a 1-line opt-in instead.

I am not sure if these are the right unit tests to have for this as they are AI generated.

@wsmoses  Please let me know if this is the wrong way to submit suggestions.

### DIFF
--- a/lib/EnzymeTestUtils/src/test_forward.jl
+++ b/lib/EnzymeTestUtils/src/test_forward.jl
@@ -23,6 +23,9 @@ constraints:
 - `rtol`: Relative tolerance for `isapprox`.
 - `atol`: Absolute tolerance for `isapprox`.
 - `testset_name`: Name to use for a testset in which all tests are evaluated.
+- `ignore_const_checks::Bool=false`: If `true`, skip the post-call assertion that
+    `Const` arguments were mutated identically by the rule and the primal function.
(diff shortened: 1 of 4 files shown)


wsmoses:
I see what you're doing here, but I suppose it may be better to have a way to mark some variables as scratch more generally than just all const ones?

jlperla:
Yes, that would be far superior.  I tried that but didn't go that direction because a new Activity type (or some test-only flag on activity?) seemed too much for just the unit tests.  But if you think that something along those lines is the right direction, then I can close this PR and make an issue on the topic.  My pirated test_forward and test_reverse solve my problems for now.

PR #3062, opened 2026-04-30
Fix type unstable parallel reverse for

wsmoses:
requires https://github.com/EnzymeAD/Enzyme/pull/2799

### DIFF
--- a/src/typeutils/recursive_add.jl
+++ b/src/typeutils/recursive_add.jl
@@ -52,7 +52,7 @@ end
 
 @inline function atomicrmw_add!(ptr::Ptr{Float64}, val::Float64)
     Base.llvmcall(
-        "atomicrmw fadd ptr %0, double %1 monotonic\nret void",
+        "atomicrmw fadd double* %0, double %1 monotonic\nret void",
         Cvoid,
         Tuple{Ptr{Float64}, Float64},
         ptr,
@@ -62,7 +62,7 @@ end
 
 @inline function atomicrmw_add!(ptr::Ptr{Float32}, val::Float32)
     Base.llvmcall(
-        "atomicrmw fadd ptr %0, float %1 monotonic\nret void",
+        "atomicrmw fadd float* %0, float %1 monotonic\nret void",
         Cvoid,
         Tuple{Ptr{Float32}, Float32},
         ptr,
--- a/src/typeutils/recursive_add.jl
+++ b/src/typeutils/recursive_add.jl
@@ -50,37 +50,171 @@ end
 @inline mutable_register(::Type{T}) where {T<:Array} = true
 @inline mutable_register(::Type{T}) where {T} = ismutabletype(T)
 
-# Recursively In-place accumulate(aka +=). E.g. generalization of x .+= f(y)
-@inline function recursive_accumulate(x::Array{T}, y::Array{T}, f::F = identity) where {T,F}
-    if !mutable_register(T)
-        for I in eachindex(x)
-            prev = x[I]
-            @inbounds x[I] = recursive_add(x[I], (@inbounds y[I]), f, mutable_register)
+@inline function atomicrmw_add!(ptr::Ptr{Float64}, val::Float64)
+    Base.llvmcall(
+        "atomicrmw fadd ptr %0, double %1 monotonic\nret void",
+        Cvoid,
+        Tuple{Ptr{Float64}, Float64},
+        ptr,
+        val
+    )
+end
+
+@inline function atomicrmw_add!(ptr::Ptr{Float32}, val::Float32)
+    Base.llvmcall(
+        "atomicrmw fadd ptr %0, float %1 monotonic\nret void",
+        Cvoid,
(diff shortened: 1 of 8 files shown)


src/typeutils/recursive_add.jl
```diff
@@ -50,37 +50,181 @@ end
...
-            ST = Core.Typeof(xi)
+    if !ismutabletype(T)
+        return :(nothing)
+    end
+    if atomic
+        return quote
+            GC.@preserve x y begin
+                px = Ptr{T}(pointer_from_objref(x))
+                py = Ptr{T}(pointer_from_objref(y))
+                atomic_accumulate!(px, py, f)
```
vchuravy, line 175: So this is weird the pointer you have here is no a `Ptr{T}`

vchuravy:
Can you add some test for different kind of structures that can be recursive accumulated? 

One of the things I am worried about is write barriers, but if I read the code correctly we should never try to use it for pointer fields?

wsmoses:
done, and yeah basically this is just that the "active" (aka floats) parts of the data are +='d atomically

test/recursive_accumulate.jl
```diff
@@ -0,0 +1,104 @@
...
+        @test x.b == 2
+        
+        # Test NestedMut
+        x = NestedMut(1.0, SimpleMut(2.0, 3))
+        y = NestedMut(4.0, SimpleMut(5.0, 6))
+        Enzyme.Compiler.recursive_accumulate(x, y)
+        @test x.a ≈ 5.0
+        @test x.inner.a ≈ 7.0
+        @test x.inner.b == 3
+
```
vchuravy:
```suggestion
        # Test NestedMut
        x = NestedMut(1.0, SimpleMut(2.0, 3))
        y = NestedMut(4.0, SimpleMut(5.0, 6))
        inner = x.inner
        Enzyme.Compiler.recursive_accumulate(x, y)
        @test x.a ≈ 5.0
        @test x.inner.a ≈ 7.0
        @test x.inner.b == 3
        # Check for pointer-identity
        @test x.inner === inner
```

test/recursive_accumulate.jl
```diff
@@ -0,0 +1,104 @@
+using Enzyme
+using Test
+
+mutable struct SimpleMut
+    a::Float64
+    b::Int
+end
```
vchuravy, line 7: What happens if any of these fields are `const` or marked `@atomic`?

test/recursive_accumulate.jl
```diff
@@ -0,0 +1,104 @@
...
+    end
+
+    @testset "atomic=true" begin
+        # Test SimpleMut
+        x = SimpleMut(1.0, 2)
+        y = SimpleMut(3.0, 4)
+        Enzyme.Compiler.recursive_accumulate(x, y, Val(true))
+        @test x.a ≈ 4.0
+        @test x.b == 2
+
```
vchuravy, line 82: Should there be a test for NestedMut as well?

wsmoses:
We regardless should += into it

test/recursive_accumulate.jl
```diff
@@ -0,0 +1,104 @@
...
+        @test x.a ≈ 5.0
+        @test x.inner.a ≈ 7.0
+        @test x.inner.b == 3
+        
+        # Test ArrayMut
+        x = ArrayMut(1.0, [2.0, 3.0])
+        y = ArrayMut(4.0, [5.0, 6.0])
+        Enzyme.Compiler.recursive_accumulate(x, y)
+        @test x.a ≈ 5.0
+        @test x.arr ≈ [7.0, 9.0]
```
vchuravy, line 44:
```suggestion
        # Test ArrayMut
        x = ArrayMut(1.0, [2.0, 3.0])
        y = ArrayMut(4.0, [5.0, 6.0])
        arr = x.arr
        Enzyme.Compiler.recursive_accumulate(x, y)
        @test x.a ≈ 5.0
        @test x.arr ≈ [7.0, 9.0]
        @test x.arr === arr # pointer identity
```

test/recursive_accumulate.jl
```diff
@@ -0,0 +1,108 @@
+using Enzyme
+using Test
+
+mutable struct SimpleMut
+    a::Float64
+    b::Int
+end
```
vchuravy, line 7: I think other cases would be a mixture of `immutable` and `mutable`?

wsmoses: I mean as is right now this function is only called for "Active" values so I'm not terribly concerned

vchuravy: Not sure I understand `mutable struct` like used here are `Duplicated`?

But I could have something like:

```
struct MyPoint
     x::Float64
    y::Float64
end

mutable struct SimpleMut
     p::MyPoint
     p_other::NTuple{3, Float64}
```

wsmoses: yeah I just mean the callers of this code [for races] are only active returns. So in this case it would be called for a ref of say an immutable version of SimpleMut

PR #3101, opened 2026-05-12
Fix ijl_lazy_load_and_lookup handler for Julia 1.13 symbol arguments

### LINKED ISSUE #3083: Julia 1.13: No derivative for `ijl_lazy_load_and_lookup` breaks BLAS differentiation
## Summary

Julia 1.13 changed how BLAS and other foreign function calls are lowered. Instead of a direct `call` to the library symbol, calls now go through `ijl_lazy_load_and_lookup`, a new lazy dynamic dispatch mechanism. Enzyme has no rule for this function and cannot differentiate through it.

## Error

```
EnzymeNoDerivativeError: No augmented forward pass found for ijl_lazy_load_and_lookup
  at context:  %N = call ptr @ijl_lazy_load_and_lookup(...)
```

## Affected tests

- `sugar` — `Jacobian` (via `BLAS.gemv!` → `dgemv_64_`)
- `tests` — `BLAS` (via `LinearAlgebra.dot` → `cblas_ddot64_`)
- `DiffTests` — `Vector to Number`, `Matrix to Number` (via `cblas_ddot64_`, `dgemm_64_`)
- `basic` — `Dot product` (via `cblas_sdot64_`, `cblas_ddot64_`)
- `advanced` — `Large dynamic tape` (via `sgemv_64_`)
- `ext/specialfunctions` — `SpecialFunctions` (via `erfc` and other symbols)

## Side effect

The `optimize` test (`Parameter removal`) expects the warning  
`"Using fallback BLAS replacements for"` when Enzyme cannot handle a BLAS call. With this change, Enzyme now fails with a hard error instead of falling back gracefully, so the warning is never emitted and that assertion test also fails.

## Notes

- From `test_1.13-dev.log`
- Enzyme's BLAS rules and/or the foreign-function handling need to be updated to account for the new `ijl_lazy_load_and_lookup` lowering in Julia 1.13.

vchuravy:
## Summary

Fixes #3083

In Julia 1.13, LAPACK symbol dispatch was changed to use `ijl_lazy_load_and_lookup` with two `ejl_inserted` GlobalVariables (`ptr addrspace(10)`) as arguments — a Julia library-reference object and a Julia Symbol — instead of the old `(library-path-string, C-string-fname)` pair.

### DIFF
--- a/src/compiler/validation.jl
+++ b/src/compiler/validation.jl
@@ -886,12 +886,12 @@ function check_ir!(interp, @nospecialize(job::CompilerJob), errors::Vector{IRErr
                 legal2, sym = absint(fname_llvm)
                 if legal2
                     sym = unbind(sym)
-                end
-                if isa(sym, GlobalRef) && isdefined(sym.mod, sym.name)
-                    sym = getfield(sym.mod, sym.name)
-                end
-                if isa(sym, Symbol)
-                    fname = String(sym)
+                    if isa(sym, GlobalRef) && isdefined(sym.mod, sym.name)
+                        sym = getfield(sym.mod, sym.name)
+                    end
+                    if isa(sym, Symbol)
+                        fname = String(sym)
+                    end
                 end
             end
 
--- a/src/compiler/validation.jl
+++ b/src/compiler/validation.jl
@@ -853,9 +853,7 @@ function check_ir!(interp, @nospecialize(job::CompilerJob), errors::Vector{IRErr
 
             ops = arg_operands_view(inst)
             @assert length(ops) == 2
-            flib_v = ops[1]
-
-            flib = flib_v
+            flib = ops[1]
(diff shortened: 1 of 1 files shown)


src/compiler/validation.jl
```diff
@@ -866,24 +864,46 @@ function check_ir!(interp, @nospecialize(job::CompilerJob), errors::Vector{IRErr
...
             end
 
-            if !isa(fname, String) || !isa(flib, String)
+            # Julia 1.13+: fname is an ejl_inserted GlobalVariable holding a Julia Symbol.
+            if !isa(fname, String) && isa(fname_llvm, LLVM.GlobalVariable)
+                legal2, sym = absint(fname_llvm)
+                if legal2
+                    sym = unbind(sym)
+                end
+                if isa(sym, GlobalRef) && isdefined(sym.mod, sym.name)
```
wsmoses: don't you need to still check legal here?

PR #3102, opened 2026-05-13
Adapt to LLVM.jl 9.8.1: pass explicit JITDylib to lookup

vchuravy:
## Summary
- `lookup(jljit, name)` is deprecated in LLVM.jl 9.8.0+ to adapt to JuliaLang/julia#60988
- Pass explicit `JITDylib` to both `lookup` call sites in `src/compiler/orcv2.jl`
- Bump LLVM compat bound to `9.8.1`

Cousin-PR to https://github.com/JuliaGPU/GPUCompiler.jl/pull/796


src/compiler/orcv2.jl
```diff
@@ -292,7 +292,8 @@ function add!(mod)
 end
 
 function lookup(name)
-    LLVM.lookup(jit[].jit, name)
+    lljit = jit[].jit
+    LLVM.lookup(lljit, JITDylib(lljit), name)
```
wsmoses, line 296: is the dylib something generated a new each time? this feels wrong if so then

wsmoses: @vchuravy without this PR doctests are red causing chaos. I'm going to merge for now to unblock CI.

Can you address this in post commit?

PR #3106, opened 2026-05-15
Add forward rules for dense linear solves

jlperla:
## Summary
- add forward-mode support for dense `A \ b` and `A \ B`
- reuse the existing `compute_lu_cache` path used by the reverse rule
- support active or constant RHS, active square `A`, and batched forward mode
- keep active rectangular solves unsupported for now

## Tests

### DIFF
--- a/src/internal_rules/linalg.jl
+++ b/src/internal_rules/linalg.jl
@@ -163,6 +163,62 @@ function EnzymeRules.reverse(
     return (nothing, nothing)
 end
 
+function EnzymeRules.forward(
+        config::EnzymeRules.FwdConfig,
+        func::Const{typeof(\)},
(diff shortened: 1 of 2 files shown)


jlperla:
@wsmoses   This one I felt more confident implementing.  Seems to pass tests/runic/etc. and I think it follows the patterns of others in that file.

Let me know if you want anything changed.

wsmoses (approved):

wsmoses:
same comment of fine for now, but we should tablegen likely here

jlperla:
Yes on tablegen.  I asked an AI to come up with tablegen but it ended up looking very complicated because of managing the pivots.  But it got the dense LU stuff working and passing tests.  Happy to post that up as a PR for you guys to look at and suggest changes/take it over.  But not sure it is worth your time.

PR #3107, opened 2026-05-15
Add forward rules for triangular linear solves

jlperla:
## Summary
- add forward-mode support for high-level triangular solves
- cover vector and matrix RHS cases, including unit triangular and transposed triangular wrappers
- reuse the existing `_zero_unused_elements!` handling for unit triangular structure
- support batched forward mode

## Tests

### DIFF
--- a/src/internal_rules/linalg.jl
+++ b/src/internal_rules/linalg.jl
@@ -163,6 +163,56 @@ function EnzymeRules.reverse(
     return (nothing, nothing)
 end
 
+function EnzymeRules.forward(
+        config::EnzymeRules.FwdConfig,
+        func::Const{typeof(\)},
+        RT::Type{<:Union{Const, DuplicatedNoNeed, Duplicated, BatchDuplicatedNoNeed, BatchDuplicated}},
+        A::Annotation{<:LinearAlgebra.AbstractTriangular},
+        B::Annotation{<:Array},
+    )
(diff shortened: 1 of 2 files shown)


jlperla:
@wsmoses  I added these in.  I think this is a better place to fix some of those issues than the tablegen for now (and it is symmetric to the reverses-rules, so not coming out of nowhere).

wsmoses (approved):

wsmoses:
This is fine for now, but I do think this may make sense to do as a tablegen rule. For regular matmul we saw substantial speed ups in doing so. Assuming this directly translates to the blas call [and not a combination of calls], I would expect similar here

jlperla:
Definitely.  If you want any more work on the https://github.com/EnzymeAD/Enzyme/pull/2825 just post comments or a code review and I will tell the AI to do its thing.  But while I don't mind using AIs for patching, I am loathe to do so when I don't understand the code.

wsmoses:
I think you still need to address the previous comments there, involving both splitting it up and also just changing the relefvant pieces?

jlperla:
@wsmoses  Sorry, I think a question might have been lost on slack.  To clarify, I wasn't sure if you wanted me to split https://github.com/EnzymeAD/Enzyme/pull/2825 into 3 separate PRs for `trsv, trsm, and potrs`?  Or something else?

wsmoses:
yes that's correct

PR #3108, opened 2026-05-16
Add reverse rule for Cholesky ldiv!

jlperla:
## Summary
- add reverse-mode support for `ldiv!(fact::Cholesky, B)`
- mirror the existing forward rule by reversing through the Cholesky triangular solves
- cache the Cholesky factor when needed for overwritten arguments
- add focused `test_reverse` coverage for vector, matrix, const, and batched paths

## Tests

### DIFF
--- a/src/internal_rules/linalg.jl
+++ b/src/internal_rules/linalg.jl
@@ -346,7 +346,8 @@ function EnzymeRules.reverse(
     U = cache_fact.U
     N = EnzymeRules.width(config)
 
-    for i in 1:N
+    ntuple(Val(N)) do i
+        Base.@_inline_meta
         dB = N == 1 ? B.dval : B.dval[i]
         zU = adjoint(U) \ dB
         zL = adjoint(L) \ zU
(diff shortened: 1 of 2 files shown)


src/internal_rules/linalg.jl
```diff
@@ -295,6 +295,77 @@ function EnzymeRules.forward(
...
+    if B isa Const
+        return (nothing, nothing)
+    end
+
+    cache_fact, cache_Lout, cache_Uout = cache
+    L = cache_fact.L
+    U = cache_fact.U
+    N = EnzymeRules.width(config)
+
+    for i in 1:N
```
wsmoses: you should make this an ntuple, not a regular for loop

jlperla: Done

wsmoses:
actually taking a step back, what fails without this?

wsmoses:
specifically what is the error (we have potrf derivatives iirc)

jlperla:
Yeah let me give you a more focused issue tomorrow or Tuesday. I was finding both failures in some cases and also some horrible performance.  Can't remember which and when.

Hold off on this until i can get the big picture to show you what is going on with errors, and/or massive performance differences

PR #3111, opened 2026-05-17
Fix race condition in parallel interpreter

### LINKED ISSUE #3103: possible race condition in `SigCache` during concurrent `autodiff` calls
transferred from https://github.com/JuliaLang/julia/issues/61790

codex reduced to pure Enzyme:

```julia
using Pkg

Pkg.activate(; temp=true)
Pkg.add("Enzyme")

using Enzyme

f(x) = sin(x) + x * x

function run_round(round)
    # Deterministic setup: force all worker tasks onto Enzyme's cold
    # rule-signature cache path. The operation under test is the public
    # Enzyme.autodiff call below.
    empty!(Enzyme.Compiler.Interpreter.SigCache)

    ntasks = Threads.nthreads()
    ready = Threads.Atomic{Int}(0)
    go = Threads.Atomic{Bool}(false)

    tasks = map(1:ntasks) do tid
        Threads.@spawn begin
            Threads.atomic_add!(ready, 1)
            while !go[]
                GC.safepoint()
            end

            x = 0.1 + tid + round / 10_000
            Enzyme.autodiff(
                Enzyme.set_abi(Enzyme.Reverse, Enzyme.NonGenABI),
                f,
                Enzyme.Active,
                Enzyme.Active(x),
            )
        end
    end

    while ready[] != ntasks
        GC.safepoint()
    end
    go[] = true

    fetch.(tasks)
    return nothing
end

println("VERSION = ", VERSION)
println("Threads.nthreads() = ", Threads.nthreads())
println("Threads.ngcthreads() = ", Threads.ngcthreads())
println("Enzyme = ", pkgversion(Enzyme))

rounds = parse(Int, get(ENV, "ROUNDS", "10000"))
for round in 1:rounds
    round % 100 == 0 && println("round ", round)
    run_round(round)
end

println("completed")
```

running with `julia +1.12 -t 2 --gcthreads=2 enzyme_mwe.jl ` gives me
```julia
...
nested task error: UndefRefError: access to undefined reference
    Stacktrace:
      [1] getindex
        @ ./essentials.jl:386 [inlined]
      [2] ht_keyindex2_shorthash!(h::Dict{Tuple, Dict{UInt64, IdSet{Type}}}, key::Tuple{typeof(EnzymeCore.EnzymeRules.augmented_primal), UnionAll})
...
```

wsmoses:
resolves https://github.com/EnzymeAD/Enzyme.jl/issues/3103

### DIFF
--- a/src/compiler/interpreter.jl
+++ b/src/compiler/interpreter.jl
@@ -85,26 +85,35 @@ struct EnzymeInterpreter{T} <: AbstractInterpreter
     handler::T
 end
 
+const SigCacheLock = ReentrantLock()
+const InterpreterLock = ReentrantLock()
 const SigCache = Dict{Tuple, Dict{UInt, Base.IdSet{Type}}}()
 function get_rule_signatures(f, TT, world)
-    subdict = if haskey(SigCache, (f, TT))
-       SigCache[(f, TT)]
-    else
-       tmp = Dict{UInt, Base.IdSet{Type}}()
-       SigCache[(f, TT)] = tmp
-       tmp
+    subdict = lock(SigCacheLock) do
+        if haskey(SigCache, (f, TT))
+           SigCache[(f, TT)]
+        else
+           tmp = Dict{UInt, Base.IdSet{Type}}()
+           SigCache[(f, TT)] = tmp
+           tmp
+        end
     end
-    if haskey(subdict, world)
-       return subdict[world]
+    
+    return lock(SigCacheLock) do
+        if haskey(subdict, world)
+           return subdict[world]
+        end
+        
+        fwdrules_meths = Base._methods(f, TT, -1, world)::Vector
+        sigs = Type[]
+        for rule in fwdrules_meths
+            push!(sigs, (rule::Core.MethodMatch).method.sig)
+        end
+        result = Base.IdSet{Type}(sigs)
+        
+        subdict[world] = result
+        return result
     end
-    fwdrules_meths = Base._methods(f, TT, -1, world)::Vector
-    sigs = Type[]
-    for rule in fwdrules_meths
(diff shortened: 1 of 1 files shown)


oscardssmith (approved):
seems reasonable.

wsmoses:
didn't fix the 1.12 windows ci problems, but at least an improvement -- can iterate on

vchuravy:
I mean the real fix would be to yank this out and put it and the other caches into the EnzymeContext.

src/compiler/interpreter.jl
```diff
@@ -146,33 +155,50 @@ function EnzymeInterpreter(
...
+        end
+
         cache_or_token = cache_or_token::CodeCache
         invalid = false
-        if forward_rules
-            fwdrules = get_rule_signatures(EnzymeRules.forward, Tuple{<:FwdConfig, <:Annotation, Type{<:Annotation}, Vararg{Annotation}}, world)
-            if !rule_sigs_equal(fwdrules, LastFwdWorld[])
-                LastFwdWorld[] = fwdrules
-                invalid = true
+        lock(InterpreterLock) do
```
vchuravy, line 178: Why do we need an interpreter lock? Or if we actually do why only here?

wsmoses:
```
                if !rule_sigs_equal(fwdrules, LastFwdWorld[])
                    LastFwdWorld[] = fwdrules
                    invalid = true
                end

...
            if invalid
                Base.empty!(cache_or_token)
            end
```

the same cache_or_token is racy across parallel interpreter constructor invocations

vchuravy: But that is the ephermal cache owned by the instance of the interpreter and shouldn't we then need locks all over the place?

PR #3112, opened 2026-05-18
Compiler.deferred_codegen: hide ccall behind @generated shell (#3091)

### LINKED ISSUE #3091: Enzyme cannot be built into a sysimage (`deferred_codegen` is not a valid Julia Method)
If you try to compile a sysimage, the Compiler will auto-generate effectively this `precompile` statement:

```julia
precompile(Tuple{typeof(Enzyme.Compiler.deferred_codegen), Type, Type, Type, Val, Val, Val, Val, Val, Type, Val, Val, Val})
```

The problem is that this function is not valid to be emitted by the Julia compiler:
```julia
@inline function deferred_codegen(
    @nospecialize(fa::Type),
    # ...
    @nospecialize(strongzero::Val)
)
    id = deferred_id_codegen(fa, a, tt, mode, width, modifiedbetween, returnprimal, shadowinit, expectedtapetype, erriffuncwritten, runtimeactivity, strongzero)
    ccall("extern deferred_codegen", llvmcall, Ptr{Cvoid}, (UInt,), id)
end
```

but Julia is generally allowed to come up with any specialization for a Method that it wants and codegen it, so Enzyme should not be relying on this not being precompiled.

This is a practical issue for sysimage builds, where `jl_compile_all_defs` ends up enqueuing the de-specialized version of this function: https://github.com/JuliaLang/julia/blob/25a4410c1081bc6193fc72711025b302f55fc618/src/precompile_utils.c#L188-L190

As a result, Enzyme cannot be built into a sysimage (at least on 1.12, I haven't checked our precompile policy on other versions)

gbaraldi:
Fixes #3091.

## Summary

`Compiler.deferred_codegen`'s body was `ccall("extern deferred_codegen", llvmcall, Ptr{Cvoid}, (UInt,), id)`. The `deferred_codegen` symbol is provided at runtime by `GPUCompiler.register_deferred_codegen` (OrcV2 `absolute_symbols`); on JIT this resolves fine, but AOT linkers (sysimage `compile=all`, juliac, PrecompileTools) walk `jl_compile_all_defs` into this body and fail to resolve the undefined symbol, breaking sysimage builds.

Move the ccall into a thin `@generated` helper. `@generated` functions have no static body until something forces a specialization, so AOT walks during precompile have nothing to despecialize unless the user actually invoked the code — the undefined ccall doesn't leak into the sysimage. JIT behavior is unchanged: at each real call site the generator splices `$id` as a literal into the existing `ccall("extern deferred_codegen", llvmcall, …)`, producing the same IR the old code did.

Mirrors [`GPUCompiler.deferred_codegen(::Val{ft}, ::Val{tt})`](https://github.com/JuliaGPU/GPUCompiler.jl/blob/master/src/driver.jl)'s pattern, which avoids the same issue by keeping its `ccall("extern …)` body behind a generator.

## Test plan

- [ ] CI — Runic formatting, build, unit tests
- [ ] Verified locally on Julia 1.12:
  - `PackageCompiler.create_sysimage([:Enzyme]; incremental=false)` (the AOT path that originally broke) succeeds ✓
  - `nm -D` on the resulting sysimage: `deferred_codegen` is no longer in undefined symbols (only `gpu_malloc`/`gpu_report_oom` from [JuliaGPU/GPUCompiler.jl#611](https://github.com/JuliaGPU/GPUCompiler.jl/issues/611) remain — already fixed in [JuliaGPU/GPUCompiler.jl#808](https://github.com/JuliaGPU/GPUCompiler.jl/pull/808) upstream, out of scope) ✓
  - `Enzyme.autodiff(Reverse, x³, Active(2.0))` → `(12.0,)` ✓
  - Higher-order forward-over-reverse HVP (`test/advanced.jl:1535`, the test the previous attempt regressed) → `[-1.0]` ✓
  - Full `basic` test pattern: 294/294 ✓
  - Full `advanced` test pattern: 221 pass + 2 pre-existing broken ✓

## Files

| File | Change |
|---|---|
| `src/compiler.jl` | Wrap the existing 12-arg `deferred_codegen` body to delegate to a `@generated _deferred_codegen_call(::Val{id})` helper. Helper splices `id` as a literal into the existing `ccall("extern deferred_codegen", llvmcall, …)`. |

### DIFF
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -7462,7 +7462,20 @@ end
     @nospecialize(strongzero::Val)
 )
     id = deferred_id_codegen(fa, a, tt, mode, width, modifiedbetween, returnprimal, shadowinit, expectedtapetype, erriffuncwritten, runtimeactivity, strongzero)
-    ccall("extern deferred_codegen", llvmcall, Ptr{Cvoid}, (UInt,), id)
+    return _deferred_codegen_call(Val(id))
+end
+
+# `@generated` shell so the `ccall("extern deferred_codegen", …)` body
+# isn't a static method body that AOT despecialization
+# (sysimage `compile=all`, juliac, PrecompileTools) trips on — fixes
+# EnzymeAD/Enzyme.jl#3091. Same pattern as
+# `GPUCompiler.deferred_codegen(::Val{ft}, ::Val{tt})`.
+@generated function _deferred_codegen_call(::Val{id}) where {id}
+    id_lit = reinterpret(UInt, id)
+    return quote
+        Base.@_inline_meta
+        ccall("extern deferred_codegen", llvmcall, Ptr{Cvoid}, (UInt,), $id_lit)
+    end
 end
 
 include("compiler/reflection.jl")


src/compiler.jl
```diff
@@ -168,6 +173,9 @@ end
 GPUCompiler.check_invocation(job::CompilerJob{EnzymeTarget}, entry::LLVM.Function) = nothing
 
 GPUCompiler.runtime_module(::CompilerJob{<:Any,<:AbstractEnzymeCompilerParams}) = Runtime
+
+# Enzyme jobs see `deferred_codegen`'s body via the EnzymeMethodTable overlay.
+GPUCompiler.method_table(@nospecialize(job::CompilerJob{<:EnzymeTarget})) = EnzymeMethodTable
```
vchuravy: This will break things w.r.t CUDA iirc

src/compiler.jl
```diff
@@ -95,6 +95,11 @@ GPUCompiler.nest_target(::EnzymeTarget, other::AbstractCompilerTarget) = EnzymeT
 GPUCompiler.have_fma(target::EnzymeTarget, T::Type) = GPUCompiler.have_fma(target.target, T)
 GPUCompiler.dwarf_version(target::EnzymeTarget) = GPUCompiler.dwarf_version(target.target)
 
+# Method table used to hide compiler-internal definitions (e.g. `deferred_codegen`)
+# whose bodies are not valid for normal Julia codegen — putting them behind this
+# table prevents `jl_compile_all_defs` from despecializing them into the sysimage.
+Base.Experimental.@MethodTable EnzymeMethodTable
```
vchuravy: I don't think we currently need an EnzymeMethodTable,

what we need is an entry in CUDA overlay table making `Enzyme.autodiff` use deferred_codegen.

gbaraldi: So I'm a bit worried that wouldn't be enough. Because Julia might decide to precompile these methods regardless

vchuravy: No the goal is to remove `autodiff_deferred` entirely, but the place that is currently using them directly is the GPU integration.

vchuravy: See https://github.com/JuliaGPU/KernelAbstractions.jl/blob/b6121febd6c86721eb62a1741f64e23eb38f4482/ext/EnzymeCore08Ext.jl#L171

gbaraldi: I've never understood what is autodiff_deffered actually doing. (One of the reasons why I'm struggling here). Is the idea to compile a stub with one interpreter, and replace it later with code from another interpreter?

vchuravy: `deferred_codegen` was introduced in GPUCompiler to support the use-case of `dynamic kernels`

Inside a kernel, you can launch another kernel. https://cuda.juliagpu.org/stable/development/kernel/#Dynamic-parallelism

For that, we essentially need a "fake call" that is creating a second GPUCompiler job that gets merged,
Enzyme hijacked that to support two use cases.

1. Higher-order AD
2. AD of GPU kernerls

For both cases we want to essentially perform a full enzyme call and then inline that into the parent module. Billy at some point fixed higher-order AD on the CPU to no longer use the deferred codegen mechanism but rather use inference.

On the GPU the problem remains that CUDA.jl and its abstract interpreter / GPUCompiler technically have no clue of Enzyme's existence. So we want to "switch" to the Enzyme stack during CUDA.jl codegen (re-use large parts of it) and then inline the result into the GPU kernel CUDA.jl is generating.


So my idea here is that we could provide a method overlay through a CUDA extension only that redefines `autodiff` and others (or just the thunk methods) to use the deferred codegen mechanism, instead of having autodiff_deferred in the first place.

gbaraldi: But how would CUDA know when to call that?

vchuravy: > But how would CUDA know when to call that?

By inserting an overlay for autodiff into the CUDA.jl method table?

wsmoses: Maybe we could make a macro which generates the deferred defn, that cuda.jl and amdgpu.jl etc can insert themselves?

That way we control the impl but also its only on the cannot be precompiled method table?

gbaraldi: I guess I was misunderstanding is the Enzyme.autodiff call inside of the kernel?

wsmoses: Yes so the set up in which this is needed is a cuda kernel which calls enzyme

gbaraldi: But @vchuravy does that get rid of the llvmcall method? Because the issue is that method existing in the method table.

test/core/abi.jl
```diff
@@ -473,10 +418,6 @@ end
     @test r[2] ≈ 100.0
     @test r[1][1] ≈ -400.0
     @test r[1][2] ≈ 200.0
-    r = autodiff_deferred(ForwardWithPrimal, Const(rosenbrock_inp), Duplicated, BatchDuplicated(x, (dx_1, dx_2)))
-    @test r[2] ≈ 100.0
```
wsmoses, line 477: You shouldn't delete deferred codegen tests?

wsmoses:
Can we fix the 1.10 thing through the use of our interpreter.

also later versions of cuda.jl still break us so this will need to be dual compatible with old and new cuda.jl

gbaraldi:
I think this requires a bump on the extension on both sides. I guess we can have a bunch of if checks but there's a clear break in both packages? (Or do we want to support older cuda on new enzyme ?)

wsmoses:
We need to support older cuda on newer enzyme

src/compiler.jl
```diff
@@ -7462,7 +7462,20 @@ end
...
-    ccall("extern deferred_codegen", llvmcall, Ptr{Cvoid}, (UInt,), id)
+    return _deferred_codegen_call(Val(id))
+end
+
+# `@generated` shell so the `ccall("extern deferred_codegen", …)` body
+# isn't a static method body that AOT despecialization
+# (sysimage `compile=all`, juliac, PrecompileTools) trips on — fixes
+# EnzymeAD/Enzyme.jl#3091. Same pattern as
+# `GPUCompiler.deferred_codegen(::Val{ft}, ::Val{tt})`.
+@generated function _deferred_codegen_call(::Val{id}) where {id}
```
vchuravy, line 7473: Why is the id a Val? We should just pass it through?

gbaraldi: I was a little afraid that if we don't do this it might creep back into the sysimage. But I'll give it a shot

gbaraldi: It does creep back sadly.

wsmoses (approved):
this is apparently blocking @gbaraldi so I'm going to give a tentative approval to merge, but @gbaraldi please work with @vchuravy post commit to figure out the Val stuff

PR #3121, opened 2026-05-24
EnzymeRules: bind free TypeVars in `has_{f,r}rule_from_sig` query type

Keno:
`_annotate_tt` synthesizes a method-table query like `Tuple{<:FwdConfig, <:Annotation{ft}, Type{<:Annotation}, tt...}` from the caller's signature. The `tt...` entries came from `_annotate`, which builds fresh `TypeVar`s but never wraps them in a `UnionAll`; `ft` (the first parameter of `unwrap_unionall(TT0)`) likewise could contain `TypeVar`s bound by `TT0`'s outer `where` clauses. Both sources left the constructed query with free `TypeVar`s, which used to be tolerated by Julia's subtype/intersection but no longer is under JuliaLang/julia#61876 (free `TypeVar`s now compare by identity, so e.g. `typeintersect(Tuple{…,Const{Float64}}, Tuple{…,V<:Annotation{Float64}})` collapses to `Union{}`).

`_annotate_tt` now returns the new `TypeVar`s alongside `ft, tt`, with `ft` rewrapped in `TT0`'s `UnionAll` chain. Each `has_*_from_sig` caller wraps the constructed query in a `UnionAll` per fresh `TypeVar` via `foldr(UnionAll, …)`. The internal callers in `Enzyme.Compiler.tfunc` are updated to the new return shape.

The behavior of free TypeVars in subtyping was previously undefined (in effect a bug); JuliaLang/julia#61876 makes them well-defined singleton identities, which is what surfaces this latent issue.

This change was AI-generated and should be reviewed carefully before merging.

### DIFF
--- a/lib/EnzymeCore/src/rules.jl
+++ b/lib/EnzymeCore/src/rules.jl
@@ -332,35 +332,30 @@ end
     return augmented_rule_return_type(rct, RT.parameters[1])
 end
 
-function _annotate(@nospecialize(T))
+function _annotate(@nospecialize(T), tvars::Vector{TypeVar})
     if isvarargtype(T)
         VA = T
-        T = _annotate(Core.Compiler.unwrapva(VA))
-        if isdefined(VA, :N)
-            return Vararg{T, VA.N}
-        else
-            return Vararg{T}
-        end
-    else
-        return TypeVar(gensym(), Annotation{T})
+        T = _annotate(Core.Compiler.unwrapva(VA), tvars)
+        return isdefined(VA, :N) ? Vararg{T, VA.N} : Vararg{T}
     end
+    tv = TypeVar(gensym(), Annotation{T})
+    push!(tvars, tv)
+    return tv
 end
 function _annotate_tt(@nospecialize(TT0))
     TT = Base.unwrap_unionall(TT0)
-    ft = TT.parameters[1]
-    tt = []
-    for TTp in TT.parameters[2:end]
-        push!(tt, _annotate(Base.rewrap_unionall(TTp, TT0)))
-    end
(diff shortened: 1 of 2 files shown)


lib/EnzymeCore/src/rules.jl
```diff
@@ -332,44 +332,39 @@ end
...
     TT = Base.unwrap_unionall(TT0)
-    ft = TT.parameters[1]
-    tt = []
-    for TTp in TT.parameters[2:end]
-        push!(tt, _annotate(Base.rewrap_unionall(TTp, TT0)))
-    end
-    return ft, tt
+    ft = Base.rewrap_unionall(TT.parameters[1], TT0)
+    tvars = TypeVar[]
+    tt = Any[_annotate(Base.rewrap_unionall(TTp, TT0), tvars) for TTp in TT.parameters[2:end]]
```
wsmoses, line 349: we changed this to an explicit loop earlier for performance reasons, any reason this needs to be a comprehension?

wsmoses:
@vchuravy @kshyatt I think you guys should review this one

lib/EnzymeCore/src/rules.jl
Keno: No

wsmoses: let's keep the loop then

wsmoses:
bump @vchuravy

lib/EnzymeCore/src/rules.jl
kshyatt: I think this still needs to be fixed?

PR #3131, opened 2026-05-28
Handle case in EnzymeTestUtils of reduction over empties

kshyatt:
This keeps screwing up my life!!!!

### DIFF
--- a/lib/EnzymeTestUtils/src/finite_difference_calls.jl
+++ b/lib/EnzymeTestUtils/src/finite_difference_calls.jl
@@ -71,7 +71,7 @@ function j′vp(fdm, f_vec, ȳ, x)
         end
     end
   end
-  mat = isempty(ẏs) ? zeros(x, 0) : transpose(reduce(hcat, ẏs))
+  mat = isempty(ẏs) ? similar(x, 0) : transpose(reduce(hcat, ẏs))
   result = zero(x)
   for i in 1:length(ȳ)
     tp = @inbounds ȳ[i] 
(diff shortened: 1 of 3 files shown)


lib/EnzymeTestUtils/src/finite_difference_calls.jl
```diff
@@ -71,7 +71,7 @@ function j′vp(fdm, f_vec, ȳ, x)
         end
     end
   end
-  mat = transpose(reduce(hcat, ẏs))
+  mat = isempty(ẏs) ? [] : transpose(reduce(hcat, ẏs))
```
wsmoses: would it be better to have it be similar or zeros of x or something, to keep the type?

wsmoses (commented):
add a test?

lib/EnzymeTestUtils/src/finite_difference_calls.jl
kshyatt: Yeah that could work, I'll try it out!

kshyatt:
> add a test?

I'd love to but it's kind of annoying because this function is only tested implicitly by other stuff... I'll see if I can think of a good way to do it. Or do you mind if I just call it directly in a test?

wsmoses:
basic gist is that if you hit this in your real code, if we don't test it -- there's a chance the functionality is reverted accidentally. so anything that exercises that

kshyatt:
Yeah I'm aware of how tests work lol, I'm just trying to find a good case because clearly this is a weird edge scenario

wsmoses:
ah yeah fair

kshyatt:
Need to engage my "goofball test case generator" powers again...

PR #3146, opened 2026-06-04
Adapt to runtime activity for memcpy

wsmoses:
companion to https://github.com/EnzymeAD/Enzyme/pull/2850


wsmoses:
@jgreener64 this is a bugfix [and required by inbound jll] that breaks molly. It shouldn't, but it likely reveals a different bug.

I may need to go ahead and merge this for the time being, but can I get your help getting a molly mwe, and we can get this fully restored shortly

wsmoses:
actually, think I got it -- waiting for a new jll so I can adapt this side a bit further

jgreener64:
Great, let me know if you need a MWE.

PR #3148, opened 2026-06-06
Add linalg support via GPUArrays extension

### LINKED ISSUE #3122: GPU matmul rules in Enzyme CUDA extension
Following up on [this comment](https://github.com/rsenne/ParallelMCMC.jl/pull/32#discussion_r...) from @wsmoses on rsenne/ParallelMCMC.jl#32 -- I'd like to upstream the GPU matmul EnzymeRules I wrote there.

Plan: a CUDA package extension registering forward / augmented_primal / reverse for Base.* on CuArray / CuMatrix / CuVector (plus transpose/adjoint variants). Rules compute primal and cotangents with plain * so the cuBLAS call stays opaque to Enzyme -- sidesteps the gc-transition abort during LLVM lowering. Width-1 to start.

Reference implementation: [ParallelMCMC.jl/ext/EnzymeExt.jl](https://github.com/rsenne/ParallelMCMC.jl/blob/GPU_Fixes_/ext/EnzymeExt.jl).

OK to open a PR?

rsenne:
Resolves #3122.

Adds custom rules for matmul/dot/sum on GPU arrays.

This came up while building GPU MCMC in [ParallelMCMC.jl#32](https://github.com/rsenne/ParallelMCMC.jl/pull/32), where @wsmoses suggested overloading matmul for GPU arrays directly in Enzyme as an extension rather than keeping per-package wrappers.

### DIFF
--- a/ext/EnzymeGPUArraysCoreExt.jl
+++ b/ext/EnzymeGPUArraysCoreExt.jl
@@ -2,7 +2,8 @@ module EnzymeGPUArraysCoreExt
 
 using GPUArraysCore
 using Enzyme
-using LinearAlgebra: LinearAlgebra, mul!, dot, Transpose, Adjoint, adjoint
+using LinearAlgebra: LinearAlgebra, mul!, dot, Transpose, Adjoint, adjoint,
+    UpperTriangular, LowerTriangular, Symmetric, Hermitian, triu, tril, diag, diagind
 using Enzyme.EnzymeCore: EnzymeCore
 using Enzyme.EnzymeCore.EnzymeRules:
     EnzymeRules,
@@ -54,13 +55,65 @@ end
 _project(::Type{<:Real}, x) = real(x)
 _project(::Type, x) = x
 
-# A GPU array or a (lazy) transpose/adjoint of one; the operand types that show
-# up in `A * B` matmuls, including `transpose(X) * y`.
+# A GPU array, or a structured/lazy wrapper of one; the operand types that show
+# up in `A * B` matmuls (`transpose(X) * y`, `Symmetric(A) * x`, ...). CUBLAS /
+# rocBLAS dispatch these to specialized kernels (trmm/symm/hemm), and the reverse
+# pass projects the cotangent back onto each wrapper's stored entries (see
+# `_accumulate_operand!`). UnitTriangular is intentionally excluded: its diagonal
+# is structurally fixed, so it cannot represent a tangent/cotangent.
 const MaybeWrappedGPU = Union{
     AbstractGPUArray,
     Transpose{<:Any, <:AbstractGPUArray},
     Adjoint{<:Any, <:AbstractGPUArray},
+    UpperTriangular{<:Any, <:AbstractGPUArray},
+    LowerTriangular{<:Any, <:AbstractGPUArray},
+    Symmetric{<:Any, <:AbstractGPUArray},
+    Hermitian{<:Any, <:AbstractGPUArray},
 }
+
+#=
+Accumulate `factor .* G` into the shadow `s` of a matmul operand, where `G` is
+the dense cotangent (`dY·B'` or `A'·dY`). For dense / transpose / adjoint shadows
+the cotangent is added directly. For structured shadows only the stored entries
+are free parameters, so `G` is projected onto that structure:
+
+  UpperTriangular  : keep `triu(G)`
+  LowerTriangular  : keep `tril(G)`
+  Symmetric(uplo)  : off-diagonal (i,j) collects `G[i,j] + G[j,i]`, diagonal `G[i,i]`
+  Hermitian(uplo)  : same with conjugation; the diagonal is real
+=#
+@inline function _accumulate_operand!(s, G, factor)
+    s .+= factor .* G
+    return nothing
+end
+
+function _accumulate_operand!(s::UpperTriangular, G, factor)
+    parent(s) .+= factor .* triu(G)
+    return nothing
+end
+
+function _accumulate_operand!(s::LowerTriangular, G, factor)
+    parent(s) .+= factor .* tril(G)
+    return nothing
+end
+
+function _accumulate_operand!(s::Symmetric, G, factor)
+    p = parent(s)
+    H = G .+ transpose(G)
+    p .+= factor .* (s.uplo == 'U' ? triu(H, 1) : tril(H, -1))
+    dg = view(p, diagind(p))
+    dg .+= factor .* diag(G)
+    return nothing
+end
+
+function _accumulate_operand!(s::Hermitian, G, factor)
+    p = parent(s)
+    H = G .+ adjoint(G)
+    p .+= factor .* (s.uplo == 'U' ? triu(H, 1) : tril(H, -1))
+    dg = view(p, diagind(p))
+    dg .+= factor .* real.(diag(G))
+    return nothing
+end
 
 #=
 mul!(C, A, B, α, β):  C = α·A·B + β·C₀
@@ -190,10 +243,10 @@ function EnzymeRules.reverse(
             Base.@_inline_meta
             dC = _bget(C.dval, Val(N), i)
             if !(A isa Const)
-                _bget(A.dval, Val(N), i) .+= αc .* (dC * adjoint(Bval))
+                _accumulate_operand!(_bget(A.dval, Val(N), i), dC * adjoint(Bval), αc)
             end
             if !(B isa Const)
-                _bget(B.dval, Val(N), i) .+= αc .* (adjoint(Aval) * dC)
+                _accumulate_operand!(_bget(B.dval, Val(N), i), adjoint(Aval) * dC, αc)
             end
             dC .*= βc
             nothing
@@ -291,10 +344,10 @@ function EnzymeRules.reverse(
             Base.@_inline_meta
             dYi = _bget(dY, Val(N), i)
             if !(A isa Const)
-                _bget(A.dval, Val(N), i) .+= dYi * adjoint(cache_B)
+                _accumulate_operand!(_bget(A.dval, Val(N), i), dYi * adjoint(cache_B), true)
             end
             if !(B isa Const)
-                _bget(B.dval, Val(N), i) .+= adjoint(cache_A) * dYi
+                _accumulate_operand!(_bget(B.dval, Val(N), i), adjoint(cache_A) * dYi, true)
             end
             fill!(dYi, zero(eltype(dYi)))
             nothing
--- a/ext/EnzymeGPUArraysCoreExt.jl
+++ b/ext/EnzymeGPUArraysCoreExt.jl
@@ -2,6 +2,18 @@ module EnzymeGPUArraysCoreExt
 
 using GPUArraysCore
 using Enzyme
+using LinearAlgebra: LinearAlgebra, mul!, dot, Transpose, Adjoint, adjoint
+using Enzyme.EnzymeCore: EnzymeCore
+using Enzyme.EnzymeCore.EnzymeRules:
+    EnzymeRules,
+    FwdConfig,
+    RevConfig,
+    Annotation,
+    AugmentedReturn,
+    needs_primal,
+    needs_shadow,
+    overwritten,
+    width
 
 function Enzyme.zerosetfn(x::AbstractGPUArray, i::Int)
     res = zero(x)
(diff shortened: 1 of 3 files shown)


wsmoses:
@kshyatt can you help review?

ext/EnzymeGPUArraysCoreExt.jl
```diff
@@ -35,4 +47,394 @@ end
...
+@inline _bget(x, ::Val{1}, ::Int) = x
+@inline _bget(x, ::Val{N}, i::Int) where {N} = x[i]
+
+# Project an accumulated cotangent onto the (possibly real) parameter type.
+_project(::Type{<:Real}, x) = real(x)
+_project(::Type, x) = x
+
+# A GPU array or a (lazy) transpose/adjoint of one; the operand types that show
+# up in `A * B` matmuls, including `transpose(X) * y`.
+const MaybeWrappedGPU = Union{
```
kshyatt: Do we want to support Triangular and Symmetric/Hermitian too? CUBLAS and ROCBLAS both handle these

ext/EnzymeGPUArraysCoreExt.jl
```diff
@@ -35,4 +47,394 @@ end
...
+          dα  = conj(⟨dC, A·B⟩)
+          dβ  = conj(⟨dC, C₀⟩)
+=#
+
+function EnzymeRules.forward(
+        config::FwdConfig,
+        func::Const{typeof(mul!)},
+        RT::Type{<:Annotation},
+        C::Annotation{<:AbstractGPUArray},
+        A::Annotation,
```
kshyatt: What if A and B are actually CPU arrays?

ext/EnzymeGPUArraysCoreExt.jl
rsenne: added in latest push

ext/EnzymeGPUArraysCoreExt.jl
rsenne: I now restricted it to gpu array types

rsenne:
anything else I can do on this?

wsmoses:
bump @kshyatt

kshyatt (approved):

ext/EnzymeGPUArraysCoreExt.jl
```diff
@@ -35,4 +48,446 @@ end
...
+
+JVP:      dC = α·(dA·B + A·dB) + β·dC₀   (+ dα·A·B + dβ·C₀)
+Pullback: dA += conj(α)·dC·B'
+          dB += conj(α)·A'·dC
+          dC := conj(β)·dC                (cotangent w.r.t. C₀)
+          dα  = conj(⟨dC, A·B⟩)
+          dβ  = conj(⟨dC, C₀⟩)
+=#
+
+function EnzymeRules.forward(
```
wsmoses: @rsenne I'm surprised the forward rules are needed and not handled by the blas handlers atm. What's the error you get without them?

ext/EnzymeGPUArraysCoreExt.jl
```diff
@@ -35,4 +48,446 @@ end
...
+    if A isa Const
+        return A.val * _bget(B.dval, Val(N), i)
+    elseif B isa Const
+        return _bget(A.dval, Val(N), i) * B.val
+    else
+        return _bget(A.dval, Val(N), i) * B.val .+ A.val * _bget(B.dval, Val(N), i)
+    end
+end
+
+function EnzymeRules.forward(
```
wsmoses: you shouldn't need the allocating matmul rule if you have the in place one?

ext/EnzymeGPUArraysCoreExt.jl
rsenne: I think this is/was related to the `gc transition` stuff. So I can try my hand at the copy rule in a PR and see if that makes this irrelevant

wsmoses: so the gc transition stuff is resolved regardless [but yes we also actually need the copy rule]

rsenne: Okay great. I'll retry this on the latest enzyme and confirm that this is till broken, but i ran on my local enzyme just now and it was the `gc transition` bug originally motivating this, so the PR as is maybe unneeded. If this is the case I'll work on the copy rule

wsmoses:
@rsenne with the other one landed can you rebase and adjust this?

rsenne:
yes that was my plan--i'll get on this as soon as i can

ext/EnzymeGPUArraysCoreExt.jl
rsenne: removing all four gives `LLVMException` on every forward case, so the gc-transition fix plus the copy rule didn't make them redundant

i can paste full stacktrace if thats helpful

ext/EnzymeGPUArraysCoreExt.jl
rsenne: i tested this empirically. this was true on `CUDA` but not in general. On `JLArrays`, deleting it makes `A * UpperTriangular(X)` return a wrong cotangent

rsenne: of course i could also rework this into just `CUDA` rules like the copy rules if preferred

ext/EnzymeGPUArraysCoreExt.jl
rsenne: `LLVM.LLVMException: LLVM error: function failed verification (4)`

```
PHI nodes not grouped at top of basic block!
  %"'dual_phi12" = phi { i64, i1 } , !dbg !36290
label %xchg160
PHI nodes not grouped at top of basic block!
  %"'dual_phi14" = phi { i64, i1 } , !dbg !36297
label %xchg166
```

```
[1]  handle_error(reason::Cstring) at context.jl:209
[2]  EnzymeCreateForwardDiff(...) at api.jl:342
[5]  enzyme!(...) at compiler.jl:2804
[6]  compile_unhooked(...) at compiler.jl:6217
[12] thunkbase(...) at compiler.jl:7346
[14] autodiff(::ForwardMode{false, FFIABI, true, false, false}, ::Const{...}, ::Type{Duplicated},
         ::Duplicated{CuArray{Float32, 2, CUDACore.DeviceMemory}},
         ::Duplicated{CuArray{Float32, 2, CUDACore.DeviceMemory}}) at Enzyme.jl:676
```

wsmoses: can you isolate [presumably off of main since not using these rules] the reproducer of?

ext/EnzymeGPUArraysCoreExt.jl
wsmoses: we shouldn't have forward rules here regardless [and also if it gives a wrong result, that's a separate issue]

ext/EnzymeGPUArraysCoreExt.jl
rsenne: here is a MWE @wsmoses 

```
using CUDA, Enzyme
using LinearAlgebra: dot

a  = CuArray(randn(Float32, 8)); b  = CuArray(randn(Float32, 8))
da = CuArray(randn(Float32, 8)); db = CuArray(randn(Float32, 8))

Enzyme.autodiff(Forward, dot, Duplicated, Duplicated(a, da), Duplicated(b, db))
```

This applies for forward mode: `dot(a,b)`, `mul!(C,A,B,1,0)`, `sum(A*B)`

Do you want me to open this as its own issue

ext/EnzymeGPUArraysCoreExt.jl
rsenne: okay i can get rid of all forward rules

rsenne: dropped in latest commit

ext/EnzymeGPUArraysCoreExt.jl
```diff
@@ -35,4 +47,330 @@ end
...
+mul!(C, A, B, α, β):  C = α·A·B + β·C₀
+
+Pullback: dA += conj(α)·dC·B'
+          dB += conj(α)·A'·dC
+          dC := conj(β)·dC                (cotangent w.r.t. C₀)
+          dα  = conj(⟨dC, A·B⟩)
+          dβ  = conj(⟨dC, C₀⟩)
+=#
+
+function EnzymeRules.augmented_primal(
```
wsmoses: okay looking even more closely at linearalgebra, I think we may want to use generic_matmatmul! and generic_matvecmul! on just GPUArray [and not the transposed version of], that version should take an arg as to whether something is transposed, and as a bonus we don't need to write our own accumulate?

rsenne:
@wsmoses i reworked again

ext/EnzymeGPUArraysCoreExt.jl
```diff
@@ -35,4 +47,697 @@ end
...
+    fB = _opflags(tB)
+    a_const = _isconst(config, A)
+    b_const = _isconst(config, B)
+    plain = _plain(tA) && _plain(tB)
+    ntuple(Val(N)) do i
+        Base.@_inline_meta
+        dC = _bget(C.dval, Val(N), i)
+        if !a_const
+            dA = _bget(A.dval, Val(N), i)
+            plain ? _pullback!(dA, tA, dC, (false, false), Bval, _adjop(fB), αc) :
```
wsmoses: I don't think we should need _accumulate_operand! here, we can just use another matmul call? [which takes the scale parameter as a float arg]

ext/EnzymeGPUArraysCoreExt.jl
```diff
@@ -35,4 +47,697 @@ end
...
+accumulation harmless, and backends that do overwrite (CUBLAS trmm) don't care.
+=#
+
+#=
+`adjoint(tfun(T_uploc(A)))`, again with a single lazy wrapper (see
+`_wrap_adjoint`). Only `identity` leaves an adjoint behind, and with it the
+flipped triangle; `transpose` and `adjoint` cancel against the outer one, the
+former down to a conjugation of the data.
+=#
+@inline function _triangular_adjoint(A, uploc::AbstractChar, isunitc::AbstractChar, tfun::F) where {F}
```
wsmoses: Let's do triangular stuff in a follow up and do rectangular first

ext/EnzymeGPUArraysCoreExt.jl
```diff
@@ -35,4 +47,697 @@ end
...
+        end
+    end
+    return (nothing, nothing)
+end
+
+# sum(x): the return cotangent broadcasts onto every element, `dx += dr`.
+
+function EnzymeRules.augmented_primal(
+        config::RevConfig,
+        func::Const{typeof(sum)},
```
wsmoses: we shouldn't need this

rsenne:
@wsmoses ready

ext/EnzymeGPUArraysCoreExt.jl
```diff
@@ -35,4 +46,434 @@ end
...
+An operand contributes no cotangent when it is `Const`, and also when runtime
+activity hands us a `Duplicated` whose shadow is the primal itself -- writing to
+that would corrupt the value the augmented pass just computed.
+=#
+@inline _isconst(_, ::Const) = true
+@inline _isconst(config, x::Annotation) =
+    EnzymeRules.runtime_activity(config) && x.dval === x.val
+
+# Every rule here returns its output argument and stashes a tape.
+@inline _augreturn(config, out::Annotation, tape) = AugmentedReturn(
```
wsmoses: there is already a helper function for this in enzymecore that determines the augmented return type

ext/EnzymeGPUArraysCoreExt.jl
```diff
@@ -35,4 +46,434 @@ end
...
+
+#=
+Cotangent of an `Active` scalar argument: `f(i)`'s raw (possibly complex)
+contribution projected onto `T`, bare at width 1 and an N-tuple when batched.
+Dispatch on `Val{N}` rather than branching on `N == 1`, which infers as `Any` --
+a reverse rule that misses the argument's exact type is a hard error.
+=#
+@inline _dscalar(::Val{1}, ::Type{T}, f::F) where {T, F} = Enzyme._project(T, f(1))::T
+@inline function _dscalar(::Val{N}, ::Type{T}, f::F) where {N, T, F}
+    return ntuple(i -> Enzyme._project(T, f(i))::T, Val(N))
```
wsmoses: avoid the use of closures

wsmoses: though I realize this one is a bit of a noisance

ext/EnzymeGPUArraysCoreExt.jl
```diff
@@ -35,4 +46,434 @@ end
...
+        cache_a, cache_b = tape
+        av = cache_a !== nothing ? cache_a : a.val
+        bv = cache_b !== nothing ? cache_b : b.val
+        N = width(config)
+        ntuple(Val(N)) do i
+            Base.@_inline_meta
+            dr = _dret(dret, Val(N), i)
+            if !_isconst(config, a)
+                # `a` entered conjugated, so its cotangent picks up conj(dr).
+                _bget(a.dval, Val(N), i) .+= conj(dr) .* bv
```
wsmoses, line 483: the daxpy call can be used for this [but unclear if faster/slower]

rsenne:
@wsmoses ready. axpy was slower again so i put the measured speed in a comment for future ref

wsmoses (approved):
okay verbose, and we may want to clean up/simplify later, but I'm fine with this for now

PR #3156, opened 2026-06-09
Thread EnzymeContext for world/MI lookup instead of LLVM attributes

vchuravy:
Stacked on top of #2643 (`vc/gutils_context`).

## What

Use the `world` and a new MI cache carried on `EnzymeContext` instead of scraping per-function LLVM attributes:

- `EnzymeContext` gains a `world` and an `mi_cache::Dict{String, Tuple{MethodInstance, Any}}` (populated in `prepare_llvm`).
- The context is now reachable from `GradientUtils` and the `TypeAnalyzer` via new external-context accessors (`EnzymeGradientUtilsGetExternalContext`, `EnzymeTypeAnalyzerGetExternalContext`), mirroring the existing `Logic` accessor (which is renamed to `enzyme_logic_context`).
- `absint` / `abs_typeof` / `enzyme_custom_extract_mi` take the context and consult the cache, replacing `enzyme_extract_world(fn)` and attribute-only MI extraction at the call sites throughout the rules and compiler.
- Removes the now-unused `emit_methodinstance!`.

## Bug fix

The reverse-mode runtime-activity error path initialized `world = nothing` and only assigned it when `val` was an `Instruction`/`Argument`, then asserted `world == enzyme_context(gutils).world`. For any other `val` (e.g. a constant/global) this asserted instead of emitting the intended `EnzymeRuntimeActivityError`. `world` is now initialized from the context and the tautological assert is dropped.

## Notes

- Committed with `--no-verify`: the whole-file `runic` pre-commit hook wants to reformat large amounts of pre-existing non-runic code on the base branch. Changed lines should be checked with the diff-scoped `git runic vc/gutils_context`.
- Follow-ups still open (not addressed here): the `mi_cache` is keyed by LLVM function name, which is more fragile across modules than per-function attributes; and the `ET_IllegalTypeAnalysis`/`ET_NoType` branches now require the analyzer's external context to be non-null.

---

**2026-09-08:** rebased onto #2643 as of 4987dacd (`enzymejl_world` removed there, world threaded explicitly). Resolutions worth a look: `emit_apply_type!`/`emit_tuple!` take the required `world::UInt` from #2643 followed by this PR's optional `enzyme_context` (used for `absint` of the arguments), so rule sites pass both; the type-analysis handlers take the context as `enzyme_context(TypeAnalyzer(data))` (the ref types are all `Ptr{Cvoid}`); the no-derivative handler gets a `Nothing` context from the batching, truncation and tracing entry points, with `enzyme_custom_extract_mi(::Nothing, f)` falling back to the attributes and the world coming from `callback_world()`. Tests (`typeunstable`, `errors`, `basic`, `rules/rules`, `absint`, `rules/rrules`, `rules/rule_inlining`, `mixed`) pass on Julia 1.12.


wsmoses:
@vchuravy can we do world first, and then mi later. Sole reason being that various other functions will copy an llvm function and those should take the mi too. World is the same everywhere so we don't need to be careful for a first look into

vchuravy:
Replaced by https://github.com/EnzymeAD/Enzyme.jl/pull/3553

Issue #2437, opened 2025-06-18
Enzyme fails to differentiate where Zygote succeeds: `jl_get_builtin_fptr`


FriesischScott:
I ran into an issue optimizing hyper parameters of a Gaussian process with `Optim` and `Enzyme`. I'm using `AbstractGPs` and `ParameterHandling` in my objective function. I have solved this problem with `Zygote` so it should be differentiable.

Using `AutoZygote()` with `Optim` suggests to either set

```julia
autodiff=AutoEnzyme(; mode=Enzyme.set_runtime_activity(Enzyme.Forward))
```
or

 ```julia
autodiff=AutoEnzyme(; mode=Enzyme.set_runtime_activity(Enzyme.Reverse))
```

which fail with 
```
ERROR: Enzyme.Compiler.EnzymeNoDerivativeError(Cstring(0x00007f4605d3885c))
Stacktrace:
  [1] _similar_for
    @ ./array.jl:670 [inlined]
  [2] _collect
    @ ./array.jl:816
  [3] collect_similar
    @ ./array.jl:720 [inlined]
  [4] map
    @ ./abstractarray.jl:3371 [inlined]
  [5] value
    @ ~/.julia/packages/ParameterHandling/WkHof/src/parameters_base.jl:17 [inlined]
  [6] fwddiffe4julia_value_32851wrap
    @ ~/.julia/packages/ParameterHandling/WkHof/src/parameters_base.jl:0
  [7] macro expansion
    @ ~/.julia/packages/Enzyme/Qi7lx/src/compiler.jl:5493 [inlined]
  [8] enzyme_call
    @ ~/.julia/packages/Enzyme/Qi7lx/src/compiler.jl:5027 [inlined]
  [9] ForwardModeThunk
    @ ~/.julia/packages/Enzyme/Qi7lx/src/compiler.jl:4918 [inlined]
 [10] runtime_generic_fwd(activity::Type{…}, runtimeActivity::Val{…}, strongZero::Val{…}, width::Val{…}, RT::Val{…}, f::typeof(ParameterHandling.value), df::Nothing, df_2::Nothing, df_3::Nothing, df_4::Nothing, primal_1::Vector{…}, shadow_1_1::Vector{…}, shadow_1_2::Vector{…}, shadow_1_3::Vector{…}, shadow_1_4::Vector{…})
    @ Enzyme.Compiler ~/.julia/packages/Enzyme/Qi7lx/src/rules/jitrules.jl:327
[...]

PR #3163, opened 2026-06-10
Fix type-unstable newstruct shadow construction for constant and mutable cases

ChrisRackauckas-Claude:
> [!NOTE]
> Draft opened by Chris's agent — please ignore until reviewed/approved by @ChrisRackauckas.

Two bugs in the type-unstable `jl_new_struct*` rules, exposed once #2437's `jl_get_builtin_fptr` fix allowed compilation to reach `runtime_newstruct_augfwd` on real workloads (differentiating a SciML `EnsembleProblem` solve, where SciMLBase's `AggregateLogger{T}` is constructed with a runtime-computed type parameter; see SciML/SciMLSensitivity.jl#1424 and @wsmoses's request there).

### 1. Constant arguments of `ActiveState` type crash `create_shadow_ret`

A constant argument (`ActivityTup[i] == false`) whose type is `ActiveState` (e.g. a `Float64` literal initializing a field) is passed shadow `nothing`, and `create_shadow_ret` had no case for it — it fell through to `batchshadowarg[]`:

```
MethodError: no method matching getindex(::Nothing)
  [2] create_shadow_ret    @ src/rules/typeunstablerules.jl:3 [inlined]
  [4] runtime_newstruct_augfwd(..., ::Type{SciMLBase.AggregateLogger{ConsoleLogger}}, ...)
```

MWE:

```julia
import Enzyme
@noinline get_any()::Any = Base.inferencebarrier(Ref(1.0))
mutable struct Wrap{T}; v::Float64; inner::T; end
f(x) = (w = Wrap(0.0, get_any()); w.v = sum(abs2, x); w.v)
Enzyme.gradient(Enzyme.set_runtime_activity(Enzyme.Reverse), f, [3.0, 1.0])
```

Fixed by giving constant + `ActiveState` arguments a `make_zero(primarg)` shadow. The reverse (`body_construct_rev`) had the matching hole — it unconditionally accumulated into the shadow for any Active/Mixed-state field, which throws `EnzymeNonScalarReturnException` for constant arguments whose shadow is `nothing`; accumulation is now gated on `ActivityTup[i]`.

### 2. Mutable structs got a `MixedState`-style `RefValue` shadow

`runtime_newstruct_augfwd` Ref-wrapped the shadow whenever *any field* was Active/Mixed state. That is correct for immutable structs (overall `MixedState`), but mutable structs are `DupState` overall — downstream consumers derive the shadow representation from `active_reg_nothrow` of the result type and expect a plain shadow object. When the constructed value is later passed to a runtime generic call (or `setfield!`), this produced e.g.

```
MethodError: no method matching EnzymeCore.Duplicated(::SciMLBase.AggregateLogger{ConsoleLogger}, ::Base.RefValue{SciMLBase.AggregateLogger{ConsoleLogger}})
```

in `create_activity_wrapper`. The Ref wrap now applies only to immutable (genuinely `MixedState`) types, with the tape read in `body_construct_rev` mirrored accordingly.

### Verification

- Both MWEs fail on current main and pass with this PR (gradients verified correct: `[6.0, 2.0]` and `[5.0, 3.0]`); both added as regression tests in `test/typeunstable.jl`.
- The second test was additionally verified to fail with only fix 1 applied (`type RefValue has no field v` in `rt_jl_setfield_aug`), i.e. it specifically covers fix 2.
- `test/typeunstable.jl` (including the #2817 runtime-activity tuple construction tests), `test/mixed.jl`, and `test/mixedapplyiter.jl` all pass locally on Julia 1.11.9.
- With this PR the SciML ensemble adjoint progresses past the Julia rules and now hits a separate native crash in Enzyme core C++ (`UpgradeAllocasToMallocs`/`RecursivelyReplaceAddressSpace`, with an `ET_GCRewrite` "Could not find use of stored value" warning on a 271-root sret) — details in SciML/SciMLSensitivity.jl#1424; that one is out of scope here.

### DIFF
--- a/src/rules/typeunstablerules.jl
+++ b/src/rules/typeunstablerules.jl
@@ -27,6 +27,12 @@
                 end
               end
             end
+            if aref == ActiveState
+                return quote
+                    Base.@_inline_meta
+                    make_zero(primarg)
+                end
+            end
         end
         if aref == DupState
             return quote
@@ -128,10 +134,15 @@ function body_construct_augfwd(
         push!(res_syms, sres)
     end
 
+    # Mutable structs are DupState overall: their shadow must be a plain shadow
+    # object, not a MixedState-style RefValue, to match what downstream consumers
+    # derive from active_reg_nothrow of the result type.
+    wrapcond = tuple ? :any_mixed : :(any_mixed && !ismutabletype(NewType))
+
     if Width == 1
         push!(results,
             quote
-                if any_mixed
+                if $wrapcond
                     $(ref_syms[1])
                 else
                     $(res_syms[1])
@@ -139,7 +150,7 @@ function body_construct_augfwd(
             end)
     else
         push!(results, quote
-                if any_mixed
+                if $wrapcond
                     ReturnType(($(ref_syms...),))
                 else
                     ReturnType(($(res_syms...),))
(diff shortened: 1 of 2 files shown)


wsmoses:
subsumed by https://github.com/EnzymeAD/Enzyme.jl/pull/3173

Issue #3179, opened 2026-06-15
Bad call signature for `EnzymeInternalError`


kshyatt:
https://github.com/EnzymeAD/Enzyme.jl/blob/main/src/llvm/transforms.jl#L1217 and https://github.com/EnzymeAD/Enzyme.jl/blob/main/src/llvm/transforms.jl#L1201

but as seen here https://github.com/EnzymeAD/Enzyme.jl/blob/main/src/errors.jl#L785 this struct needs the `mi` and `world` arguments too

wsmoses:
can you make an mwe?

vchuravy:
```
MethodError: no method matching Enzyme.Compiler.EnzymeInternalError(::String, ::String, ::Vector{Base.StackTraces.StackFrame})
  The type `Enzyme.Compiler.EnzymeInternalError` exists, but no method is defined for this combination of argument types when trying to construct it.

  Closest candidates are:
    Enzyme.Compiler.EnzymeInternalError(::String, ::Union{Nothing, String}, ::Union{Nothing, Vector{Base.StackTraces.StackFrame}}, ::MI, ::WT) where {MI, WT}
     @ Enzyme ~/.julia/packages/Enzyme/uVf6m/src/errors.jl:786

  Stacktrace:
    [1] (::Enzyme.Compiler.var"#getparent#nodecayed_phis!##0"{LLVM.Context, LLVM.Function, LLVM.IntegerType, Int64, Dict{LLVM.PHIInst, LLVM.PHIInst}, Dict{LLVM.PHIInst, LLVM.PHIInst}, LLVM.PHIInst, LLVM.GetElementPtrInst})(b::LLVM.IRBuilder, v::LLVM.Value, offset::LLVM.Value, hasload::Bool, phicache::Dict{LLVM.PHIInst, Tuple{LLVM.PHIInst, LLVM.PHIInst}})
      @ Enzyme.Compiler ~/.julia/packages/Enzyme/uVf6m/src/llvm/transforms.jl:1217
[...]

PR #3180, opened 2026-06-15
Fix error throwing error

wsmoses:
partial fix for https://github.com/EnzymeAD/Enzyme.jl/issues/3179


src/llvm/transforms.jl
```diff
@@ -1198,7 +1198,13 @@ function nodecayed_phis!(mod::LLVM.Module)
                                         println(io, " rhs_skipload", rhs_skipload)
                                     end
                                     bt = GPUCompiler.backtrace(inst)
-                                    throw(EnzymeInternalError(msg, string(f), bt))
+				    mi, _ = Compiler.enzyme_custom_extract_mi(fn, false) #=error=#
```
kshyatt, line 1201: This doesn't work because `fn` isn't defined

PR #3203, opened 2026-06-20
Fix non-matched dea arg elimination

wsmoses:


wsmoses:
@copilot fix the test to have a cast for the fn argument callee so it works with untyped pointers like in julia 1.10

wsmoses:
@copilot 

```
Error in testset Mismatched calling convention/function type DAE safety:
Error During Test at /home/runner/work/Enzyme.jl/Enzyme.jl/test/passes.jl:146
  LLVM error: error: constant expression type mismatch: got type 'void (i32 addrspace(10)*, i32 addrspace(10)*)*' but expected 'void (i32 addrspace(10)*, i32 addrspace(10)*)* (i32 addrspace(10)*, i32 addrspace(10)*)*'
    call fastcc void (i32 addrspace(10)*, i32 addrspace(10)*)* bitcast (void (i32*, i32*)* @callee to void (i32 addrspace(10)*, i32 addrspace(10)*)*)(i32 addrspace(10)* %arg1, i32 addrspace(10)* %arg2)
```

PR #3207, opened 2026-06-21
Allow compatibility with Metal 

### LINKED ISSUE #978: Metal support broken
Currently don't pass on CI https://github.com/EnzymeAD/Enzyme.jl/blob/main/test/metal.jl

pw0908:
Fixes #978 by using more flexible type for target machines.


pw0908:
As a quick follow-up, while this PR does allow for Enzyme to work on Metal for simple mathematical operations, it fails when trying to run transcendental functions. As an MWE: 
```julia
using Metal
using Enzyme

function mul_kernel(A)
    i = thread_position_in_grid().x
    if i <= length(A)
        A[i] = log(A[i])
    end
    return nothing
end

function grad_mul_kernel(A, dA)
    autodiff_deferred(Reverse, Const(mul_kernel), Const, Duplicated(A, dA))
    return nothing
end

A = Metal.ones(64,)*2
@metal threads=length(A) mul_kernel(A) # Passes
A = Metal.ones(64,)
dA = similar(A)
dA .= 1
@metal threads=length(A) grad_mul_kernel(A, dA) # Fails
```

Do custom rules need to be defined when running Enzyme on metal?

PR #3211, opened 2026-06-22
[Windows] Fix calling conv tests

wsmoses:


wsmoses:
@gbaraldi @oscardssmith anything obvious missing here?

PR #3137, opened 2026-05-30
Mark make_zero/make_zero! shadow-init builtins inactive in forward mode (#3135)

ChrisRackauckas-Claude:
## Summary

Follow-up to #3135 (per @wsmoses: *"Open the pr for those two functions to start; we can do the eqtable stuff in a follow up"*).

`make_zero` / `make_zero!` emit two bookkeeping builtins while initializing the inner-reverse shadow:

- `jl_field_isdefined_checked` — the `isdefined(prev, i)` field walk over undef-able / abstract fields (`make_zero.jl`).
- `jl_idset_peek_bp` — the `prev in seen` `IdSet` aliasing check (`make_zero!.jl` → `base/idset.jl`).

Enzyme **forward mode** had no rule for either, so nesting **forward-over-reverse** (HVPs/Hessians) over a value with an undef-able/abstract field failed with `EnzymeNoDerivativeError`.

This PR adds the `jl_`/`ijl_` variants of both to **`nofreefns`** and **`inactivefns`** in `src/llvm/attributes.jl`, plus a forward-mode regression testset.

## What this fixes (verified locally)

On a dev'd checkout of `main` (`0670cc2`, Julia 1.11.9), both isolated reproducers fail on `main` and pass with this change, returning the correct derivative:

| Builtin | Reproducer | `main` | this PR |
|---|---|---|---|
| `jl_field_isdefined_checked` | `isdefined(m, i)` field walk, fwd-mode | `No forward mode derivative found for jl_field_isdefined_checked` | `(2.0,)` ✓ |
| `jl_idset_peek_bp` | `v in Base.IdSet`, fwd-mode | `No forward mode derivative found for jl_idset_peek_bp` | `(12.0,)` ✓ |

The two named errors are gone from the issue's MWEs as well.

## Scope / what's deferred

This is **not** sufficient for the full forward-over-reverse HVP yet. With the two builtins suppressed, the `make_zero` / `make_zero!` `seen`-table path cascades to the table-mutation builtins:

- MWE1 (`make_zero`, IdDict `seen`) → `jl_eqtable_get` / `jl_eqtable_put`
- MWE2 (`make_zero!`, IdSet `seen`) → `jl_idset_put_key` / `jl_idset_put_idx`

[...]

Issue #3135, opened 2026-05-29
Forward-over-reverse: no forward rule for jl_field_isdefined_checked / jl_idset_peek_bp in make_zero/make_zero! (inner reverse shadow of value with undef-able/abstract field)


ChrisRackauckas-Claude:
## Forward-over-reverse: `make_zero`/`make_zero!` shadow-init has no forward rule (`jl_field_isdefined_checked` / `jl_idset_peek_bp`) when the inner reverse shadows a value with an undef-able/abstract field

When nesting Enzyme **forward over reverse** (e.g. for HVPs/Hessians), the outer forward pass cannot differentiate the `make_zero` / `make_zero!` shadow-allocation the *inner* reverse uses — **if** the value the inner reverse differentiates contains an **undef-able / abstract / isbits-Union field**. Enzyme forward has no rule for the `jl_field_isdefined_checked` builtin (emitted by `make_zero` when it recurses into a type with possibly-undefined fields) nor for `jl_idset_peek_bp` (emitted by `make_zero!`'s `IdSet` aliasing check).

Reduced to **Enzyme-only** MWEs (no other packages). Julia 1.11.9, Enzyme v0.13.151.

### MWE 1 — `make_zero` → `jl_field_isdefined_checked`

```julia
import Enzyme
RA_R = Enzyme.set_runtime_activity(Enzyme.Reverse)
RA_F = Enzyme.set_runtime_activity(Enzyme.Forward)

struct Par
    w::Vector{Float64}
    extra::Any            # undef-able pointer slot  <-- the necessary ingredient
end
loss1(p) = sum(abs2, p.w)

p  = Par([1.0, 2.0, 3.0], "ignored")
dp = Par([1.0, 0.0, 0.0], "ignored")
Enzyme.autodiff(RA_F, Enzyme.Const(x -> Enzyme.gradient(RA_R, loss1, x)[1]), Enzyme.Duplicated(p, dp))
```

```
EnzymeNoDerivativeError: No forward mode derivative found for jl_field_isdefined_checked
 at context:   %51 = call i32 @jl_field_isdefined_checked(... i64 noundef 0) ...
  [1] make_zero        @ Enzyme/src/typeutils/make_zero.jl:255
  [3] macro expansion  @ Enzyme/src/sugar.jl:334      # inner reverse's shadow init
  [4] gradient         @ Enzyme/src/sugar.jl:274
```

[...]

Issue #1427, opened 2024-05-11
EnzymeTestUtils : fails on GPU Buffers


wsmoses:
@sethaxen I think this should just simply be adding an extension to enzymetestutils for gpuarrays?

mind taking a look?

log
```

test_reverse: scatter! with return activity Const on (::typeof(-), Const), (::ROCArray{Float64, 2, AMDGPU.Runtime.Mem.HIPBuffer}, BatchDuplicated), (::ROCArray{Float64, 2, AMDGPU.Runtime.Mem.HIPBuffer}, BatchDuplicated), (::ROCArray{Int64, 1, AMDGPU.Runtime.Mem.HIPBuffer}, Const): Error During Test at /root/.cache/julia-buildkite-plugin/depots/dc18a9a2-eed5-4c7e-b514-fdcbd06a5a91/packages/EnzymeTestUtils/JLwtd/src/test_reverse.jl:89
--
  | Got exception outside of a @test
  | MethodError: no method matching Base.Threads.Atomic(::Int64)
  |  
  | Closest candidates are:
  | Base.Threads.Atomic()
  | @ Base atomics.jl:81
  |  
  | Stacktrace:
  | [1] map_fields_recursive(f::Function, x::Base.Threads.Atomic{Int64})
  | @ EnzymeTestUtils ~/.cache/julia-buildkite-plugin/depots/dc18a9a2-eed5-4c7e-b514-fdcbd06a5a91/packages/EnzymeTestUtils/JLwtd/src/generate_tangent.jl:9
  | [2] (::EnzymeTestUtils.var"#30#31"{Base.Fix1{typeof(EnzymeTestUtils.rand_tangent), TaskLocalRNG}})(xi::Base.Threads.Atomic{Int64})
  | @ EnzymeTestUtils ~/.cache/julia-buildkite-plugin/depots/dc18a9a2-eed5-4c7e-b514-fdcbd06a5a91/packages/EnzymeTestUtils/JLwtd/src/generate_tangent.jl:7
  | [3] map(f::EnzymeTestUtils.var"#30#31"{Base.Fix1{typeof(EnzymeTestUtils.rand_tangent), TaskLocalRNG}}, t::Tuple{AMDGPU.Runtime.Mem.HIPBuffer, typeof(AMDGPU._free_buf), Base.Threads.Atomic{Int64}})
  | @ Base ./tuple.jl:275
[...]

PR #3213, opened 2026-06-23
Forward-mode rules for jl_eqtable_get/jl_eqtable_put (#3135 follow-up)

ChrisRackauckas-Claude:
## Forward-mode rules for `jl_eqtable_get` / `jl_eqtable_put`

This is the `jl_eqtable_*` follow-up deferred in #3137 / #3135.

### Problem

Forward mode had erroring rules for the eqtable builtins (`eqtableget_fwd` /
`eqtableput_fwd` just `emit_error("Not yet implemented forward for jl_eqtable_*")`
unless the result was provably constant). So **forward-over-reverse** nesting
(HVPs / Hessians) failed whenever the inner reverse's `make_zero` recursed
through its `seen` `IdDict` — which it does to dedup aliased mutable objects.
After #3137 fixed `jl_field_isdefined_checked` / `jl_idset_peek_bp`, the
make_zero forward-over-reverse path advanced to exactly this `jl_eqtable_get`
site (reproduced on released Enzyme in #3135 and SciML/SciMLSensitivity.jl#1427).

### Fix

Replace the erroring `eqtableget_fwd` / `eqtableput_fwd` bodies with real
shadow-propagating rules that mirror the existing reverse augmented-forward
rules (`eqtableget_augfwd` / `eqtableput_augfwd`): invert the table (and
default / value) pointers and re-issue the call on the inverted arguments via
`batch_call_same_with_inverted_arg_if_active!`, including the same mixed-activity
handling for a constant default/value.

One deliberate difference from the reverse put rule: forward mode does **not**
use the `eqtable_shadow_active` guard. In forward mode the tangent *is* the
shadow value, so storing an active scalar value is correct — whereas the reverse
rule rejects it because active scalars are accumulated via the tape. The
constant-result short-circuit at the top of each rule is unchanged, so this can
only *enable* previously-erroring forward cases, never change a constant one.

### What this enables

- Forward-over-reverse through `make_zero`'s `seen` table for aliased objects.
- Plain forward-mode differentiation through `Dict` / `IdDict` build + lookup.

### Tests

Added to `test/make_zero.jl` (auto-discovered, runs on 1.10 and 1.11):

- forward-over-reverse HVP through an aliased value (exercises the `seen` IdDict;
  `loss = 2·sum(abs2,a)` ⇒ Hessian `4I`, checked against `4·seed`),
- plain forward-mode `Dict{Int,Float64}` and `IdDict{Any,Float64}` build+lookup
  (the active-scalar-value case the reverse rule can't take).

Verified locally on Julia 1.11.9 and 1.10.11 (dev'd checkout): full
`test/make_zero.jl` green, and the aliased case fails on stock `main`
(`EnzymeRuntimeException` / `jl_eqtable_get`) and passes with this change.

### Not yet closed

The struct-with-undef-able-field `make_zero` path (#3135 MWE, the SciML #1427
neural ODE) advances **past** `jl_eqtable_get` with this change but then hits
forward-mode **mixed-activity `jl_new_struct`** (`make_zero` reconstructing a
struct with one active + one constant field). That is a separate, deeper
follow-up.

---

@wsmoses this is the eqtable follow-up you asked about in #3135 — opening as draft for your review (per my workflow, please treat as not-ready-to-merge until reviewed).

### DIFF
--- a/src/rules/llvmrules.jl
+++ b/src/rules/llvmrules.jl
@@ -1174,25 +1174,70 @@ end
         return true
     end
 
-    err = emit_error(B, orig, "Enzyme: Not yet implemented forward for jl_eqtable_get")
+    width = get_width(gutils)
 
-    newo = new_from_original(gutils, orig)
-    API.moveBefore(newo, err, B)
-    normal =
-        (unsafe_load(normalR) != C_NULL) ? LLVM.Instruction(unsafe_load(normalR)) : nothing
-    if shadowR != C_NULL && normal !== nothing
-        width = get_width(gutils)
-        shadowres = UndefValue(LLVM.LLVMType(API.EnzymeGetShadowType(width, value_type(orig))))
-        for idx in 1:width
-            if width == 1
-                shadowres = normal
-            else
-                shadowres = insert_value!(B, shadowres, normal, idx - 1)
-            end
-        end
-        unsafe_store!(shadowR, shadowres.ref)
+    origh, origkey, origdflt = arg_operands_view(orig)
+
+    if is_constant_value(gutils, origh)
+        emit_error(
+            B,
+            orig,
+            "Enzyme: Not yet implemented forward constant table in jl_eqtable_get " *
+                string(orig),
+        )
+        return false
     end
 
+    shadowh = invert_pointer(gutils, origh, B)
+
+    shadowdflt = if is_constant_value(gutils, origdflt)
+        shadowdflt2 = julia_error(
+            Base.unsafe_convert(
+                Cstring,
+                "Mixed activity for default of jl_eqtable_get " *
+                    string(orig) *
+                    " " *
+                    string(origdflt),
+            ),
+            orig.ref,
+            API.ET_MixedActivityError,
+            gutils.ref,
+            origdflt.ref,
+            B.ref,
+        )
+        if shadowdflt2 != C_NULL
+            LLVM.Value(shadowdflt2)
+        else
+            nop = new_from_original(gutils, origdflt)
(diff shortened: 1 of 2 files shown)


ChrisRackauckas-Claude:
Confirmed fixed on the released **Enzyme 0.13.169** (Julia 1.11.9) — looks resolved by #3220. The reduced forward-over-reverse `make_zero` MWE from #3135 that was throwing `jl_eqtable_get` on 0.13.163 now returns the correct HVP:

```julia
import Enzyme
RA_R = Enzyme.set_runtime_activity(Enzyme.Reverse)
RA_F = Enzyme.set_runtime_activity(Enzyme.Forward)
struct Par; w::Vector{Float64}; extra::Any; end   # undef-able field = the ingredient
loss1(p) = sum(abs2, p.w)
p  = Par([1.0,2.0,3.0], "ignored"); dp = Par([1.0,0.0,0.0], "ignored")
Enzyme.autodiff(RA_F, Enzyme.Const(x -> Enzyme.gradient(RA_R, loss1, x)[1]), Enzyme.Duplicated(p, dp))
# 0.13.163: EnzymeRuntimeException: Not yet implemented forward for jl_eqtable_get
# 0.13.169: @NamedTuple{1}((Par([2.0, 0.0, 0.0], "ignored"),))   ✅
```

Good to close. The downstream SciML neural-ODE forward-over-reverse (SciML/SciMLSensitivity.jl#1427) now gets past this and stops one step later at `Not yet implemented, mixed activity for jl_new_struct`; I'll open that separately once I can reduce it to a dependency-free MWE.

PR #3222, opened 2026-06-24
1.13 integration ci

wsmoses:
another day another julia


.github/workflows/Integration.yml
vchuravy, line 66:
```suggestion
          version: ${{ matrix.version }}
          include-all-prereleases: true
```

Issue #2925, opened 2026-01-24
MethodError: no method matching blocks(::LLVM.ConstantExpr)


gdalle:
Found this in https://github.com/JuliaDiff/DifferentiationInterface.jl/pull/955 while trying to activate runtime activity on the bug in https://github.com/EnzymeAD/Enzyme.jl/issues/2854, so that DI tests at least pass while it is being investigated on the Enzyme side. It only happens on Julia 1.10, and only with matrices. Strangely enough, the code works when runtime activity is turned off!

```julia
using Enzyme

function arr_to_num(x::AbstractArray)
    return sum(vec(x .^ 4) .* transpose(vec(x .^ 6)))
end

function f(x, c)
    c[1] = arr_to_num(x)
    return c[1]
end

function g(x, c)  # gradient
    dx = zero(x)
    dc = zero(c)
    autodiff(set_runtime_activity(Reverse), f, Duplicated(x, dx), Duplicated(c, dc))
    return dx
end

function h(x, c, dx_batch)  # HVP
    dc_batch = map(dx -> zero(c), dx_batch)
    result = autodiff(
        set_runtime_activity(Forward),
        g,
        BatchDuplicated(x, dx_batch),
        BatchDuplicated(c, dc_batch)
    )
    return result
end

let
    xv = [3.0, 5.0]
    xm = [3.0 5.0]
    dxv = ([1.0, 0.0],)
    dxm = ([1.0 0.0],)
    c = [0.0]
    h(xv, c, dxv)  # works
    h(xm, c, dxm)  # fails on Julia 1.10
end
```

Stacktrace:

```julia
ERROR: MethodError: no method matching blocks(::LLVM.ConstantExpr)

Closest candidates are:
  blocks(::LLVM.Function)
   @ LLVM ~/.julia/packages/LLVM/iza6e/src/core/function.jl:249

Stacktrace:
  [1] 
    @ Enzyme.Compiler ~/.julia/packages/Enzyme/2lHFw/src/gradientutils.jl:135
  [2] get_or_insert_conditional_execute!
    @ ~/.julia/packages/Enzyme/2lHFw/src/gradientutils.jl:105 [inlined]
[...]

PR #3232, opened 2026-06-25
DEA wrapper

wsmoses:


wsmoses:
fixes https://github.com/EnzymeAD/Enzyme.jl/issues/2925

PR #3255, opened 2026-06-28
Make enum padding appear as integer

wsmoses:


vchuravy:
This updates the Project.toml with a dependency that is unused?

wsmoses:
Was fixed on main

src/typetree.jl
```diff
@@ -400,8 +400,8 @@ function typetree_inner(@nospecialize(T::Type), ctx, dl, seen::TypeTreeTable)
             Int(fieldoffset(T, f+1))
         end
 
-        # Fill in padding gaps with Anything, unless subT was an integer
-        padding_T = subT <: Integer ? API.DT_Integer : API.DT_Anything
+        # Fill in padding gaps with Anything, unless subT was an integer or Enum
+        padding_T = (subT <: Integer || subT <: Enum) ? API.DT_Integer : API.DT_Anything
```
vchuravy, line 404: I don't understand why Enum needs to be treated differently?  Or even broader why are we treating Int differently?

PR #3256, opened 2026-06-29
Fix forward deepcopy rule return

wsmoses:
the original change was required to fix a fwd rule return which must be Duplicated{Any}. However this led to an issue where we can't actually build a Duplicated{Any} atm...

@vchuravy is this reasonable?


vchuravy:
Used it as an opportunity to refactor the code a bit, without introducing breaking changes (like a `CongruencyError`)

wsmoses (commented):
sgtm!

vchuravy (approved):

Issue #3253, opened 2026-06-28
Forward-mode `deepcopy` of a general struct under runtime activity: "Unimplemented deepcopy with runtime activity for type ..."


ChrisRackauckas-Claude:
## Forward-mode `deepcopy` of a general struct under runtime activity: `AssertionError: Unimplemented deepcopy with runtime activity for type ...`

Enzyme forward-mode (`set_runtime_activity(Forward)`) differentiation of `Base.deepcopy` applied to a **general struct** (anything that is neither an `Array` leaf nor a purely-active register type) throws:

```
AssertionError: Unimplemented deepcopy with runtime activity for type <T>
  @ Enzyme src/internal_rules/core.jl:59
```

The forward rule `EnzymeRules.forward(::Const{typeof(deepcopy)}, ::Type{<:Duplicated}, ::Duplicated)` delegates to `deepcopy_rtact` (`src/internal_rules/core.jl`), whose mixed-activity (non-`ActiveState`/non-`AnyState`) branch only implements the `RT <: Array` case and otherwise falls straight through to the `throw`:

```julia
if RT <: Array
    # … recurse over elements …
end
throw(AssertionError("Unimplemented deepcopy with runtime activity for type $RT"))
```

So any struct with a mix of active-capable and inactive fields (the common "duplicated register" case) is unsupported.

### Dependency-free MWE

```julia
import Enzyme
const RA_F = Enzyme.set_runtime_activity(Enzyme.Forward)

# Neither an Array nor purely-active: an active-capable pointer field (Vector)
# mixed with an inactive isbits field (Int).
struct Wrap
    v::Vector{Float64}
    tag::Int
end

f(w) = Base.deepcopy(w)
w  = Wrap([1.0, 2.0], 7)
dw = Wrap([1.0, 0.0], 0)
Enzyme.autodiff(RA_F, Enzyme.Const(f), Enzyme.Duplicated, Enzyme.Duplicated(w, dw))
# AssertionError: Unimplemented deepcopy with runtime activity for type Wrap
```

### Control (passes)

`deepcopy` of a plain `Vector` — the one implemented leaf case — works:

```julia
g(x) = Base.deepcopy(x)
Enzyme.autodiff(RA_F, Enzyme.Const(g), Enzyme.Duplicated, Enzyme.Duplicated([1.0, 2.0], [1.0, 0.0]))
# (var"1" = [1.0, 0.0],)
```

### Suggested fix

[...]

Issue #3246, opened 2026-06-27
Forward-over-reverse: "mixed activity for jl_new_struct" when a runtime-computed type parameter forces generic construction


ChrisRackauckas-Claude:
## Forward-over-reverse: "Not yet implemented, mixed activity for jl_new_struct" when a runtime-computed type parameter forces generic construction

Enzyme forward-over-reverse (an HVP: outer `Forward` over inner `gradient(Reverse)`, both under `set_runtime_activity`) throws

```
Enzyme: Not yet implemented, mixed activity for jl_new_struct constants=Bool[1, 0, 0]
  ... @ijl_new_structv ...
  Tuple{Bool, DataType, GPUCompiler.ArgumentCC}[
    (1, Vector{Float64},          GPUCompiler.BITS_REF),   # u0     inactive
    (1, Tuple{Float64, Float64},  GPUCompiler.MUT_REF),    # tspan  active
    (1, Vector{Float64},          GPUCompiler.MUT_REF)]    # p      active
```

when a struct is built whose **type parameter is computed from a runtime value** (so the construction is type-unstable and Enzyme lowers the `jl_new_struct` through its generic runtime, `runtime_generic_fwd`/`runtime_generic_augfwd`), and that struct mixes an **inactive pointer field**, an **inactive isbits field**, and the **active pointer field** in one allocation.

### Dependency-free MWE (Enzyme only)

```julia
import Enzyme
const RA_R = Enzyme.set_runtime_activity(Enzyme.Reverse)
const RA_F = Enzyme.set_runtime_activity(Enzyme.Forward)

# A runtime Bool Enzyme/Julia cannot constant-fold (stands in for SciML's `isinplace(f, 4)`).
const FLAG = Ref(true)
@noinline runtime_iip() = FLAG[]

# ODEProblem-shaped immutable struct, parameterized by `iip`.
struct Prob{iip, U, T, P}
    u0::U      # inactive array
    tspan::T   # inactive isbits tuple
    p::P       # active array
end
Prob{iip}(u0, tspan, p) where {iip} =
    Prob{iip, typeof(u0), typeof(tspan), typeof(p)}(u0, tspan, p)

const U0  = [1.0, 2.0]
const TSP = (0.0, 1.0)

# Type-unstable construction, behind its own call boundary (like ODEProblem's ctor chain).
@noinline build(u0, tspan, p) = Prob{runtime_iip()}(u0, tspan, p)

[...]

PR #3261, opened 2026-06-29
Forward-mode deepcopy: handle Const argument with Duplicated return (fixes #3258)

### LINKED ISSUE #3258: Forward-mode `deepcopy` with a `Const` argument and `Duplicated` return: missing `EnzymeRules.forward` method
## Forward-mode `deepcopy` with a `Const` argument and a `Duplicated` return: missing `EnzymeRules.forward` method

Follow-on to #3253 (fixed by #3254, which added general-struct recursion to `deepcopy_rtact`). With that in, the next forward-mode `deepcopy` case is still uncovered: when the argument is `Const` but a `Duplicated` return is requested (the runtime-activity-widening path), there's no matching `EnzymeRules.forward` method:

```
MethodError: no method matching forward(
  ::EnzymeRules.FwdConfigWidth{1, true, true, true, false},
  ::Const{typeof(deepcopy)},
  ::Type{Duplicated{Vector{Float64}}},
  ::Const{Vector{Float64}})
```

The forward `deepcopy` rules in `src/internal_rules/core.jl` define only `Duplicated`/`BatchDuplicated` (and the `…NoNeed` variants) *argument* methods; none accept a `Const` argument while returning `Duplicated`.

### Dependency-free MWE

```julia
import Enzyme
const RA_F = Enzyme.set_runtime_activity(Enzyme.Forward)

f(x) = Base.deepcopy(x)
x = [1.0, 2.0]

Enzyme.autodiff(RA_F, Enzyme.Const(f), Enzyme.Const(x))
# MethodError: no method matching forward(
#   ::FwdConfigWidth{1, true, true, true, false}, ::Const{typeof(deepcopy)},
#   ::Type{Duplicated{Vector{Float64}}}, ::Const{Vector{Float64}})
```

(The same `MethodError` also occurs under plain `Enzyme.Forward` — `FwdConfigWidth{1, true, true, false, false}` — so it isn't strictly runtime-activity-specific, but runtime activity is how the SciML workload reaches it.)

### Control (passes)

`Duplicated` argument — the implemented path — works:

```julia
Enzyme.autodiff(RA_F, Enzyme.Const(f), Enzyme.Duplicated(x, [1.0, 0.0]))
# (var"1" = [1.0, 0.0],)
```

### Suggested fix

Add a `forward(::FwdConfig, ::Const{typeof(deepcopy)}, ::Type{<:Duplicated}, x::Const)` method (and the `BatchDuplicated`/`…NoNeed` analogues) returning the primal `deepcopy(x.val)` with a zeroed shadow (`Enzyme.make_zero` of the copy), so an inactive value can be deepcopied where a `Duplicated` result is requested.

### Context

Next blocker (after #3135 → #3137, #3246 → #3251, #3253 → #3254) for an Enzyme forward-over-reverse HVP of a neural-ODE loss in **SciML/SciMLSensitivity.jl#1427**. With #3253 fixed on 0.13.172, the workload now reaches the inner reverse, where DiffEqBase's `solve_up` rule deepcopies `u0` (a `Vector{Float32}`, inactive → `Const`) while the surrounding forward-over-reverse requests a `Duplicated` result — hitting exactly the `MethodError` above.

### Versions

Enzyme **0.13.172** (latest release), Julia 1.11.9, x86_64-linux.

ChrisRackauckas-Claude:
## Summary

Fixes #3258. Adds the missing forward-mode `EnzymeRules.forward` methods for `Base.deepcopy` when the **argument is `Const`** but a `Duplicated`/`BatchDuplicated` result is requested (the runtime-activity-widening path).

Previously `src/internal_rules/core.jl` defined forward `deepcopy` rules only for `Duplicated`/`BatchDuplicated` (and the `…NoNeed`) *arguments*. With a `Const` argument and a `Duplicated` return requested, no method matched:

```
MethodError: no method matching forward(
  ::EnzymeRules.FwdConfigWidth{1, true, true, true, false},
  ::Const{typeof(deepcopy)},
  ::Type{Duplicated{Vector{Float64}}},
  ::Const{Vector{Float64}})
```

## Fix

Add the four `Const`-argument variants — `Duplicated`, `DuplicatedNoNeed`, `BatchDuplicated`, `BatchDuplicatedNoNeed` — returning the deepcopied primal paired with a freshly zeroed shadow (`Enzyme.make_zero` of the copy). The derivative of deepcopying an inactive value is zero, so a zeroed shadow is the correct result. Batch width is taken from `EnzymeRules.width(config)`.

## Verification (run locally, Julia 1.11.9, this branch)

Issue MWE — previously a `MethodError`, now returns a zeroed shadow:

```julia
import Enzyme
const RA_F = Enzyme.set_runtime_activity(Enzyme.Forward)
f(x) = Base.deepcopy(x)
x = [1.0, 2.0]
Enzyme.autodiff(RA_F, Enzyme.Const(f), Enzyme.Const(x))
# (var"1" = [0.0, 0.0],)
```

Output observed:
```
Const arg (RA): (var"1" = [0.0, 0.0],)
Const arg (plain): (var"1" = [0.0, 0.0],)
Duplicated arg control: (var"1" = [1.0, 0.0],)            # implemented path still works
BatchDuplicated arg control: (var"1" = (var"1" = [1.0, 0.0], var"2" = [0.0, 1.0]),)
```

## Tests

Added a testset to `test/rules/internal_rules/deepcopy.jl` covering the `Const`-argument case through the public API (width-1 `Duplicated` under both `Forward` and runtime-activity `Forward`) and driving the rule methods directly to exercise the `BatchDuplicated` (width 2) and `…NoNeed` variants, which the public API collapses to width-1 `Duplicated` when all arguments are `Const`. Full file passes locally (17/17).

## Context

This is the next blocker (after #3135 → #3137, #3246 → #3251, #3253 → #3254) for an Enzyme forward-over-reverse HVP of a neural-ODE loss in SciML/SciMLSensitivity.jl#1427, where DiffEqBase's `solve_up` rule deepcopies an inactive `u0` (`Const`) while the surrounding forward-over-reverse requests a `Duplicated` result.

---

Please ignore until reviewed by @ChrisRackauckas. Implemented at @wsmoses's request in the issue thread.

### DIFF
--- a/src/internal_rules/core.jl
+++ b/src/internal_rules/core.jl
@@ -129,6 +129,52 @@ function EnzymeRules.forward(
     end)
 end
 
+# A `Const` argument carries no shadow, but a `Duplicated` result may still be
+# requested (e.g. via runtime-activity widening), so deepcopy the primal and
+# pair it with a freshly zeroed shadow.
+function EnzymeRules.forward(
+    config::EnzymeRules.FwdConfig,
+    func::Const{typeof(Base.deepcopy)},
+    ::Type{<:DuplicatedNoNeed},
+    x::Const,
+)
+    return Enzyme.make_zero(func.val(x.val))
+end
+
+function EnzymeRules.forward(
+    config::EnzymeRules.FwdConfig,
+    func::Const{typeof(Base.deepcopy)},
+    ::Type{<:BatchDuplicatedNoNeed},
+    x::Const,
+)
+    primal = func.val(x.val)
+    return ntuple(Val(EnzymeRules.width(config))) do _
+        Enzyme.make_zero(primal)
+    end
+end
+
+function EnzymeRules.forward(
+    config::EnzymeRules.FwdConfig,
+    func::Const{typeof(Base.deepcopy)},
+    ::Type{<:Duplicated},
+    x::Const,
+)
+    primal = func.val(x.val)
+    return Duplicated(primal, Enzyme.make_zero(primal))
+end
+
+function EnzymeRules.forward(
+    config::EnzymeRules.FwdConfig,
+    func::Const{typeof(Base.deepcopy)},
+    ::Type{<:BatchDuplicated},
+    x::Const,
+)
+    primal = func.val(x.val)
+    return BatchDuplicated(primal, ntuple(Val(EnzymeRules.width(config))) do _
+        Enzyme.make_zero(primal)
+    end)
+end
+
 function EnzymeRules.augmented_primal(
     config::EnzymeRules.RevConfig,
     func::Const{typeof(Base.deepcopy)},
--- a/test/rules/internal_rules/deepcopy.jl
+++ b/test/rules/internal_rules/deepcopy.jl
@@ -1,4 +1,5 @@
 using Enzyme
+using Enzyme.EnzymeRules
 using Test
 
 @testset "core fallback traversal" begin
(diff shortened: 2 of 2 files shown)


src/internal_rules/core.jl
```diff
@@ -129,6 +129,52 @@ function EnzymeRules.forward(
...
+# A `Const` argument carries no shadow, but a `Duplicated` result may still be
+# requested (e.g. via runtime-activity widening), so deepcopy the primal and
+# pair it with a freshly zeroed shadow.
+function EnzymeRules.forward(
+    config::EnzymeRules.FwdConfig,
+    func::Const{typeof(Base.deepcopy)},
+    ::Type{<:DuplicatedNoNeed},
+    x::Const,
+)
+    return Enzyme.make_zero(func.val(x.val))
```
wsmoses, line 141: @ChrisRackauckas-Claude we can avoid the func.val here and just apply make_zero, right? [and similar below]

wsmoses: @ChrisRackauckas-Claude actually never mind this is correct as is because of other non-differentiable fields that need to be copied

wsmoses (approved):

Issue #3259, opened 2026-06-29
Reverse-mode over ModelingToolkit MTKParameters promotion (__apply_copy_template) -> undersized shadow Memory BoundsError / TypeAnalysis hang (v0.13.171)


ChrisRackauckas-Claude:
## Summary
Reverse-mode `Enzyme.autodiff`/`Enzyme.gradient` (with `set_runtime_activity`) that differentiates **ModelingToolkit's `MTKParameters` tunable-eltype promotion** — `ModelingToolkitBase.__apply_copy_template`, a `mapreduce`/`vcat` over a `Vector{Union{UnitRange{Int64}, SciMLStructures...ParameterIndex{Tunable,UnitRange}}}` — fails. The failure mode is **state/compilation-order dependent**:

1. **Runtime `BoundsError: attempt to access MemoryRef{Float64} at index [N]`** (N≈6–7). Enzyme's `override_bc_mapreduce` builds a shadow `Memory{Float64}` shorter than the primal, and the in-place `copyto!` overruns it.
2. **TypeAnalysis blowup/hang** (cold): >20 min stuck in `checkedOrIn` / `TypeTree` merge (`typetree_inner`, `src/typetree.jl:441`).
3. **`error: Linking globals named 'julia___NNNNN': symbol multiply defined!`** (when a similar function was compiled earlier in the same session).

## Versions
Julia 1.10.11 / 1.11 / 1.12, Enzyme **v0.13.171**, ModelingToolkit v11.29.0. (Originates in SciML/SciMLSensitivity.jl CI — Core8 — where it is the runtime BoundsError form.)

## Repro
Reverse-diff a `solve` of an MTK DAE (algebraic constraint ⇒ DAE initialization) whose adjoint computes the init-sensitivity gradient with a nested `Enzyme.gradient`:
```julia
using ModelingToolkit, OrdinaryDiffEq, SciMLSensitivity, Enzyme
import SciMLStructures as SS
using ModelingToolkit: t_nounits as t, D_nounits as D
@parameters σ ρ β; @variables x(t) y(t) z(t) w(t) w2(t)
eqs = [D(D(x)) ~ σ*(y-x), D(y) ~ x*(ρ-z)-y, D(z) ~ x*y-β*z,
       w ~ x+y+z+2β, 0 ~ x^2 + y^2 - w2^2]
@mtkbuild sys = ODESystem(eqs, t)
[...]

PR #3279, opened 2026-07-03
Add link_split_existing! to fix spurious "symbol multiply defined" link error (#3259)

ChrisRackauckas-Claude:
### Summary

Fixes the `error: Linking globals named 'julia___NNNNN': symbol multiply defined!` failure from #3259 (mode #3).

When Enzyme links a deferred / nested-codegen submodule (`enzyme_context.modules_to_link`) into the main compile unit, a function — e.g. a `julia___NNNNN` specialization — can be **defined in both** modules. `LLVM.link!` then aborts fatally with `symbol multiply defined!` before compilation can finish.

This adds `link_split_existing!(mod, newmod)`: before linking, it renames any function that is defined in both modules, giving the incoming (`newmod`) copy a unique suffix so both definitions are preserved as distinct symbols. It's wired into the two `modules_to_link` deferred-link sites.

### Why the collision is spurious

I verified on the #3259 reproducer (reverse-mode over an MTK `DefaultInit` DAE solve with `GaussAdjoint`/`EnzymeVJP`): with this change and **stock LLVM.jl**, the previously-failing `DefaultInit` case compiles through and returns the **correct** gradient — identical (to 14 digits) to the `BrownFullBasicInit` path and to the OP's expected values:

```
dtun = [-88324.83309568025, 46528.856366258646, 99324.87439708965]
```

So the duplicate definitions were the same symbol; splitting them off is safe.

### Test

Adds a `link_split_existing!` testset in `test/passes.jl`: two modules each defining `@julia___dup`; after `link_split_existing!` the destination keeps its own `@julia___dup`, the incoming copy is preserved as `@julia___dup_split`, non-colliding functions link unchanged, and the incoming caller is retargeted to the split copy. (Passes locally, 7/7.)

---

Please ignore until reviewed by @ChrisRackauckas.

https://claude.ai/code/session_01Mx15rEbvLviquZn9yk5mgL


src/compiler.jl
```diff
@@ -5314,6 +5314,34 @@ const DumpPreCheck = Ref(false)
...
+before linking, so its definition is preserved as a distinct symbol instead of
+triggering a `symbol multiply defined` linker error.
+"""
+function link_split_existing!(mod::LLVM.Module, newmod::LLVM.Module)
+    modfns = functions(mod)
+    newfns = functions(newmod)
+    for f in collect(newfns)
+        isdeclaration(f) && continue
+        fname = LLVM.name(f)
+        (haskey(modfns, fname) && !isdeclaration(modfns[fname])) || continue
```
wsmoses: @ChrisRackauckas-Claude if !haskey modfns, you should continue, then you can do the declaration fn check

ChrisRackauckas-Claude: Done in 5e1c144 — split into `haskey(modfns, fname) || continue` then `isdeclaration(modfns[fname]) && continue`. Re-ran the unit test (7/7) and Runic is clean.

vchuravy:
This is generally wrong, we should figure out why we are getting duplication instead of papering over it during the link stage.

wsmoses:
@vchuravy so I actually do think this is right/necessary and understand one way it can pop up. So via gpucompiler we recursively get all fn defs that are in a type stable set. So if we want to AD f(), and f calls g, we in the main module have g. Suppose for a custom rule we need the def of h, which also calls g. The nested module contains the def of h and g. Thus we end up with the multiple symbol collision on link.

vchuravy:
There should be a lookup table/ worklist we go through that ensures that we only perform codegen once.

wsmoses:
we have that, but only for the outermost mi we generate for, not any of the dependency functions

wsmoses:
this is said worklist: https://github.com/EnzymeAD/Enzyme.jl/blob/03439e85af6615135cdb2c7079ef19368cd8ecf5/src/compiler.jl#L6053

vchuravy:
No, this is still the wrong change. We are generating excessively much code and then discarding it in the end is not helpful.

GPUCompiler already has a similar worklist where we map from CI to llvm function to avoid compiling the same method a dozen times. This should use a similar method.

PR #3271, opened 2026-07-01
CI + target transformation info for Metal support

pw0908:

PR #3281, opened 2026-07-05
Guard uses of `GPUCompiler.llvm_targetinfo`

giordano:
#3271 broke Oceananigans because it used `GPUCompiler.llvm_targetinfo`, which [was introduced](https://github.com/JuliaGPU/GPUCompiler.jl/pull/782) in [GPUCompiler@v1.10.0](https://github.com/JuliaGPU/GPUCompiler.jl/releases/tag/v1.10.0), but #3271 didn't change the compat bound of GPUCompiler.

***Edit***: also retroactively changing compat bounds in the registry with https://github.com/JuliaRegistries/General/pull/160099.


wsmoses:
instead of changing compat can we do isdefined(GPUCompiler, :llvm_target), where relevant?

giordano:
Not just define (that'd be easy to guard), it's also used at https://github.com/EnzymeAD/Enzyme.jl/blob/37b0ab3f33e6829653d235178c8ca2ccf95cca0a/src/compiler.jl#L5645 and then https://github.com/EnzymeAD/Enzyme.jl/blob/37b0ab3f33e6829653d235178c8ca2ccf95cca0a/src/compiler.jl#L5680 https://github.com/EnzymeAD/Enzyme.jl/blob/37b0ab3f33e6829653d235178c8ca2ccf95cca0a/src/compiler.jl#L6235 I don't know what would happen if you guard those lines for old GPUCompiler versions.

vchuravy:
Maybe better to use a `isedefined` instead?

giordano:
> Maybe better to use a `isedefined` instead?

Yes, but would that break anything if guarding the [lines linked above](https://github.com/EnzymeAD/Enzyme.jl/pull/3281#issuecomment-4887530234)?

giordano:
Uhm, both `optimize!` and `post_optimize!` got an optional argument `tti` in #3271, so setting `target_info` to `nothing` if `GPUCompiler.llvm_targetinfo` is not defined should be fine.

wsmoses:
yeah that's fine

PR #3288, opened 2026-07-06
Fix Windows heap corruption from literal-pointer symbol name collisions

gbaraldi:
Fixes the Windows 1.10 CI crash (`STATUS_HEAP_CORRUPTION`, exit code 3221226356) debugged via #3228.

## The bug

On Julia 1.10, ccalls reach Enzyme's module as literal `inttoptr` addresses. For each such call site, `check_ir!` recovers a symbol name via `jl_lookup_code_address`, creates a declaration under that name carrying `enzymejl_needs_restoration=<ptr>`, and `restore_lookups` later replaces the declaration with its recorded address.

`jl_lookup_code_address` reports the *nearest preceding* symbol, and on Windows (export-table-only symbol info) two distinct pointers can come back with the same name. The second call site then reused the first site's declaration, and restoration stamped **one address onto both call sites**.

In the post-`restore_lookups` IR from the #3228 debug run, the compiled `Base.GMP.MPZ.set_si(1)` (building the literal `1` of the BigFloat `@easy_rule` shadow) shows both GMP calls resolved to the same address:

```llvm
call void inttoptr (i64 1791842560 to void (i8*, i32)*)(i8* %coercion, i32 0)  ; dbg scope: init2!
call void inttoptr (i64 1791842560 to void (i8*, i32)*)(i8* %coercion, i32 1)  ; dbg scope: set_si!
```

so `__gmpz_init2` never runs and `mpz_set_si` writes a limb through the uninitialized `d` pointer of a non-zeroed pool allocation → heap corruption. On Julia 1.11+ the same call sites go through the GOT path and keep their identity as `ejlstr$__gmpz_init2$libgmp-10.dll` / `ejlstr$__gmpz_set_si$libgmp-10.dll`, which is why only 1.10 crashes. (Windows is also the only platform where the easy-rule literal `1` takes the BigInt/GMP path at all, since `Clong == Int32` there.)

## The fix

Two layers in `check_ir!`:

1. **Verify recovered names against the loader** (`resolve_symbol_name`): ask which module contains the pointer (`dladdr` on POSIX, `GetModuleHandleExW(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS)` on Windows) and resolve the candidate name within exactly that module. If it resolves to a different address, the name is provably wrong and is dropped (the call site keeps its correct literal pointer). The lookup is module-scoped on purpose: a process-wide search could find an unrelated library's copy of a symbol (libm vs openlibm, etc.) and misreport a correct name; comparing addresses instead of names also keeps exported aliases working.
2. **Never merge call sites whose pointers differ**: when a declaration with the recovered name already exists but records a different restoration pointer, key the new declaration by pointer instead of reusing it. This is the backstop for names that cannot be verified.

This is the same disease as the existing macOS `memcpy`/`memmove` collision workaround in `FFI.ptr_map` (curated names remain trusted as-is).

## Testing

- Julia 1.10.11 and 1.12.6 (Linux): the #3228 reproducer passes (`autodiff(Forward, +, Duplicated, Duplicated(a, da), Duplicated(b, db))[1] ≈ da + db`), full `test/rules/internal_rules/bigfloat.jl` passes, and `test/passes.jl` passes including the new `"Literal-pointer symbol resolution"` testset covering match / nearest-symbol mismatch / unresolvable-file / non-module-pointer cases.
- Windows 1.10 is the real target; CI on this PR is the proof.


wsmoses:
@vchuravy I'm going to merge for now to unblock windows but please post-commit review this

PR #3302, opened 2026-07-09
Add forward/reverse mode counterpart helpers to EnzymeCore

rsenne:
As discussed here, I've implemented the forward/reverse counterparts into EnzymeCore so that upstream changes will not break the usage in DI.

This adds two unexported helpers alongside the existing `Split``/Combined`:

`forward_counterpart(::Mode)`: maps `ReverseMode`/`ReverseModeSplit` to a` ForwardMode`, preserving `ReturnPrimal`, `ABI`, `ErrIfFuncWritten`,` RuntimeActivity`, and `StrongZero` (drops `Holomorphic` and split-specific parameters). Identity on `ForwardMode`.
`reverse_counterpart(::Mode)`: maps` ForwardMode` to a `ReverseMode` preserving the same settings, with` Holomorphic = false`. Identity on `ReverseMode`/`ReverseModeSplit`.
Happy to rename or adjust semantics as you see fit. ~Also bumps `EnzymeCore` to 0.8.22.~

cc @wsmoses @gdalle


Project.toml
```diff
@@ -41,7 +41,7 @@ ADTypes = "1"
 BFloat16s = "0.2, 0.3, 0.4, 0.5, 0.6"
 CEnum = "0.4, 0.5"
 ChainRulesCore = "1"
-EnzymeCore = "0.8.21"
+EnzymeCore = "0.8.22"
```
wsmoses: this doesn't need a bump

lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -832,4 +832,68 @@ end
 
 Combined(mode::ReverseMode) = mode
 
+"""
+    forward_counterpart(::Mode)
+
+Turn a [`ReverseMode`](@ref) or [`ReverseModeSplit`](@ref) object into a [`ForwardMode`](@ref) object while preserving as many of the settings as possible.
```
wsmoses, line 838: implementation lgtm, but the question is what the name should be

@vchuravy thoughts?

lib/EnzymeCore/src/EnzymeCore.jl
```diff
@@ -832,4 +832,68 @@ end
...
+        StrongZero,
+        ABI,
+        Holomorphic,
+        ErrIfFuncWritten,
+    }
+    return ForwardMode{ReturnPrimal, ABI, ErrIfFuncWritten, RuntimeActivity, StrongZero}()
+end
+
+forward_counterpart(mode::ReverseModeSplit) = forward_counterpart(Combined(mode))
+forward_counterpart(mode::ForwardMode) = mode
```
wsmoses: I'm not sure we should have this defined? what's the argument for it

rsenne: The identity method?  The purpose is to make forward_counterpart total over all modes, so callers can apply it unconditionally to normalize a user-supplied mode.. Maybe a downstream package has some method that does candidate selection of the AD mode for performance reasons:

```julia

function select_mode(f, mode::EnzymeCore.Mode)
    if cheaper_in_forward(f)
        forward_counterpart(mode)
    else
        reverse_counterpart(mode)
    end
end
```

wsmoses: Yeah the name counterpart to me just implies that it's only valid to swap the mode, so if we want that to be valid maybe a different name?

gdalle: Maybe `to_forward` and `to_reverse`?

rsenne: @wsmoses how do you feel about above?

wsmoses: I'm okay with those names, but don't super love them. @vchuravy any thoughts/other ideas?

rsenne: `as_forward_mode` / `as_reverse_mode` ?

wsmoses: I like as_forward / as_reverse !

rsenne: okay I'll do the name change!

gdalle (approved):

rsenne:
@wsmoses anything else left on this?

wsmoses:
A review from @vchuravy

PR #3308, opened 2026-07-10
Adapt to GPUCompiler 2.0 (CompilerCaching.jl)

### LINKED ISSUE #3492: GPUCompiler v2?
The new CUDA.jl etc are all on GPUCompiler.jl v2 which is (iirc) the CompilerCaching based thing, it would be great if Enzyme was also compatible with that.  cc @ChrisRackauckas

vchuravy:
Adapts Enzyme to JuliaGPU/GPUCompiler.jl#794, which moves GPUCompiler's code-instance caching onto CompilerCaching.jl:

- `GPUCompiler.ci_cache_token` is now `GPUCompiler.cache_owner`
- `GPUCompiler.ci_cache` (Julia 1.10 only) is now `GPUCompiler.get_code_cache`
- `GPUCompiler.codegen` was removed; use `compile_unhooked` directly
- `GPUCompiler.runtime_slug` was removed (the cache owner subsumes it)
- `GPUCompiler.compile_method_instance` now drives inference through `drive_inference!`, which is only provided for `GPUInterpreter`; add methods for `EnzymeInterpreter` (via `CompilerCaching.typeinf!` on 1.11+, and a copy of the 1.10 `CodeCache`-based implementation)

Close #3339


Project.toml
```diff
@@ -66,3 +66,8 @@ StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 
 [sources.EnzymeCore]
 path = "lib/EnzymeCore"
+
+# XXX: remove before merging, once GPUCompiler 2.0 is released
+[sources.GPUCompiler]
+url = "https://github.com/JuliaGPU/GPUCompiler.jl"
+rev = "tb/compilercaching"
```
giordano:
```suggestion
```

christiangnrd:
> @vchuravy has a pr starting to look into it.
>
> That said, @maleadt as the changes here don't look huge, is it possible to conditionally gate this with something like
>
> if gpucompilerver >= 2
>  a()
>  else
>   b()
>  end
>
>  that way downstream packages that haven't adapted yet [like enzyme/reactant] can still run ci with?
>
>  And the same for cuda/metal?
>
> _Originally posted by @wsmoses in https://github.com/JuliaGPU/AMDGPU.jl/issues/961#issuecomment-4961676791_
            

It would probably be easier do the conditional gating conditionally gate here by using checks along the lines of`@static isdefined(GPUCompiler, :codegen)` and shared helper functions if that's something you guys are willing to maintain

src/compiler/relocation.jl
```diff
@@ -0,0 +1,66 @@
+"""
+    relocate_julia_globals!(mod::LLVM.Module, job::GPUCompiler.CompilerJob, meta)
```
maleadt, line 2: Note that I just released and tagged https://github.com/JuliaGPU/GPUCompiler.jl/pull/878, which may change the story here.

kshyatt:
Is there anything others can do to help, here? This is currently blocking anyone using Enzyme from using recent CUDA releases which have bug fixes.

vchuravy:
Replaced by #3512

PR #3309, opened 2026-07-10
Start supporting SPIR-V

vchuravy:
**Start supporting SPIR-V**


wsmoses (approved):
reasonable already, but also presumably needs other changes/tests too?

PR #3335, opened 2026-07-13
Fix deepcopy runtime activity

### LINKED ISSUE #3334: Reverse-mode deepcopy mutates captured array and corrupts repeated gradients
## Summary

Reverse-mode differentiation through `deepcopy` mutates a captured constant array. The first gradient is numerically correct but corrupts the primal array; a second identical gradient consequently returns a wrong value.

This persists on Enzyme 0.13.182 after #3146 and appears closely related to #2868. I could not reopen #2868 because this account lacks repository admin rights, so I am filing the current Enzyme-only reproducer separately.

## MWE

```julia
import Enzyme

const U0 = [2.0]

function loss(p)
    copied = deepcopy((u0 = U0, p = p))
    return (copied.p[1] * copied.u0[1])^2
end

p = [2.0]
mode = Enzyme.set_runtime_activity(Enzyme.Reverse)

u0_before = copy(U0)
first_gradient = only(Enzyme.gradient(mode, loss, p))
u0_after_first = copy(U0)
second_gradient = only(Enzyme.gradient(mode, loss, p))
u0_after_second = copy(U0)

@show first_gradient
@show second_gradient
@show u0_before
@show u0_after_first
@show u0_after_second
```

## Observed output

This reproduces identically on Julia 1.11.9 and Julia 1.12.6 with Enzyme 0.13.182 / Enzyme_jll 0.0.282:

```text
first_gradient = [16.0]
second_gradient = [1296.0]
u0_before = [2.0]
u0_after_first = [18.0]
u0_after_second = [162.0]
```

The expected gradient is `16.0`. The first reverse pass returns that correct gradient but mutates `U0`, even though the differentiated function only reads it through a `deepcopy`. The second pass then differentiates corrupted state.

As a control, removing `deepcopy` gives `([16.0],)` on both repeated gradient calls and leaves `U0 == [2.0]`.

## Environment

The reproducer project has only Enzyme as a direct dependency, pinned to 0.13.182. It was tested with:

- Julia 1.11.9
- Julia 1.12.6
- Enzyme 0.13.182
- Enzyme_jll 0.0.282

## Context

This initially surfaced through a SciML ensemble solve and looked like a serial/threading discrepancy. Reversing the call order showed that whichever gradient ran second was wrong, leading to the SciML-free reduction above.

Related deepcopy issue: #2429.

wsmoses:
fixes https://github.com/EnzymeAD/Enzyme.jl/issues/3334

### DIFF
--- a/src/compiler/reflection.jl
+++ b/src/compiler/reflection.jl
@@ -99,6 +99,7 @@ function enzyme_code_llvm(
     debuginfo::Symbol = :default,
     dump_module::Bool = false,
     mode = API.DEM_ReverseModeCombined,
+    kwargs...
 )
     if mode == API.DEM_ForwardMode
         if A <: Active
@@ -111,7 +112,7 @@ function enzyme_code_llvm(
         end
     end
     JuliaContext() do ctx
-        entry_fn, ir = reflect(func, A, types; optimize, run_enzyme, second_stage, mode)
+        entry_fn, ir = reflect(func, A, types; optimize, run_enzyme, second_stage, mode, kwargs...)
         ts_mod = ThreadSafeModule(ir)
         GC.@preserve ts_mod entry_fn begin
             value = Ref(jl_llvmf_dump(ts_mod.ref, entry_fn.ref))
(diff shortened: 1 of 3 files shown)


wsmoses:
@vchuravy the metal tests are failing for an unrelated gpucompiler compat reason?

PR #3341, opened 2026-07-14
Rewrite cfunction ABI-converter dispatch to go through jl_apply_generic

### LINKED ISSUE #3284: Julia 1.12: segfault from @cfunction ABI mismatch (gcstack_arg) when custom rules execute nested-compiled host code
## Summary

When a custom rule executes host code that was compiled into the differentiated module by `nested_codegen!` (`src/rules/customrules.jl:1704`), any `@cfunction` in that code produces a trampoline whose ABI disagrees with what Julia 1.12's runtime cfunction converter hands back, leading to a segfault. Found while debugging the long-red CUDA CI "Reverse Kernel" testset; this is the next failure after the jl-inst-simplify fix (EnzymeAD/Enzyme#2919).

## Reproducer

Julia 1.12.6, Enzyme main + EnzymeAD/Enzyme#2919 (via `deps/build_local.jl`), CUDA v6:

```julia
using CUDA, Enzyme

function square_kernel!(x)
    i = threadIdx().x
    x[i] *= x[i]
    return nothing
end

function square!(x)
    @cuda blocks = 1 threads = length(x) square_kernel!(x)
    return nothing
end

A = CuArray(collect(Float32, 1:64))
dA = CUDA.ones(64)
Enzyme.autodiff(Reverse, square!, Duplicated(A, dA))  # SIGSEGV
```

CUDACore's `augmented_primal` calls `Enzyme.tape_type` at runtime; the nested compile's `optimize!` invokes LLVM.jl custom passes, and the process crashes inside `module_callback` (LLVM.jl `newpm.jl:140`). The identical `tape_type` call works at top level — only the copy of `run!` compiled into the differentiated module misbehaves.

## Root cause

1. `nested_codegen!` compiles LLVM.jl's `run!` into the module using GPUCompiler's `CodegenParams(gcstack_arg = false)` (GPUCompiler `jlgen.jl:791`), so specsig functions there do **not** take pgcstack as an argument.
2. Julia 1.12 lowers the `@cfunction(module_callback, ...)` inside `run!` to the abi-converter pattern (`gen_cfun_wrapper`): a `jlcapi_module_callback_*` trampoline, cache slots, and a `jl_get_abi_converter` guard. Under `gcstack_arg=false` the trampoline's indirect converter call passes only `(ptr, ptr)` — no pgcstack (verified in the post-AD module dump; cfuncdata flags have specsig=1).
3. At runtime `jl_get_abi_converter` (`runtime_ccall.cpp:445-449`) finds `module_callback`'s code instance compiled by the **stock JIT** (`gcstack_arg=true`, pgcstack in swiftself/r13), sees an exact declrt/sigt match, and returns the **raw specsig pointer**.
4. The trampoline calls it without pgcstack. gdb at the fault: `rdi` = LLVM Module, `rsi` = thunk, and `r13` = leftover thunk pointer — the callee treats r13 as pgcstack, pushes a GC frame into the `CustomPassState` object, and segfaults.

Not Enzyme-specific in principle: any GPUCompiler-compiled host code that *executes* an `@cfunction` on 1.12 has this mismatch. Enzyme is the main consumer that actually runs nested-compiled host code, because rules pull their whole call graph (here: the entire Enzyme.jl compiler plus LLVM.jl's pass machinery — the post-AD module grows from 647 to 3615 functions) into the differentiated module.

## Ruled out

- GC rooting inside rule-executed code is sound (WeakRef + `GC.@preserve` probe inside a custom rule passes).
- The registered callback pointer in the emitted IR is the correct in-module trampoline; the bad pointer comes from the runtime converter at world-guard refresh.

## Fix options

- (a) **Julia**: encode `gcstack_arg` in the cfuncdata flags so `jl_get_abi_converter` refuses the raw-specsig fast path when the caller's ABI differs from the native JIT's. The runtime currently assumes all callers share the stock ABI.
- (b) GPUCompiler `gcstack_arg=true` for host-executable targets: **tested, not viable** — Enzyme.jl fails its own precompilation with "AssertionError: Swiftself attribute coming from differentiable context is not supported" (`lower_convention`).
- (c) **Enzyme.jl**: stop compiling rule-reachable compiler code into the differentiated module and call rules through a dynamic-dispatch boundary. This would also remove the exposure that made the jl-inst-simplify miscompile (#2919) reachable: unsound-but-latent pass behavior currently gets applied to the whole Enzyme/GPUCompiler/LLVM.jl compiler stack on every rule compile.

Happy to file the corresponding JuliaLang/julia issue for (a) if that is the preferred direction.

vchuravy:
## Summary

Fixes #3284 (Julia 1.12 segfault from `@cfunction` ABI mismatch when custom rules execute nested-compiled host code; also the cause of the SciML 1.12 CI segfaults).

Julia 1.12+ lowers `@cfunction` to a world-age-guarded dispatch site that calls `jl_get_abi_converter` to resolve the target in the current world. That runtime resolver hands back pointers with the **native JIT's ABI**: GPUCompiler compiles with `gcstack_arg = false`, while the native JIT publishes specsig pointers that expect pgcstack in the swiftself register (r13/x20). When the runtime returns a raw specsig pointer (exact sigt/declrt match, e.g. after the target was warmed up by the stock JIT), the GPUCompiler-emitted trampoline calls it without setting the register and the callee reads a garbage GC stack.

## Fix

`rewrite_abi_converter_calls!` (called from `prepare_llvm`, so it covers both the primal module and every `nested_codegen!` module) replaces each call to `(i)jl_get_abi_converter` with the in-module `unspecialized` dispatcher thunk that Julia's own codegen already emitted into the cfuncdata global for every site. That thunk boxes the arguments and calls `jl_apply_generic` via the ABI-stable boxed calling convention, and the thunk itself carries the module's own ABI, so no native-ABI pointer ever reaches the indirect call. (Julia's AOT thunk generation cannot pre-bind the in-module target itself because it filters code instances on `ci->owner`, and GPUCompiler-owned CIs are excluded — the fptr slot is prefilled with this same unspecialized thunk, but the world-age guard immediately replaces it via the runtime converter on first call.)

Note that, like the runtime converter before it, this still resolves the callee through the native dispatch cache and method table — it does not route the call through Enzyme's own code cache; it only removes the ABI mismatch.

Handles both lowerings:
- 1.12: `jl_get_abi_converter(ct, fptr, last_world, data)`, six-slot cfuncdata, `unspecialized` in slot 3
- 1.13+: `jl_get_abi_converter(ct, data)`, eight-slot cfuncdata, `unspecialized` in slot 5

## Tests

- `test/runtime_calls.jl`: end-to-end reproducer from #3284 — a custom rule whose `augmented_primal` executes an `@cfunction` callback (runs on all Julia versions).
- `test/passes.jl`: FileCheck tests of `rewrite_abi_converter_calls!` against hand-written IR for both the 1.12 and the 1.13+ dispatch-site patterns (gated to 1.12+, since older LLVM cannot parse opaque-pointer IR).

## Verification

- The (register-clobber hardened) reproducer from #3284 segfaults inside `my_callback` on current main and returns `((1.0,),)` with this PR (Julia 1.12.6).
- `Pkg.test(test_args=["passes", "runtime_calls"])` passes on Julia 1.10, 1.11, and 1.12.


src/llvm/transforms.jl
```diff
@@ -297,6 +297,61 @@ function fixup_1p12_sret!(f::LLVM.Function)
...
+raw specsig pointer handed back is called with a mismatched ABI and reads a
+garbage GC stack out of the swiftself register (#3284).
+
+Rewrite each such site to act like `jl_apply_generic` instead: Julia's codegen
+already emitted an in-module `unspecialized` dispatcher thunk for every site
+(stored in the cfuncdata global) that boxes the arguments and dispatches via
+`jl_apply_generic`. That thunk was compiled with this module's own ABI and
+resolves the callee from the correct cache in the current world, so replace
+every `jl_get_abi_converter` call with it and let the world-age guard fold
+away.
```
vchuravy, line 318: This reduces performance a bit, since we pessimize each call after the world-age call. We should probably call something like `enzyme_get_abi_converter` and do the extra work, but this seems to work for now.

wsmoses: alternatively, can we ourselves call nested_codegen ?

vchuravy: What would that get us? The issue here is that we have a cfunction call that might have become invalid, (e.g. world change). So we dont know the world yet that we would be compiling for, so we need to go through a runtime call, which we do here. 

I think this code is still performing a context escape, since we go an run the unspecialized dispatcher thunk, and I am not sure that this is handling using the right cache.

vchuravy: This gets us back into the 1.11 world, but I am pretty sure we are not yet able to differentiate through something that uses a cfunction and this is essentially a blackbox

wsmoses: so the only cases with this in practice rn are within custom rules, so it doesn't matter if we can AD it -- so long as we can still execute it as intended/fast

src/llvm/transforms.jl
```diff
@@ -297,6 +297,61 @@ function fixup_1p12_sret!(f::LLVM.Function)
...
+            # struct cfuncdata_t: the unspecialized thunk is the 3rd field on 1.12
+            # ([plast_codeinst, last_codeinst, unspecialized, declrt, sigt, flags]) and
+            # the 5th on 1.13+ ([fptr, last_world, plast_codeinst, last_codeinst,
+            # unspecialized, declrt, sigt, flags]).
+            slot = nslots == 6 ? 3 : nslots == 8 ? 5 : error("Enzyme internal error: unknown cfuncdata layout with $nslots slots in $(string(data))")
+            unspec = operands(init)[slot]
+            if !isa(unspec, LLVM.Function)
+                error("Enzyme internal error: cfuncdata of $fname has no unspecialized dispatcher in $(string(data))")
+            end
+            replace_uses!(ci, unspec)
```
wsmoses, line 348: I don't understand where the jl_apply_generic comes from here? this just grabs an operand?

wsmoses: ah is the other cptr is supposed to be the unspecialized version?

wsmoses (approved):

wsmoses:
cuda 1.12 now passes!

wsmoses:
@vchuravy this does not, however, resolve the sciml segfault [which was seemingly resolved by the 1.12 julia patch?]

https://github.com/SciML/SciMLSensitivity.jl/pull/1490

PR #3343, opened 2026-07-14
Debug windows 1.12 threads

wsmoses:
Refactor tests in runtests.jl to use @testset for better organization and clarity. Added various thread-related tests and improved autodiff functionality.


wsmoses:
@vchuravy dumb but could i get a quick assist in making the ci test with two threads

vchuravy:
Eh the reorg of runtests.jl makes that more annoying, you need to launch a new Julia process with the right flags, as an example see how we used to do it before the ParallelTestRunner change, or we are currently doing this with PRT. 

Alternativly I believe `Pkg.test` has `julia_args` as a keyword argument (or may even honor JULIA_CPU_THREADS`)

wsmoses:
@vchuravy is there anything else special about paralleltestrunner as distinct from regular ci, here this now pases, but within PTR it didnt?

vchuravy:
The crash is fixed, but there is a potential flakiness issue

```
Error in testset Advanced, Active-var Threads 2:
Test Failed at D:\a\Enzyme.jl\Enzyme.jl\test\threads.jl:80
  Expression: (res[1])[2] ≈ 2.0
   Evaluated: 1.0 ≈ 2.0
ERROR: LoadError: Test run finished with errors
```

which doesnt happen locally for me.

wsmoses:
re-using this pr to look into atomics since already setup

PR #3359, opened 2026-07-18
Handle nowrite shadows

wsmoses:


wsmoses:
Something is obviously busted here but because of the api change and me wanting to be able to actually check ci before cutting another jll Im going to intentionally merge while red

PR #3363, opened 2026-07-19
Add CUDA specific copy rules

### LINKED ISSUE #3235: Reverse-mode over a CuArray aborts (`gc-transition` on `cuMemcpyDtoHAsync_v2`)
On ParallelMCMC I have been trying to use true Second Order AD for hvps. I've reduced my issues via two MWREs and it seems like Reverse Mode AD on the GPU isn't working properly. 

MWE 1
---
First i tried the native HVP.

```julia
using Enzyme, CUDA
@assert CUDA.functional()

logp(β) = -0.5f0 * sum(abs2, β)      

β = CUDA.CuArray(Float32.(1:8) ./ 8)
v = CUDA.CuArray(ones(Float32, 8))
Enzyme.hvp(logp, β, v)                 
```

This aborts.

MWE 2
---
So then, I was curious if maybe the reverse part of the HVP calculation was broken so i tried this:

```julia
using Enzyme, CUDA
@assert CUDA.functional()

logp(β) = -0.5f0 * sum(abs2, β) 

β = CUDA.CuArray(Float32.(1:8) ./ 8)
Enzyme.gradient(Enzyme.Reverse, logp, β)
```

This also aborts and interestingly in the same way as example 1.

Error
---
```
unsupported tag gc-transition for   %13 = call i32 @cuMemcpyDtoHAsync_v2(
  ptr %"dst::Ptr", i64 %"src::CuPtr", i64 4, ptr %.unbox)
  [ "jl_roots"(...), "gc-transition"() ]
UNREACHABLE executed at .../Enzyme/GradientUtils.cpp:309!

signal (6): Aborted
getInvertedBundles            at .../Enzyme/GradientUtils.cpp:309
recursivelyHandleSubfunction  at .../Enzyme/AdjointGenerator.h:5636
...
```

So seems related to my previous issue #3195 (gc-transition tag) and also #2139 (cuMemcpyDtoHAsync_v2).

Version Info
---
- Julia 1.12.6
- Enzyme.jl 0.13.153, EnzymeCore 0.8.20
- CUDA.jl 5.11.2 (CUDA runtime 13.0)
- GPUCompiler 1.16.1, LLVM.jl 9.8.2
- NVIDIA L40S

rsenne:
@wsmoses added these copy rules per your suggestion and my understanding. Here is what it now allows that prior would break:

- `Array(x)` / `collect(x)` pulling a `CuArray `back to host mid-function. before this it'd just die with "no augmented/forward pass found for cuMemcpyDtoHAsync_v2". e.g., both fwd and rev `sum(abs2, Array(x))` gives back 2x like you'd want.
- `CuArray(y)` / `copyto!(gpu, host)`
- copy(x::CuArray)
- round-trips: Array(CuArray(y)) and back

however, this did not fix #3235 and need further fixes for `sum(abs2, CuArray(x))`

```
EnzymeNoDerivativeError:
No create nofree of empty function (ijl_eqtable_put) ijl_eqtable_put)
 at context:   call fastcc void @julia_task_local_storage_86635(i32 zeroext 0) ... (julia_task_local_storage_86635)

Stacktrace:
 [1] macro expansion
    @ GPUArraysCore/src/GPUArraysCore.jl:205   
 [2] #_mapreduce#48
    @ GPUArrays/src/host/mapreduce.jl:88         
```

`dot` also fails

```
# reverse
EnzymeNoDerivativeError:
No augmented forward pass found for ejlstr$cuMemcpyDtoHAsync_v2$nvcuda
 at context:   %17 = call i32 @"ejlstr$cuMemcpyDtoHAsync_v2$nvcuda"(
     ptr %"dst::Ptr", i64 %"src::CuPtr", i64 4, ptr %.unbox)
     [ "jl_roots"(...), "gc-transition"() ]

# forward
EnzymeNoDerivativeError:
No forward mode derivative found for ejlstr$cuMemcpyDtoHAsync_v2$nvcuda
 at context:   %17 = call i32 @"ejlstr$cuMemcpyDtoHAsync_v2$nvcuda"( ... i64 4 ... )
```

I figured i wouldn't add anything beyond these copy rules unless you think its in scope for this PR otherwise i can continue to target this


ext/EnzymeGPUArraysCoreExt.jl
```diff
@@ -35,4 +36,114 @@ end
...
+
+const _COPY_DIRS = (
+    (:(AbstractGPUArray), :(Array)),           # host -> device
+    (:(Array), :(AbstractGPUArray)),           # device -> host
+    (:(AbstractGPUArray), :(AbstractGPUArray)), # device -> device
+)
+
+for (DT, ST) in _COPY_DIRS
+    @eval begin
+        function EnzymeRules.forward(
```
wsmoses: let's not do the forward copy rule, just the reverse one

wsmoses: otherwise lgtm, add tests though!

wsmoses:
the dot case implies that this is defined on the wrong method. For now don't worry about gpuarrays.

Define rules for https://github.com/JuliaGPU/CUDA.jl/blob/cac0ac97adcd43c23abf0b3c8bb3f99364d9b590/CUDACore/lib/cudadrv/memory.jl#L421

and

https://github.com/JuliaGPU/CUDA.jl/blob/cac0ac97adcd43c23abf0b3c8bb3f99364d9b590/CUDACore/lib/cudadrv/memory.jl#L432

and

https://github.com/JuliaGPU/CUDA.jl/blob/cac0ac97adcd43c23abf0b3c8bb3f99364d9b590/CUDACore/lib/cudadrv/memory.jl#L450

and similar unsafe_copyto down

rsenne:
> Define rules

So IIUC the current methods I defined rules for are a level too high. The current extension I wrote these in does not have CUDA as a dep --- should I add it as a weakdep, or move this PR to `CUDA.jl` and add it in the` EnzymeCoreExt`?

Also, assuming the pointer-level rules subsume it, should I drop the array-level rule I have here or keep it as a fallback?

wsmoses:
Yeah basically.

Re where they live either is a reasonable choice -- mostly one of whose methods/internals seem more used. In the case of the current cuda kernel launch rules those require a lot of the cuda internals so make sense to live there. 

These don't really have any internals so could live anywhere imo.

A relevant fyi, current cuda.jl main doesn't pkg resolve with enzyme atm due to the GPUCompiler 2 bump, though also ci there isn't running regardless and @kshyatt  was looking into that .

Maybe makes sense to have here then. But I'm fine either way

rsenne:
@wsmoses i think this is ready

Project.toml
```diff
@@ -41,6 +43,7 @@ ADTypes = "1"
 BFloat16s = "0.2, 0.3, 0.4, 0.5, 0.6"
 CEnum = "0.4, 0.5"
 ChainRulesCore = "1"
+CUDA = "6"
```
wsmoses: explicitly we should not require cuda 6

ext/EnzymeCUDAExt.jl
```diff
@@ -0,0 +1,256 @@
...
+        dsrc .+= tmp
+    end
+    fill!(ddest, zero(eltype(ddest)))
+    _commit!(src, soff, dsrc, n)
+    _commit!(dest, doff, ddest, n)
+    return nothing
+end
+
+function _zero!(dest, doff, n)
+    ddest = _stage(dest, doff, n)
```
wsmoses: we shouldn't require this.

if dest is a host pointer{<:AbstractFloat} we can use memset(dest + doff, sizeof(eltype) * n)

if a device pointer, we can use function cudaMemset

ext/EnzymeCUDAExt.jl
```diff
@@ -0,0 +1,256 @@
...
+    return nothing
+end
+
+function _accumulate_and_zero!(src, soff, dest, doff, n::Integer)
+    n == 0 && return nothing
+    typeof(src) === typeof(dest) && src == dest && soff == doff && return nothing
+
+    dsrc = _stage(src, soff, n)
+    ddest = _stage(dest, doff, n)
+    if dsrc isa Array && ddest isa Array || dsrc isa CuArray && ddest isa CuArray
```
wsmoses: we should never have the case where src and dst are host mem here

ext/EnzymeCUDAExt.jl
```diff
@@ -0,0 +1,256 @@
...
+        dsrc .+= ddest
```
wsmoses: in the device/device case, we should just write our own kernel here rather than wrapping to a cuarray and broadcasting.

Kernel should be easy [and you can use KA too if you like]

ext/EnzymeCUDAExt.jl
```diff
@@ -0,0 +1,256 @@
...
+    fill!(ddest, zero(eltype(ddest)))
+    _commit!(dest, doff, ddest, n)
+    return nothing
+end
+
+@inline function _shadow(x, config, batch)
+    return EnzymeRules.width(config) == 1 ? x.dval : x.dval[batch]
+end
+
+function EnzymeRules.augmented_primal(
```
wsmoses: we shouldn't need this?

rsenne: I think this is needed---I tried removing it and all the copy-rule gradients come back zero. Without it the reverse `unsafe_copyto!` rules accumulate into an unconnected pointer because unlike say `pointer(::Array)`, `pointer(::CuArray)` isn't an intrinsic Enzyme shadows (at least that is how i understand it)

wsmoses: Er that is a bug then, can you upload an isolated as mwe possible case with enzyme.api.printall!(true) on

ext/EnzymeCUDAExt.jl
```diff
@@ -0,0 +1,284 @@
...
+function _zero!(ptr::CuArrayPtr{T}, off::Integer, n::Integer) where {T}
+    n == 0 && return nothing
+    buffer = _stage(ptr, off, n)
+    fill!(buffer, zero(eltype(buffer)))
+    _commit!(ptr, off, buffer, n)
+    return nothing
+end
+
+# Device kernel for `dst[i] += src[i]`; avoids materializing a broadcast.
+function _addto_kernel!(dst, src, n)
```
kshyatt: Maybe a dumb question but if you've already got CUDA.jl as a dep, why not use the CUBLAS axpy here? It might well be faster than a naive kernel

test/cuda.jl
```diff
@@ -1,7 +1,164 @@
 using CUDA
 using Enzyme
+using LinearAlgebra: dot
 using Test
 
+@testset "CUDA memory copies" begin
```
kshyatt, line 6: Would be good to check this works with all the CUDA memory types (Device, Host, Unified)

ext/EnzymeCUDAExt.jl
wsmoses: oh right, we absolutely should do that here

vchuravy:
Same comment as https://github.com/EnzymeAD/Enzyme.jl/pull/3375#issuecomment-5045719824 the CUDA API ones should live in EnzymeAD/Enzyme?

rsenne:
@wsmoses here is the MWE.

```julia
using CUDA, Enzyme

f(x) = sum(abs2, Array(CuArray(x)))

x  = Float32[1, 2, 3]
dx = zeros(Float32, 3)

get(ENV, "PRINTALL", "0") == "1" && Enzyme.API.printall!(true)
Enzyme.autodiff(Reverse, f, Active, Duplicated(x, dx))
Enzyme.API.printall!(false)

println("dx       = ", dx)
println("expected = ", 2 .* x)
println("correct? = ", dx == 2 .* x)
```

For me, with `x = Float32[1,2,3]`, keeping the `pointer(::StridedCuArray)` rule gives `dx = [2,4,6]`  and removing it gives `dx = [3,6,9]`.

ext/EnzymeCUDAExt.jl
```diff
@@ -0,0 +1,274 @@
...
+function _zero!(ptr::CuPtr{T}, off::Integer, n::Integer) where {T <: AbstractFloat}
+    n == 0 && return nothing
+    bytes = reinterpret(CuPtr{UInt8}, ptr + off * sizeof(T))
+    CUDA.memset(bytes, UInt8(0), n * sizeof(T))
+    return nothing
+end
+function _zero!(ptr::CuArrayPtr{T}, off::Integer, n::Integer) where {T}
+    n == 0 && return nothing
+    buffer = _stage(ptr, off, n)
+    fill!(buffer, zero(eltype(buffer)))
```
wsmoses: this can call the same zero as CuPtr{T}, right?

ext/EnzymeCUDAExt.jl
```diff
@@ -0,0 +1,274 @@
...
+@inline _stage(ptr::Union{Ptr, CuPtr}, offset::Integer, n::Integer) =
+    _asarray(ptr + offset, n)
+function _stage(ptr::CuArrayPtr{T}, offset::Integer, n::Integer) where {T}
+    buffer = CuArray{T}(undef, n)
+    Base.unsafe_copyto!(pointer(buffer), ptr, offset, n)
+    return buffer
+end
+
+@inline _commit!(::Union{Ptr, CuPtr}, offset, buffer, n) = nothing
+function _commit!(ptr::CuArrayPtr, offset, buffer, n)
```
wsmoses: what is this whole commit thing and why is it needed?

ext/EnzymeCUDAExt.jl
```diff
@@ -0,0 +1,274 @@
...
+    axpy!(one(eltype(dst)), src, dst)
+    return nothing
+end
+_accumulate!(dst::Array, src::Array) = (dst .+= src; nothing)
+
+function _accumulate_and_zero!(src, soff, dest, doff, n::Integer)
+    n == 0 && return nothing
+    typeof(src) === typeof(dest) && src == dest && soff == doff && return nothing
+
+    dsrc = _stage(src, soff, n)
```
wsmoses: get rid of the stage/commit stuff

ext/EnzymeCUDAExt.jl
```diff
@@ -0,0 +1,274 @@
...
+    return nothing
+end
+
+function _accumulate!(dst::CuArray, src::CuArray)
+    length(dst) == 0 && return nothing
+    # `dst .+= src` via cuBLAS; covers Float16/32/64 and complex shadows.
+    axpy!(one(eltype(dst)), src, dst)
+    return nothing
+end
+_accumulate!(dst::Array, src::Array) = (dst .+= src; nothing)
```
wsmoses: we should benchmark if we should still use axpy here, I would hope it to be better than broadcasting?

ext/EnzymeCUDAExt.jl
```diff
@@ -0,0 +1,274 @@
...
+
+function _accumulate_and_zero!(src, soff, dest, doff, n::Integer)
+    n == 0 && return nothing
+    typeof(src) === typeof(dest) && src == dest && soff == doff && return nothing
+
+    dsrc = _stage(src, soff, n)
+    ddest = _stage(dest, doff, n)
+    if dsrc isa CuArray && ddest isa CuArray
+        _accumulate!(dsrc, ddest)
```
wsmoses: just do this inside overrides of `_accumulate!`

ext/EnzymeCUDAExt.jl
```diff
@@ -0,0 +1,274 @@
...
+    length(dst) == 0 && return nothing
+    # `dst .+= src` via cuBLAS; covers Float16/32/64 and complex shadows.
+    axpy!(one(eltype(dst)), src, dst)
+    return nothing
+end
+_accumulate!(dst::Array, src::Array) = (dst .+= src; nothing)
+
+function _accumulate_and_zero!(src, soff, dest, doff, n::Integer)
+    n == 0 && return nothing
+    typeof(src) === typeof(dest) && src == dest && soff == doff && return nothing
```
wsmoses: why would this ever be true?

test/cuda.jl
rsenne: Added device / unified / host coverage. Device passes; unified and host are `@test_broken` for now. The blocker is the `pointer(::StridedCuArray)` rule: for unified/host memory `pointer(::CuArray)`. This is related to the other issue i am trying to figure out so leaving it as a flag for now

ext/EnzymeCUDAExt.jl
rsenne: i do not think so as `memset` is defined only for `CuPtr{T}`

wsmoses: i would assume a convert or something exists given https://github.com/JuliaGPU/CUDA.jl/blob/8fc49feda7cf563fa9fb4900e329dfecc99ed0fd/CUDACore/src/array.jl#L848

rsenne: hmm i tried the following:

```julia
using CUDA
mem = CUDA.alloc(CUDA.ArrayMemory{Float32}, (4,))

convert(CuPtr{Float32}, mem) # ERROR: ArgumentError: Illegal conversion of a ArrayMemory{Float32, 1} to a CuPtr{Float32}

Base.unsafe_convert(CuPtr{Float32}, mem) # ERROR: ArgumentError: Illegal conversion of a ArrayMemory{Float32, 1} to a CuPtr{Float32}

CUDA.memset(reinterpret(CuPtr{UInt8}, pointer(mem)), UInt8(0), 16) # ERROR: CUDA error: invalid argument (code 1, ERROR_INVALID_VALUE)
```

If i understand the code snippet linked, `pointer(A)` is already a` CuPtr`, because A is a `DenseCuArray`. So `convert(CuPtr{U}, pointer(A)) `is a `CuPtr` -> `CuPtr` retype

wsmoses: this just feels like a missing memset impl in cuda.jl [and we should open an issue for]?

```
### Array memory

Array memory is a special kind of memory that is optimized for 2D and 3D access patterns. The memory is opaquely managed by the CUDA runtime, and is typically only used on combination with texture intrinsics.

``@docs
CUDA.ArrayMemory
CUDA.alloc(::Type{CUDA.ArrayMemory{T}}, ::Dims) where T
``

### Pointers

To work with these buffers, you need to `convert` them to a `Ptr`, `CuPtr`, or in the case
of `ArrayMemory` an `CuArrayPtr`. You can then use common Julia methods on these pointers,
such as `unsafe_copyto!`. CUDA.jl also provides some specialized functionality that does not
match standard Julia functionality:

``@docs
CUDA.unsafe_copy2d!
CUDA.unsafe_copy3d!
CUDA.memset
``
```

In any case let's make our lives easier then and just leave the method unimplemented for CuArrayPtr for now

ext/EnzymeCUDAExt.jl
rsenne:
```
eltype    n           broadcast(us)   axpy(us)        speedup(bcast/axpy)
Float32   1           7.896           13.734          0.57
Float32   16          8.115           13.123          0.62
Float32   256         8.668           13.615          0.64
Float32   4096        8.096           13.318          0.61
Float32   65536       8.483           13.476          0.63
Float32   1048576     8.266           13.597          0.61
Float32   16777216    96.727          95.242          1.02
Float64   1           9.326           13.458          0.69
Float64   16          9.092           13.423          0.68
Float64   256         9.014           14.109          0.64
Float64   4096        8.579           13.794          0.62
Float64   65536       8.256           14.423          0.57
Float64   1048576     9.157           13.786          0.66
Float64   16777216    256.104         257.62          0.99
```

apparently not ?

wsmoses: s a v a g e.

but sure, add a comment that we could do axpy [called like X], but it is faster [copying this table] to use a bcast

rsenne:
i think i got everything except for the putative bug

ext/EnzymeCUDAExt.jl
wsmoses: sorry, i meant, lets actually delete it [and delete commit/asarray/etc]

ext/EnzymeCUDAExt.jl
wsmoses: bump here, can we remove

ext/EnzymeCUDAExt.jl
wsmoses: bump

ext/EnzymeCUDAExt.jl
wsmoses: bump

rsenne:
I think you mayhaps looked at an old commit :)

wsmoses (commented):
okay looks good to me minus the pointer rule. Can you comment it out and upload the log from enzyme.api.printall!(true) set to investigate?

rsenne:
[enzyme.log](https://github.com/user-attachments/files/30315745/enzyme.log)

@wsmoses

wsmoses:
what was the julia code to generate that [to match with log]

rsenne:
```julia
using CUDA, Enzyme

f(x) = sum(abs2, Array(CuArray(x)))

x  = Float32[1, 2, 3]
dx = zeros(Float32, 3)

open("enzyme.log", "w") do io
    redirect_stderr(io) do
        Enzyme.API.printall!(true)

        Enzyme.autodiff(
            Reverse,
            f,
            Active,
            Duplicated(x, dx),
        )

        Enzyme.API.printall!(false)
    end
end
```

with `augmented_primal` commented out

wsmoses:
can you reduce the test case to just the call julia side which causes an error, its too complex otherwise

rsenne:
sorry got caught up--i'm minimizing the MWE now

rsenne:
hows this?
 
 ```julia
 using CUDA, Enzyme

f(x) = @inbounds Array(x)[1]

x  = CuArray(Float32[1, 2, 3])
dx = CUDA.zeros(Float32, 3)

open("enzyme.log", "w") do io
    redirect_stderr(io) do
        Enzyme.API.printall!(true)

        Enzyme.autodiff(
            Reverse,
            f,
            Active,
            Duplicated(x, dx),
        )

        Enzyme.API.printall!(false)
    end
end
 ```
[enzyme.log](https://github.com/user-attachments/files/30322679/enzyme.log)

wsmoses:
made an alternate solution, which seems to work [will merge/tag but reopen if still an issue]

rsenne:
great thanks!

wsmoses:
@vchuravy somehow this seems to be a source of gc nondetermisim on ci, if you see anything obvious here

Issue #3237, opened 2026-06-26
LaTeX not being rendered correctly for math line of documentation in Box example


sai-kalai:
This line
https://github.com/EnzymeAD/Enzyme.jl/blob/b5886809ac64e9dc5baba03a9039f175ad78c9cc/examples/box.jl#L360
is not rendered in Chrome or Firefox, and inspecting the code in Overleaf I see that there is use of `\right|` without a corresponding left call. I assume that this is intended as math notation for "restricted to". Looking at some search results, this requires a `\left.\ to match. 
https://tex.stackexchange.com/questions/22252/how-to-typeset-function-restrictions 
Could someone please confirm that the rendering is broken? Or does this work for some LaTex interpreter? 
Thanks!

<img width="1290" height="357" alt="Image" src="https://github.com/user-attachments/assets/e4b040ba-7a5d-4e26-a640-e8827f2aeed8" />

sai-kalai:
Modified code now reads
```
\overline{\text{state_now}} = \left.\frac{\partial \text{out_now}}{\partial \text{state_now}}\right|_\text{state_now} \overline{\text{out_now}} + \left.\frac{\partial \text{out_old}}{\partial \text{state_now}}\right|_\text{state_now} \overline{\text{out_old}}
```

and renders as 

<img width="933" height="118" alt="Image" src="https://github.com/user-attachments/assets/8cb21bdd-66ef-4922-b3bc-0b97c1e2a3ae" />

Can someone confirm that this is the intended rendering? Thanks! @vchuravy @swilliamson7 @just-walk

wsmoses:
That rendering LGTM

wsmoses:
@sai-kalai was this merged/fixed [and thus can now be closed]?

sai-kalai:
Hi @wsmoses! 
Thanks so much for the followup.
Unfortunately i got caught up with schoolwork and didn't manage to merge this. I will do that asap and tag you.

PR #3366, opened 2026-07-19
docs: fix math rendering in box example

sai-kalai:
The LaTeX rendering was broken in one line of the box example due to incorrect usage of matching `\left.` for every `\right|`

### DIFF
--- a/examples/box.jl
+++ b/examples/box.jl
@@ -356,9 +356,9 @@ autodiff(Reverse,
 # `state_now`, and acted with this gradient on what we gave as `dout_now` (in our case,
 # all ones). Using AD notation for reverse mode, this is
 
-# ```math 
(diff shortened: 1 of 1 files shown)


sai-kalai:
Solves #3237 @wsmoses

sai-kalai:
I am unsure if this will pass the CI pipeline, because when I try to build the docs myself by running 
```bash
julia --project=docs docs/make.jl -- doctest=false  
```
from the project's root, I get an error. I'm unsure what I need to do to be able to build the docs.

```
 Error: failed to run `@example` block in docs/src/generated/box.md:586-588
│ ```@example box
│ @show abs.(diffs .- dstate_old[5])./dstate_old[5]
│ ```
│   exception =
│    UndefVarError: `diffs` not defined in `Main.__atexample__named__box`
│    Suggestion: check for spelling errors or missing imports.
```

wsmoses:
we should fix that too [cc @swilliamson7 ], though we can get this merged first

Issue #2837, opened 2025-12-07
autodiff failure with Flux and CUDA


CarloLucibello:
```julia
using CUDA, cuDNN
using Flux
using Enzyme

model = Dense(2 => 2) |> gpu
dmodel = Duplicated(model, Enzyme.make_zero(model))
ad = Enzyme.set_runtime_activity(Reverse)
f(m, x) = sum(m(x))
x = rand(Float32, 2) |> gpu
Enzyme.autodiff(ad, Const(f), Active, dmodel, Const(x))
```
```
ERROR: EnzymeNoDerivativeError:
No create nofree of empty function (jl_gc_safe_enter) jl_gc_safe_enter)
 at context:   call fastcc void @julia_setindex__22298({} addrspace(10)* noundef nonnull readonly align 8 dereferenceable(8) %46, float noundef 0.000000e+00) #703, !dbg !1715 (julia_setindex__22298)

Stacktrace:
  [1] CuRefValue
    @ ~/.julia/packages/CUDA/x8d2s/src/refpointer.jl:55
  [2] CuRefValue
    @ ~/.julia/packages/CUDA/x8d2s/src/refpointer.jl:58
  [3] CuRef
    @ ~/.julia/packages/CUDA/x8d2s/src/refpointer.jl:28
  [4] convert
    @ ~/.julia/packages/CUDA/x8d2s/src/refpointer.jl:30
  [5] cconvert
    @ ./essentials.jl:543
  [6] macro expansion
    @ ~/.julia/packages/GPUToolbox/JLBB1/src/ccalls.jl:153
  [7] macro expansion
    @ ~/.julia/packages/CUDA/x8d2s/lib/cublas/libcublas.jl:1879
  [8] #426
    @ ~/.julia/packages/GPUToolbox/JLBB1/src/ccalls.jl:34
  [9] retry_reclaim
    @ ~/.julia/packages/CUDA/x8d2s/src/memory.jl:434
 [10] check
    @ ~/.julia/packages/CUDA/x8d2s/lib/cublas/libcublas.jl:22
 [11] cublasSgemv_v2_64
    @ ~/.julia/packages/GPUToolbox/JLBB1/src/ccalls.jl:33
 [12] gemv!
    @ ~/.julia/packages/CUDA/x8d2s/lib/cublas/wrappers.jl:530

Failure within method:

[...]

PR #3375, opened 2026-07-21
Add mul! (gemv/gemm) rule for GPU arrays (#2837)

CarloLucibello:
update after #3363 landed:

Fixes the remaining gradient in #2837 — the copy rules landed separately in #3363.

Reverse-mode `mul!` on GPU arrays returned a **zero matrix gradient** (and on newer `Enzyme_jll` an internal `cmpxchg` error): the backend BLAS call (`cublasSgemv` etc.) is a foreign ccall Enzyme can't differentiate through. Enzyme core differentiates CPU BLAS fine, just not the GPU path.

- **`mul!` forward + reverse rules** implementing the analytic gemm/gemv adjoint (`Ā += α·C̄·Bᵀ`, `B̄ += α·Aᵀ·C̄`, plus α/β scalar and batch handling), added to `EnzymeGPUArraysCoreExt`.
- **Dispatched on `AnyGPUArray`**, so it covers all GPU backends and wrapped operands (`transpose`/`adjoint`/`view`), while excluding CPU `Array` — leaving Enzyme core's CPU-BLAS path untouched. A plain shadow accumulates via in-place `mul!`; a wrapped shadow (e.g. `transpose(dW)`, which isn't a valid BLAS output) via broadcast.

Verified on an RTX 5090: gemv, gemm, a wrapped-transpose case, the issue MWE, and forward mode all match analytic gradients (`test/cuda.jl`).

### DIFF
--- a/ext/EnzymeGPUArraysCoreExt.jl
+++ b/ext/EnzymeGPUArraysCoreExt.jl
@@ -1,7 +1,10 @@
 module EnzymeGPUArraysCoreExt
 
 using GPUArraysCore
+using GPUArraysCore: AnyGPUArray
 using Enzyme
+using Enzyme: EnzymeRules
+using LinearAlgebra: mul!
 
 function Enzyme.zerosetfn(x::AbstractGPUArray, i::Int)
     res = zero(x)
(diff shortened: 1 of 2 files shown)


CarloLucibello:
Should this stuff live in the enzyme extensions in CUDA.jl instead?
https://github.com/JuliaGPU/CUDA.jl/blob/cac0ac97adcd43c23abf0b3c8bb3f99364d9b590/lib/cublas/ext/EnzymeCoreExt.j
https://github.com/JuliaGPU/CUDA.jl/blob/cac0ac97adcd43c23abf0b3c8bb3f99364d9b590/CUDACore/ext/EnzymeCoreExt.jl

vchuravy:
I think the common CUDA API derivatives belong into EnzymeAD/Enzyme and that also has support for cuBLAS (but maybe missing the function you want to use)?

CarloLucibello:
Thanks @vchuravy — I dug into the Enzyme core cuBLAS support and it looks like the story is a bit more specific than I expected. Sharing what I found in case it changes where these belong:

**Core already differentiates the cuBLAS gemv correctly.** With *only* a rule for `unsafe_copyto!` (the device-pointer memcpy) and **no `mul!` rule at all**, `sum(W*x)` on `CuArray`s gives the correct `dW`. `extractBLAS` already lists `gemv`/`gemm` and matches the `cublas` prefix with the `_v2`/`_64` suffixes, and the Julia `ejlstr$cublasSgemv_v2$…` name is cleaned up via the `enzyme_math` attribute. So the gemv/gemm derivative in this PR is largely redundant with core.

**The actual blocker is the CUDA async memcpy**, not cuBLAS: cuBLAS stages the gemv `alpha`/`beta` scalars through a `CuRefValue` whose `setindex!` does a host→device `unsafe_copyto!` → `cuMemcpyHtoDAsync_v2`. That ccall (with a `gc-transition` bundle) has no augmented-forward pass, which is the original `No augmented forward pass found for cuMemcpyHtoDAsync_v2` error. Once it's handled, core takes over the gemv.

So the piece that could move into EnzymeAD/Enzyme is the **`cuMemcpy{HtoD,DtoH,DtoD}Async_v2` handling** (near the `memcpy`/`memmove` path in `CallDerivatives.cpp`). One wrinkle: CUDA device pointers are `CUdeviceptr` (`i64`) in the IR rather than typed pointers, so `visitMemTransferCommon` would need to shadow the device side correctly — which is partly why the Julia-side `EnzymeRules` version (with real `CuArray`/`CuPtr` shadow tracking) is convenient.

Questions:
1. Would you prefer the `cuMemcpy*Async_v2` copy derivative in core, or is the Julia-side `unsafe_copyto!` rule acceptable (given the `i64` device-pointer handling)?
2. If core handles gemv, I'll drop the `mul!` rule here — though note the full `sum(W*x .+ b)` MWE also trips a separate KernelAbstractions broadcast issue (`CuArray{EnzymeTapeToLoad}` tape allocation), which the `mul!` rule currently steers compilation around. Happy to file that separately.

Can share the exact reproducers/IR for any of the above.

wsmoses:
copy ones landed here: https://github.com/EnzymeAD/Enzyme.jl/pull/3363

CarloLucibello:
Updated to reflect the discussion:

- **Dropped the `unsafe_copyto!` rules** — those landed in #3363. Rebased onto current `main`.
- **Kept only the `mul!` gemv/gemm rule**, since Enzyme core can't differentiate the GPU backend BLAS ccall (foreign call → zero matrix gradient, and an internal `cmpxchg` error on the newer jll). My rule provides the analytic adjoint directly and sidesteps that path.
- **Moved it to `EnzymeGPUArraysCoreExt` and dispatched on `AnyGPUArray`** instead of `CuArray`: the rule body is backend-agnostic, so this covers AMDGPU/Metal too and — via `AnyGPUArray` — wrapped operands (`transpose`/`adjoint`/`view`). CPU `Array` is excluded, so your core CPU-BLAS handling is untouched.

Verified on CUDA: gemv, gemm, wrapped-transpose, the #2837 MWE, and forward mode all match analytic gradients.

CarloLucibello:
@wsmoses can this go (once tests pass)?

wsmoses:
@CarloLucibello I think this should be at least partially handled by a [apologies concurrent] PR in https://github.com/EnzymeAD/Enzyme.jl/pull/3148

Can you see if that resolves the issues here, and if not what rules still remain needed

CarloLucibello:
reverse mode is covered in the merged pr more generically than here. the forward mode of mul! (or lower level ops) is still missing

PR #3379, opened 2026-07-21
Fix reverse-mode race on Active scalars captured by Threads.@threads (#3378)

### LINKED ISSUE #3378: Reverse-mode race: parallel Threads.@threads loses Active-scalar gradient contributions
## Summary

Reverse-mode autodiff of a `Threads.@threads` loop that writes an **`Active` scalar** into each element loses gradient contributions under thread contention. The per-thread adjoints of the active variable are not accumulated atomically, so concurrent updates race and the returned gradient is too small.

This is the existing `test/threads.jl` "Advanced, Active-var Threads" testset:

```julia
function f_multi(out, in)
    Threads.@threads for idx in 1:length(out)
        out[idx] = in
    end
    return nothing
end

out  = [1.0, 2.0]
dout = [1.0, 1.0]
res = autodiff(Reverse, f_multi, Const, Duplicated(out, dout), Active(2.0))
@test res[1][2] ≈ 2.0   # each of the 2 iterations contributes dout[idx]=1 -> 2.0
```

Expected `d(in) = dout[1] + dout[2] = 2.0`; under a race one contribution is lost and it returns `1.0`.

## Reproducer (amplified)

The 2-element case only races occasionally, so it is easier to see with a larger loop and more threads. `d(in)` should equal `length(out)`:

```julia
using Enzyme
function f_multi(out, in)
    Threads.@threads for idx in 1:length(out)
        out[idx] = in
    end
    return nothing
end

N = 2000
let out = zeros(N), dout = ones(N)   # warmup
    autodiff(Reverse, f_multi, Const, Duplicated(out, dout), Active(2.0))
end
bad = 0; mn = Inf
for _ in 1:100
    out = zeros(N); dout = ones(N)
    res = autodiff(Reverse, f_multi, Const, Duplicated(out, dout), Active(2.0))
    v = res[1][2]
    mn = min(mn, v)
    v ≈ N || (bad += 1)
end
@show Threads.nthreads(), bad, mn
```

Observed (Windows, Julia 1.12.6, Enzyme_jll 0.0.288):

| `--threads` | bad / 100 | min value (expected 2000) |
|---|---|---|
| 2 | 0 | 2000 |
| 4 | 0 | 2000 |
| 8 | 6–9 | 1750 |

The number of lost updates grows with thread contention, which is the signature of a non-atomic read-modify-write of the active variable's shadow. On a contended 2‑vCPU CI runner the 2-element/2-thread case reproduces directly (returns `1.0`).

## Environment

- Windows 11, x86_64; also expected on other OSes under enough contention
- Julia 1.12.6, Enzyme_jll 0.0.288+0, GPUCompiler 1.23.0, LLVM.jl 9.10.1

## Notes

- This is **independent of the linkage change** in #3377 — it reproduces with and without it and lives in the parallel active-variable reduction, not in JIT linkage. It became visible on Windows CI only because #3377 stops the earlier `threads/2` crash (#3374), so the suite now runs far enough to reach this testset. Previously the Windows worker terminated before it.
- Likely fix area: the per-thread accumulation of an `Active` argument's adjoint in the `runtime_pfor_*` / parallel reverse path ([src/rules/parallelrules.jl](https://github.com/EnzymeAD/Enzyme.jl/blob/main/src/rules/parallelrules.jl)) needs an atomic add (or a per-thread partial + reduction) instead of a plain `+=`.

vchuravy:
Fixes #3378.

## Problem

Reverse-mode autodiff of a `Threads.@threads` loop that writes an **`Active` scalar** loses gradient contributions under thread contention:

```julia
function f_multi(out, in)
    Threads.@threads for idx in 1:length(out)
        out[idx] = in
    end
    return nothing
end
res = autodiff(Reverse, f_multi, Const, Duplicated(out, dout), Active(2.0))
# res[1][2] should be dout[1] + dout[2] = 2.0, but races to 1.0
```

The captured scalar `in` lives in an **immutable** closure whose shadow is a single `Base.RefValue` shared by every thread (`ft = Duplicated{Base.RefValue{closure}}`). Because the field is immutable, each thread's adjoint accumulates via a non-atomic whole-struct read-modify-write of that shared `Ref`, so concurrent updates race and contributions are lost. The array capture (`out`/`dout`) doesn't race — it's a shared mutable pointer accumulated at distinct indices.

## Fix

In `runtime_pfor_rev`, when the shadow is a shared `Ref` closure, give each thread its own shadow closure and reduce afterward:

- **Mutable / inactive captures** (arrays, boxes, ...) are **shared** with the caller's shadow. Their shadow carries the incoming cotangents the reverse pass *reads* (e.g. the loop's `dout`), and their in-place accumulation must reach the caller.
- **Active immutable captures** (floats, immutable structs of floats) are **zeroed** so each thread accumulates into its own partial.
- After the parallel region the partials are summed back into the caller's shadow with `recursive_add(..., mutable_register)` (which leaves the shared mutable data untouched).

## Verification

- Amplified reproducer (N=2000, 8 threads): before **2–9 bad / 100**, min 1750; after **0 bad**, min 2000 (0 bad over 1000 runs).
- `test/threads.jl` passes with 4 and 8 threads; added an amplified check to the "Advanced, Active-var Threads" testset since the 2-element case only raced occasionally.

## Open design question 🙏

`zero_active_shadow` is a *share-mutable / zero-active* traversal that overlaps with the existing `make_zero` / `recursive_add` machinery but is not either of them:

- `make_zero` **zeros** mutable shadows by default — which would read an incoming `dout` as zeros and give a zero gradient. It *can* share them by pre-seeding its `seen::IdDict` with each mutable object mapped to itself, but collecting those objects is the same traversal implemented here.
- The reduction reuses `recursive_add(x, y, identity, mutable_register)`.

So this adds a third variant of the "walk a shadow" pattern. **Should it be unified with `make_zero`/`recursive_add`, and are users meant to be able to extend it for their own types?** Feedback welcome — happy to refactor toward whatever the maintainers prefer.

### DIFF
--- a/src/rules/parallelrules.jl
+++ b/src/rules/parallelrules.jl
@@ -160,7 +160,7 @@ function (st::ReversePFor{ThunkTy, FT, AnyJL, byRef, TT})(tid) where {ThunkTy, F
     else
         @inbounds st.tapes[tid]
     end
-    
+
     if byRef
         st.thunk(Const(referenceCaller), st.ft, Const(tid), tres)
     else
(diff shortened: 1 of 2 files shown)


wsmoses (changes requested):
this is the wrong fix, someone or something isnt doing the atmoic like it should

PR #3380, opened 2026-07-21
1.13: vendor one-dim broadcast indexing

wsmoses:


wsmoses:
@vchuravy @giordano this now makes ci green [minus the lux integration test time out on 1.10 just instantiating for some reason....]

PR #3383, opened 2026-07-22
Add Downgrade CI

vchuravy:
Adds a Downgrade CI workflow that tests the lower bounds of the `[compat]` entries in `Project.toml` actually resolve and pass, using [`julia-actions/julia-downgrade-compat`](https://github.com/julia-actions/julia-downgrade-compat).

- Runs on Julia 1.10 (the minimum supported version) on `ubuntu-24.04`.
- Dev's in `EnzymeCore`/`EnzymeTestUtils` (path deps) and skips `EnzymeCore` from downgrade since it's a local source.
- Triggered on the same path filters as `CI.yml`.


.github/workflows/Downgrade.yml
```diff
@@ -0,0 +1,75 @@
+name: Downgrade
+on:
+  pull_request:
+    paths:
```
giordano: GitHub somewhat recently added support for anchors, so you don't have to duplicate lists
```suggestion
    paths: &paths
```

.github/workflows/Downgrade.yml
```diff
@@ -0,0 +1,75 @@
...
+      - release-*
+    paths:
+      - '.github/workflows/Downgrade.yml'
+      - 'deps/**'
+      - 'ext/**'
+      - 'lib/**'
+      - 'src/**'
+      - 'test/**'
+      - '!test/integration/**'
+      - 'Project.toml'
```
giordano:
```suggestion
    paths: *paths
```

.github/workflows/Downgrade.yml
```diff
@@ -0,0 +1,75 @@
...
+          version: ${{ matrix.version }}
+          arch: ${{ matrix.arch }}
+      - uses: julia-actions/cache@v3
+      - name: add EnzymeCore EnzymeTestUtils
+        shell: julia --color=yes --project=. {0}
+        run: |
+          using Pkg
+          Pkg.develop([PackageSpec(; path) for path in ("lib/EnzymeCore", "lib/EnzymeTestUtils")])
+        env:
+          JULIA_PKG_SERVER_REGISTRY_PREFERENCE: eager
```
giordano, line 53: When are we moving to v1.11+? 😁

giordano (approved):

Project.toml
```diff
@@ -49,7 +49,7 @@ LLVM = "9.10"
 LogExpFunctions = "0.3, 1"
 ObjectFile = "0.4, 0.5"
 PrecompileTools = "1"
-Preferences = "1.4"
+Preferences = "1.5.2"
 SparseArrays = "1"
 SpecialFunctions = "1, 2"
```
giordano: Based on https://github.com/EnzymeAD/Enzyme.jl/actions/runs/29918794945/job/88918850322?pr=3383#step:8:533
```
│ ┌ Error: Error during loading of extension EnzymeSpecialFunctionsExt of Enzyme, use `Base.retry_load_extensions()` to retry.
│ │   exception =
│ │    1-element ExceptionStack:
│ │    InitError: UndefVarError: `_logabsgamma` not defined
│ │    Stacktrace:
│ │      [1] getproperty
│ │        @ ./Base.jl:31 [inlined]
│ │      [2] __init__()
│ │        @ EnzymeSpecialFunctionsExt ~/work/Enzyme.jl/Enzyme.jl/ext/EnzymeSpecialFunctionsExt.jl:7
```
I presume this should be (see https://github.com/JuliaMath/SpecialFunctions.jl/commit/86d1910b1d733f8fc21f42456edc5e7e204d4b56)
```suggestion
SpecialFunctions = "2.2"
```

Project.toml
```diff
@@ -38,7 +38,7 @@ EnzymeStaticArraysExt = "StaticArrays"
 
 [compat]
 ADTypes = "1"
-BFloat16s = "0.2, 0.3, 0.4, 0.5, 0.6"
+BFloat16s = "0.4, 0.5, 0.6"
 CEnum = "0.4, 0.5"
 ChainRulesCore = "1"
```
giordano: https://github.com/EnzymeAD/Enzyme.jl/actions/runs/29918794945/job/88918850322?pr=3383#step:8:740
```
Error in testset test_forward: sort with return activity EnzymeCore.Duplicated on (::Vector{Float64}, EnzymeCore.Duplicated):
Error During Test at /home/runner/work/Enzyme.jl/Enzyme.jl/lib/EnzymeTestUtils/src/test_forward.jl:71
  MethodError: no method matching getindex(::Nothing, ::Int64)
  Stacktrace:
    [1] forward
      @ ~/work/Enzyme.jl/Enzyme.jl/ext/EnzymeChainRulesCoreExt.jl:57 [inlined]
    [2] CallWithKWargs
      @ ~/work/Enzyme.jl/Enzyme.jl/lib/EnzymeTestUtils/src/test_utils.jl:7 [inlined]
    [3] fwddiffejulia_CallWithKWargs_26010wrap
      @ ~/work/Enzyme.jl/Enzyme.jl/lib/EnzymeTestUtils/src/test_utils.jl:0
    [4] macro expansion
      @ ~/work/Enzyme.jl/Enzyme.jl/src/compiler.jl:7003 [inlined]
    [5] enzyme_call
      @ ~/work/Enzyme.jl/Enzyme.jl/src/compiler.jl:6473 [inlined]
    [6] ForwardModeThunk
      @ ~/work/Enzyme.jl/Enzyme.jl/src/compiler.jl:6373 [inlined]
    [7] autodiff
      @ ~/work/Enzyme.jl/Enzyme.jl/src/Enzyme.jl:691 [inlined]
```
Don't quite know the fix here, but presumably we need a somewhat more recent version as lower bound

.github/workflows/Downgrade.yml
vchuravy: For this CI run we could immediately

Project.toml
giordano: After doing some bisection, tests pass locally for me with
```suggestion
ChainRulesCore = "1.13"
```

Project.toml
```diff
@@ -49,9 +49,9 @@ LLVM = "9.10"
 LogExpFunctions = "0.3, 1"
```
giordano:
```suggestion
LogExpFunctions = "0.3.2, 1"
```

Project.toml
giordano, line 1: What's up with this diff? Why it looks like the entire file changed?

giordano: Looks like that when applying a review suggestion (c5ec28968130811449973a38eb7c304a8d64c97c) GitHub decided it was a good idea to change the line ending to CR LF.  I restored LF in e713b8908427e74b8d4c91cca566990a9685a5d0.

giordano: Oh, that already happened in dda4285b4f9ac5e4bde6260ff0e38131c1c4a5bc, but only on line 43 (the `ChainRulesCore` compat bound), but not 54 (the `SpecialFunctions` one) because of course mixing up different line endings in the same file is a totally sensible idea.  WTH GitHub?!?

.github/workflows/Downgrade.yml
```diff
@@ -0,0 +1,67 @@
...
+jobs:
+  test:
+    timeout-minutes: 120
+    name: Downgrade - Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        version:
+          - '1.10'
```
vchuravy:
```suggestion
          - '1'
```

.github/workflows/Downgrade.yml
giordano: This step can be removed then
```suggestion
```

vchuravy: turns out that for Downgrade we also want the lowest version of Julia we support.

giordano: Aaaaah, makes sense

PR #3410, opened 2026-07-28
BigFloat consistency for kwargs

kshyatt:
I think we got hit by this over at https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/actions/runs/29334378385/job/87115065119?pr=257


wsmoses (commented):
we should....add a test for this

kshyatt:
Added

PR #3415, opened 2026-07-29
Easy rule for zero(BigFloat)

kshyatt:


kshyatt:
Fail here doesn't look related?

wsmoses:
so one thing for sure is that this doesn't fail on main

kshyatt:
yikes

wsmoses:
failure here is clearly related

kshyatt:
yeah the easy rule was doing a lot better

kshyatt:
OK that seemed to fix the issue on 1.11. I think this is good to go.

Issue #3429, opened 2026-08-03
`illegal calling convention fixup` strikes again


kshyatt:
Here's the dump: https://gist.github.com/kshyatt/cf9ac2866cf2542657f1384664f6a081 (too big for a GitHub issue)

To repro (sorry this is heavy):

Check out https://github.com/QuantumKitHub/PEPSKit.jl/tree/ksh/enzyme
Run this MWE with `julia --project=test mwe.jl`

```julia
using Test
using Random
using PEPSKit
using TensorKit
using Enzyme 
using OptimKit
using KrylovKit

Enzyme.Compiler.VERBOSE_ERRORS[] = true

## Test models, gradmodes and CTMRG algorithm
# -------------------------------------------
χbond = 2
χenv = 6
Pspaces = [ComplexSpace(2), Vect[FermionParity](0 => 1, 1 => 1)]
Vspaces = [ComplexSpace(χbond), Vect[FermionParity](0 => χbond / 2, 1 => χbond / 2)]
Espaces = [ComplexSpace(χenv), Vect[FermionParity](0 => χenv / 2, 1 => χenv / 2)]
models = [heisenberg_XYZ(InfiniteSquare()), ]
names = ["Heisenberg",]

gradtol = 1.0e-4
ctmrg_verbosity = 0
ctmrg_algs = [[:SimultaneousCTMRG],]
projector_algs = [[:HalfInfiniteProjector,],]
svd_rrule_algs = [[:FullPullback],]
gradient_algs = [[:FixedPointGradient],]
gradient_solver_algs = [
    [:GeomSum,],
]
steps = -0.01:0.005:0.01

# don't check naive AD gradients for all algorithm combinations, since it's slow
naive_gradient_combinations = [
    (:SimultaneousCTMRG,),
]
naive_gradient_done = Set()

# fixed-point differentiation is incompatible with sequential CTMRG
function _check_disallowed_combination(
        ctmrg_alg, projector_alg, decomposition_rrule_alg, gradient_alg
    )
    ctmrg_alg == :SequentialCTMRG && !isnothing(gradient_alg) && return true
    return false
end

@testset "AD CTMRG energy gradients for $(names[i]) model" verbose = true for i in
    eachindex(
        models
    )
    Pspace = Pspaces[i]
    Vspace = Vspaces[i]
    Espace = Espaces[i]
    calgs = ctmrg_algs[i]
    palgs = projector_algs[i]
[...]

PR #3435, opened 2026-08-05
Box tape aggregates that Julia will not store inline

kshyatt:
OK I threw a 🤖 at #3429 and here's the result:

The tape ABI assumes Julia's lowering of the tape type is memory-layout identical to Enzyme's LLVM tape type: the augmented forward stores the tape using Enzyme's layout into a slot typed by the Julia tape type, and the reverse pass reads it back. Julia only keeps a struct-typed field inline when `Base.allocatedinline` holds, and that fails once a type has pointers and any field of 32768 bytes or more -- `jl_fielddesc16_t` has a 15-bit size field, so the layout falls back to `fielddesc_type == 2`, which `jl_datatype_isinlinealloc` rejects. Julia then boxes the field, the two layouts diverge, and the tape can neither be stored nor read back.

Model that explicitly rather than letting the layouts silently disagree:

- `to_tape_type` maps any nested aggregate Julia would not store inline to `Any`, and `julia_conv_fixup` / `calling_conv_fixup` box and unbox it in the ABI wrappers.
- `tape_type(...; boxnested = false)` keeps the old layout-faithful type for the places that size or zero raw tape storage rather than describe a Julia field: `zero_allocation`, `julia_allocator`, and the anonymous-tape paths.
- `raw_storage_type` builds a flat NamedTuple with explicit `NTuple{n, UInt8}` padding so every leaf keeps its original byte offset, for allocations whose nested type cannot be laid out faithfully. Pointer leaves stay `Any` so the GC still traces them, and it returns `nothing` unless Julia agrees on both the total size and every field offset.
- The anonymous tape is passed as an object pointer and loaded in the wrapper instead of being loaded as a by-value aggregate out of an addrspace(10) pointer.

`ENZYME_FORCE_ANONYMOUS_TAPE=1` forces heap-allocated tapes on for everything, which is the only way to reach that path from Julia in tests.

### DIFF
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -2014,7 +2014,8 @@ function zero_allocation(B::LLVM.API.LLVMBuilderRef, LLVMType::LLVM.API.LLVMType
     B = LLVM.IRBuilder(B)
     LLVMType = LLVM.LLVMType(LLVMType)
     obj = LLVM.Value(obj)
-    jlType = Compiler.tape_type(LLVMType)
+    # raw tape storage: needs the layout-faithful type, not the boxed one
+    jlType = Compiler.tape_type(LLVMType; boxnested = false)
     zeroAll = isTape == 0
     func = LLVM.parent(position(B))
     mod = LLVM.parent(func)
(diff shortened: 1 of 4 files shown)


wsmoses:
this is super complex and thus I'm not sure if the right fix -- can you at minimum have claude add a mwe test that fails without these changes

kshyatt:
Yeah can do, I have a reproducer it boiled down for me in the issue so I'll try to get it to make a test. Sorry this guy is so large.

kshyatt:
Here's one at least. I can move it into another file if you prefer.

Issue #3392, opened 2026-07-25
Very large compilation burden (time & memory) for `test_reverse`/`autodiff_thunk` with complicated types


kshyatt:
OK, this one kind of *does* need heavy dependencies, sorry! It's also kind of hard to repro without this specific branch which I realize sucks :(

The problem I'm having is that the Enzyme tests for some linear algebra factorizations at https://github.com/QuantumKitHub/TensorKit.jl/pull/464 are OOMing on (GitHub free tier) CI, although they pass no problem locally on my M4 MacBook Air. I've included a "MWE" below you can run with the branch of that PR checked out if interested. The calls to `GC.gc` and `empty_globalcaches` are attempts to lower the memory pressure, mostly without success. If I run this locally on my laptop, I see:

```
Trivial svd_trunc: 37.560453 seconds (215.69 M allocations: 10.541 GiB, 3.56% gc time, 99.79% compilation time)
Trivial lq_compact: 15.272431 seconds (106.74 M allocations: 4.694 GiB, 4.47% gc time, 99.77% compilation time)
Trivial qr_compact: 8.192286 seconds (63.99 M allocations: 2.558 GiB, 4.91% gc time, 99.58% compilation time)
Test Summary:                                 | Pass  Total     Time
Enzyme - Factorizations: Trivial (ComplexF64) |  245    245  1m08.5s
Trivial svd_trunc: 6.035741 seconds (43.80 M allocations: 2.221 GiB, 2.54% gc time, 96.20% compilation time)
Trivial lq_compact: 7.930535 seconds (62.44 M allocations: 2.795 GiB, 2.65% gc time, 96.93% compilation time)
Trivial qr_compact: 7.081363 seconds (57.10 M allocations: 2.527 GiB, 2.79% gc time, 96.62% compilation time)
Test Summary:                                 | Pass  Total   Time
Enzyme - Factorizations: Trivial (ComplexF64) |  272    272  22.6s
Irrep[U₁] svd_trunc: 16.507090 seconds (97.34 M allocations: 4.629 GiB, 2.16% gc time, 98.63% compilation time)
Irrep[U₁] lq_compact: 13.432332 seconds (89.11 M allocations: 3.708 GiB, 2.17% gc time, 98.58% compilation time)
Irrep[U₁] qr_compact: 13.312716 seconds (86.75 M allocations: 3.578 GiB, 3.73% gc time, 96.88% compilation time)
Test Summary:                                   | Pass  Total   Time
[...]

PR #3436, opened 2026-08-05
Emit only the entry from nested_codegen!

kshyatt:
More 🤖 stuff, this time actually addressing the `only_entry` stuff called out in #3392:

nested_codegen! built its private module with the whole transitive callee graph of each rule, once per rule per thunk. For the qr_compact! pullback in #3392 that is 464 function definitions / 216,995 LLVM instructions, of which the pullback itself is ~2%. check_ir, set_module_types! and optimize! all walk that module, and the JIT then retains it, so the graph is paid for four times over.

Hand jl_emit_native only the entry CodeInstance instead. Julia's resolve_workqueue already covers this case ("make sure all referenced methods get fixed up, particularly if the user declined to compile them"): every dangling call site gets an emit_tojlinvoke thunk, plus an emit_specsig_to_fptr1 adapter where the site used the specialized ABI. The callees then dispatch into Julia's own globally cached native code rather than being re-emitted here, which is ABI-correct by construction and compiles each callee once for the whole process.

This is sound only because these modules are never differentiated (run_enzyme = false) and are linked in after enzyme! has run, so Enzyme never needs to see through the calls that become thunks.

Note that GPUCompiler's own only_entry flag does not work for this: it empties non-entry bodies after InternalizePass has made them internal, yielding internal-linkage declarations, and it emits the full graph first so it saves no irgen time. It is only used from reflection.jl, where the module is printed rather than JIT'd.

The flag is task-local so it cannot leak into onehot_internal, which shares PrimalCompilerParams but consumes its module as an llvmcall string and so needs its callees emitted. Gated to Julia 1.12+, where jl_emit_native takes an explicit codeinfos list; 1.10 and 1.11 are unchanged.

On a custom rule over A*B with sum/norm/tr and mul! in the pullback, the reverse rule's module drops from 52 defs / 28,185 instructions to 19 / 2,288, and a steady-state thunk build from 0.75s to 0.35s. The cost is that rule callees become boxed jl_invoke dispatches: +22% on 4x4 matrices where dispatch dominates, neutral by 64x64. Interpreter.OnlyEntry[] = false restores the old behaviour.

### DIFF
--- a/src/compiler/interpreter.jl
+++ b/src/compiler/interpreter.jl
@@ -1506,4 +1506,96 @@ function abstract_call_known(
     )
 end
 
+"""
+    HAS_CODEINFO_LIST
+
+Whether `jl_emit_native` takes an explicit list of `CodeInstance => CodeInfo` pairs
+describing what to emit (as opposed to being handed a `MethodInstance` and walking the
+call graph itself). Only on such versions can `nested_codegen!` ask for the entry alone
+and have Julia stub out the callees; see `OnlyEntryKey`.
+"""
+const HAS_CODEINFO_LIST = VERSION >= v"1.12.0-DEV.1823"
+
+"""
+    OnlyEntryKey
+
+Task-local-storage key guarding the "emit the entry only" mode used by
+`Enzyme.Compiler.nested_codegen!`.
+
+When this flag is set, `GPUCompiler.ci_cache_populate` hands `jl_emit_native`
+only the entry `CodeInstance` instead of the whole transitively reachable callee graph.
+Julia's `resolve_workqueue` then patches up every call site that was left dangling by
+emitting a `jl_invoke` thunk (plus a specsig-to-fptr1 adapter where the call site uses the
+specialized ABI), so the callees dispatch into Julia's own natively compiled code rather
+than being re-emitted into Enzyme's private module.
+
+This matters because `nested_codegen!` is called once per rule per thunk, and the module it
+builds is checked, type-annotated, optimized and then retained by the JIT. Emitting the
+entry alone keeps that module small in all four respects. It is only correct because these
+modules are never differentiated (`run_enzyme = false`) and are linked in after `enzyme!`
+has already run, so Enzyme never needs to see through the calls that become thunks.
+
+The flag is task-local rather than global because it must not leak into unrelated jobs that
+share the interpreter — notably `Enzyme.onehot_internal`, which uses the same
+`PrimalCompilerParams` but consumes the emitted module as an `llvmcall` string and so needs
+its callees emitted.
+"""
+const OnlyEntryKey = gensym(:enzyme_only_entry)
+
+"""
+    OnlyEntry
+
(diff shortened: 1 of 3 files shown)


kshyatt:
This does seem to have fixed my issue with the factorizations tests that sparked all this

src/compiler/interpreter.jl
```diff
@@ -1506,4 +1506,96 @@ function abstract_call_known(
...
+            push!(codeinfos, ci => src)
+        end
+        return codeinfos
+    end
+
+    function GPUCompiler.ci_cache_populate(
+            @nospecialize(interp::EnzymeInterpreter), cache, mi::MethodInstance,
+            min_world::UInt, max_world::UInt
+        )
+        if only_entry_requested()
```
wsmoses: this is hacky, just make it a param of CompilerConfig rather than a global?

src/compiler/interpreter.jl
```diff
@@ -1506,4 +1506,96 @@ function abstract_call_known(
...
 
+"""
+    HAS_CODEINFO_LIST
+
+Whether `jl_emit_native` takes an explicit list of `CodeInstance => CodeInfo` pairs
+describing what to emit (as opposed to being handed a `MethodInstance` and walking the
+call graph itself). Only on such versions can `nested_codegen!` ask for the entry alone
+and have Julia stub out the callees; see `OnlyEntryKey`.
+"""
+const HAS_CODEINFO_LIST = VERSION >= v"1.12.0-DEV.1823"
```
wsmoses, line 1529: this only works 1.12, how do you fix for 1.10/1.11?

kshyatt: Apparently we can't fix this on earlier versions because of GPUCompiler? Apparently [this](https://github.com/JuliaGPU/GPUCompiler.jl/blob/3c19d2fb778bb5b9e58570cc73622c2c8fb2a536/src/jlgen.jl#L573) is the culprit but I gotta confess this is tooooo deep in compiler world for me

vchuravy:
I believe this is the right direction, but we should consider being more aggressive and actually just emit an invoke for the rule.

src/compiler/interpreter.jl
```diff
@@ -1506,4 +1518,81 @@ function abstract_call_known(
...
+            @nospecialize(interp::EnzymeInterpreter), cache, mi::MethodInstance,
+            min_world::UInt, max_world::UInt
+        )
+        if interp.only_entry
+            return entry_codeinfo(interp, mi)
+        end
+        return Base.@invoke GPUCompiler.ci_cache_populate(
+            interp::Any, cache::Any, mi::Any, min_world::Any, max_world::Any
+        )
+    end
```
vchuravy, line 1594: This change makes me nervous.

vchuravy:
Should be fixed by #3479

@kshyatt you might have to use `@noinline` to annotate your rules if you still see blow up, but at least some testing shows that it is now fixed.

PR #3105, opened 2026-05-15
Add some CUDA-specific tests for EnzymeTestUtils

kshyatt:
I'm really the dog-with-chemistry-set meme here but tried to actually test some of this and see if it works, it seems there is some stuff that isn't quite functioning?

PR #3437, opened 2026-08-05
Fix complex copies of GPU arrays for Enzyme (split from #3105)

kshyatt:
Fixes some type signatures on rules and adds tests to make sure they're hit now.


kshyatt:
I was able to hit these lines with some tests from MatrixAlgebraKit, trying to see if I can boil that down enough to make a decent test set for `test/cuda.jl`

gbaraldi:
The test comments look a bit odd to me but LGTM

gbaraldi (approved):

kshyatt:
I can strip them out if you like

PR #3358, opened 2026-07-18
Fix vararg warning

wsmoses:
attempting to resolve

```
Wrapping Vararg directly in UnionAll is deprecated (wrap the tuple instead).
```

PR #3447, opened 2026-08-10
Fix Vararg-in-UnionAll deprecation in absint

CarloLucibello:
When abstractly interpreting `Core.apply_type(Vararg, X)` with a partial (statically unknown) argument, `absint` introduced a fresh `TypeVar` and wrapped the result via `UnionAll(u, res)`. Because `res` is a bare `Vararg` (not a `Tuple`), this produced the deprecated `Vararg{T} where T` and emitted:

```
Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
  ... absint(arg::LLVM.Value, ...) @ Enzyme.Compiler src/absint.jl
```

during downstream test runs (reported from Flux's Enzyme tests).

**Fix:** detect the bare-`Vararg` case and widen each partial typevar to its upper bound instead of wrapping in a `UnionAll`. This is semantically identical — `UnionAll(u, Vararg{u})` for an unbounded `u` already evaluated to `Vararg{Any}` — while avoiding the deprecated code path. The other `UnionAll` in this file always wraps a `Tuple`, so it is unaffected.

Verified in isolation under `julia --depwarn=error`: the old path raises exactly this deprecation, the new path returns `Vararg{Any}` cleanly.

Fixes the Enzyme-side deprecation reported in FluxML/Flux.jl#2570.

### DIFF
--- a/src/absint.jl
+++ b/src/absint.jl
@@ -139,8 +139,18 @@ function absint(@nospecialize(arg::LLVM.Value), partial::Bool = false, istracked
 
                 if legal
                     res = Ty{found...}
-                    for u in unionalls
-                        res = UnionAll(u, res)
+                    if res isa Core.TypeofVararg
+                        # A bare `Vararg` cannot be wrapped in a `UnionAll`
+                        # (deprecated: `Vararg{T} where T`). Widen each partial
+                        # typevar to its upper bound instead.
+                        if !isempty(unionalls)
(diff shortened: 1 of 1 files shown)


wsmoses:
dup of https://github.com/EnzymeAD/Enzyme.jl/pull/3358/changes

PR #3461, opened 2026-08-15
Add rules for BigFloat integer constructors

kshyatt:
`BigFloat(x::Clong)` and `BigFloat(x::Culong)` lower to `mpfr_set_si` / `mpfr_set_ui`, for which there is no derivative, so any active use of an integer-constructed BigFloat failed with `EnzymeNoDerivativeError`.

The argument is an integer, so the result never carries a derivative. This cannot be expressed as `inactive`, though: the result is a fresh mutable BigFloat that subsequent MPFR calls may write into, so it still needs a shadow to exist -- the same reason the existing no-argument `BigFloat()` rule allocates one. The shadow is built through the same constructor with a zero value, so it lands at the requested precision as a properly initialised MPFR value rather than raw zeroed memory.

Having a rule here also keeps Enzyme from compiling the body of these constructors at all, which matters beyond the missing derivative: `eps(::Type{BigFloat})` is `nextfloat(BigFloat(1)) - BigFloat(1)`, and compiling that construction inside an Enzyme-generated function miscompiles it into a BigFloat with an invalid `d::Memory`, segfaulting on first use. This repro segfaults before the change and passes after:

    f(out) = (out[1] = eps(BigFloat) > 0 ? 1.0 : 0.0; nothing)
    Enzyme.autodiff(Forward, Const(f), Const(zeros(3)))

Developed with 🤖

### DIFF
--- a/src/internal_rules/bigfloat.jl
+++ b/src/internal_rules/bigfloat.jl
@@ -67,6 +67,87 @@ function EnzymeRules.reverse(
     return ()
 end
 
+# `BigFloat(x::Clong)` / `BigFloat(x::Culong)` lower to `mpfr_set_si` / `mpfr_set_ui`,
+# for which there is no derivative, so any active use of an integer-constructed
+# BigFloat otherwise fails with `EnzymeNoDerivativeError`.
+function EnzymeRules.forward(
+    config::EnzymeRules.FwdConfig,
+    Ty::Const{Type{BigFloat}},
+    RT::Type{<:Union{DuplicatedNoNeed,Duplicated,BatchDuplicated,BatchDuplicatedNoNeed}},
+    x::Const{Ti},
+    rs::Const{Base.MPFR.MPFRRoundingMode}...;
+    kwargs...,
+    ) where {Ti <: Union{Clong, Culong}}
+    rvals = map(r -> r.val, rs)
+    if EnzymeRules.needs_primal(config) && EnzymeRules.needs_shadow(config)
+        if EnzymeRules.width(config) == 1
+            return remove_innerty(RT)(
+                Ty.val(x.val, rvals...; kwargs...),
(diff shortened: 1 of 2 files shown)


src/internal_rules/bigfloat.jl
```diff
@@ -67,6 +67,87 @@ function EnzymeRules.reverse(
...
+    config::EnzymeRules.RevConfig,
+    Ty::Const{Type{BigFloat}},
+    RT::Type{<:Union{DuplicatedNoNeed,Duplicated,BatchDuplicated,BatchDuplicatedNoNeed}},
+    tape,
+    x::Const{<:Union{Culong, Clong}},
+    rs::Const{Base.MPFR.MPFRRoundingMode}...;
+    kwargs...,
+)
+    # the integer argument carries no derivative
+    return ntuple(i -> nothing, Val(1 + length(rs)))
```
wsmoses, line 148: change to returns?

src/internal_rules/bigfloat.jl
```diff
@@ -67,6 +67,87 @@ function EnzymeRules.reverse(
...
+
+function EnzymeRules.augmented_primal(
+    config::EnzymeRules.RevConfig,
+    Ty::Const{Type{BigFloat}},
+    RT::Type{<:Union{DuplicatedNoNeed,Duplicated,BatchDuplicated,BatchDuplicatedNoNeed}},
+    x::Const{Ti},
+    rs::Const{Base.MPFR.MPFRRoundingMode}...;
+    kwargs...,
+    ) where {Ti <: Union{Clong, Culong}}
+    rvals = map(r -> r.val, rs)
```
wsmoses, line 119: closures are bad

vchuravy: > closures are bad

Citation needed?

PR #3463, opened 2026-08-15
Let Bigfloat be handled by to_vec as well for testing

kshyatt:
I'm hitting this with tests of arrays of `BigFloat`s


wsmoses (approved):

kshyatt:
The 1.13 fail is present on main so I will merge

Issue #3322, opened 2026-07-11
Nested differentiation through a custom rule whose body calls a function returning a mixed pointer/isbits struct: CallingConventionMismatchError in classify_arguments (Julia <= 1.11)


ChrisRackauckas-Claude:
Under nested (forward-over-forward) differentiation, when a custom rule's body calls a non-inlined function that returns an immutable struct with both pointer and isbits fields, Enzyme's module type-classification fails while compiling the OUTER differentiation:

```
ERROR: LoadError: CallingConventionMismatchError: Enzyme hit an internal error trying to parse the julia calling convention definition for:

makecache(::Int64, ::Int64)
     @ Main ~/claude_tmp/claude-25022/-home-crackauc-sandbox-tmp-20260710-063942-23305/7bf73f89-ee49-42ce-86e5-b5335e721373/scratchpad/mwe4_callconv.jl:26

Hint: catch this exception as `err` and call `code_typed(err)` to inspect the errornous code.

expectLen != length(parameters(f))
void ({ {} addrspace(10)*, {} addrspace(10)*, i8 }*, [2 x {} addrspace(10)*]*, i64, i64)
expectLen=3
has_swiftself=false
has_sret=true
has_returnroots=true
source_sig=Tuple{typeof(makecache), Int64, Int64}
parmsRemoved=UInt64[0x0000000000000001]


Stacktrace:
  [1] classify_arguments(source_sig::Type, codegen_ft::LLVM.FunctionType, has_sret::Bool, has_returnroots::Bool, has_swiftself::Bool, parmsRemoved::Vector{UInt64}, mi::Core.MethodInstance, world::UInt64)
    @ Enzyme.Compiler ~/.julia/packages/Enzyme/CCx0o/src/typeutils/jltypes.jl:339
  [2] set_module_types!(interp::Enzyme.Compiler.Interpreter.EnzymeInterpreter{Nothing}, mod::LLVM.Module, primalf::LLVM.Function, job::GPUCompiler.CompilerJob{Enzyme.Compiler.EnzymeTarget{GPUCompiler.NativeCompilerTarget}, Enzyme.Compiler.EnzymeCompilerParams{Enzyme.Compiler.PrimalCompilerParams}}, edges::Vector{Any}, run_enzyme::Bool, mode::Enzyme.API.CDerivativeMode)
    @ Enzyme.Compiler ~/.julia/packages/Enzyme/CCx0o/src/compiler.jl:1163
[...]

Issue #2513, opened 2025-08-16
Calling convention mismatch with imported chainrule


gdalle:
Hi there!
Following up on https://github.com/JuliaDecisionFocusedLearning/ImplicitDifferentiation.jl/issues/35, here's an MWE with a boiled down version of ImplicitDifferentiation. The Zygote call works, which proves that the rrule is ok, but the Enzyme prints a "calling convention mismatch" error. It's not exactly the same error as the original issue (not sure why) but maybe it is related? It seems to originate from the same `src/rules/customrules.jl` file anyway.

```julia
using ChainRulesCore
using Enzyme
using ForwardDiff
using Zygote

## Mini version of ImplicitDifferentiation.jl

struct ImplicitFunction{S,C}
    solver::S
    conditions::C
end

(implicit::ImplicitFunction)(x) = implicit.solver(x)

function ChainRulesCore.rrule(implicit::ImplicitFunction, x::Array)
    (; solver, conditions) = implicit
    y = solver(x)
    A = ForwardDiff.jacobian(_y -> conditions(x, _y), y)
    B = ForwardDiff.jacobian(_x -> conditions(_x, y), x)
    J = -A \ B
    function implicit_pullback(dy)
        dx = transpose(J) * dy
        return NoTangent(), dx
    end
    return y, implicit_pullback
end

## Implicit square root solver

struct CustomError{C} <: Exception
    content::C
end

function solver(x)
    y = sqrt.(x)
    # make Enzyme error to check it doesn't differentiate through the solver
    try
        throw(CustomError(y))
    catch e
        if e isa CustomError
            return e.content
        end
    end
end

conditions(x, y) = y .^ 2 .- x

## Test

x0 = [2.0, 3.0]
implicit = ImplicitFunction(solver, conditions)
y0 = implicit(x0)

ForwardDiff.jacobian(solver, x0)  # works

Zygote.jacobian(solver, x0)  # errors
Zygote.jacobian(implicit, x0)  # works

Enzyme.@import_rrule typeof(implicit) Vector{Float64}
Enzyme.jacobian(Enzyme.set_runtime_activity(Enzyme.Reverse), solver, x0)  # returns zero
Enzyme.jacobian(Enzyme.set_runtime_activity(Enzyme.Reverse), implicit, x0)  # still returns zero and prints an error
```

Error warning:

```julia
[...]

PR #3479, opened 2026-08-24
Call custom rules through their natively compiled CodeInstance

vchuravy:
Alternative to #3436. Stacked on #3480 (its commit is the first one here): the swiftself branch of the rule handlers has to work for this.

#3436 keeps nested-codegen'ing the rule entry into the calling module and stubs out its *callees* as `jl_invoke` thunks. This PR instead lets the rule's inlining annotation decide how the differentiated code reaches it, exactly as for a regular call (`rule_call_convention`):

- a rule declared `@inline` has its LLVM IR emitted into the calling module (`nested_codegen!`) and is always-inlined there;
- a rule declared `@noinline` is compiled once, natively, and called directly through its specialized entry point;
- an unannotated rule follows Julia's own inlining heuristics for that method: inlined if Julia would inline it, called otherwise.

A rule whose result is a compile-time constant is emitted inline whatever its annotation: Julia compiles no code for it.

### The call ABI

On Julia 1.12+, `invoke_codegen!` (replacing `nested_codegen!` at the three rule call sites) infers and compiles the rule with Julia's **native** interpreter and JIT (`typeinf_ext_toplevel` with `SOURCE_MODE_ABI`, see `rule_codeinst`), exactly as an ordinary call of the rule would. The rule therefore shares its `CodeInstance` and native code with ordinary callers and is compiled at most once per process. `jl_read_codeinst_invoke` gives the specialized entry point, which is declared in the calling module with the signature Julia's codegen gives it (`get_specsig_function`: `[sret][return roots][pgcstack swiftself] args...`, swift calling convention) and bound through the existing `enzymejl_needs_restoration` mechanism. No boxing, no dispatch, and no copy of the rule body walked by `check_ir`/`set_module_types!`/`optimize!` per calling thunk.

The declaration is shaped exactly like the codegen'd rule function used to be (parameters, `sret`/`swiftself` attributes, `enzymejl_mi`/`enzymejl_rt`), so `customrules.jl` keeps building arguments and reading the return convention as before. `rule_specsig` derives it by the same rules as Julia's codegen: arguments are classified as ghost, boxed (`deserves_argbox`, i.e. anything but a singleton or a concrete immutable Julia can allocate inline; a struct with uninitialized fields is boxed although `jl_type_to_llvm` gives it a struct type), by reference (aggregates, plus an inline-roots pointer on 1.12+), or by value (with `zeroext`/`signext` on small integers, copied to the call sites since only they survive the pointer restoration). A `Union` of ghosts returns its selector byte. The rule handlers take the tape type and the reverse rule's return type from the same native inference (`native_rule_return_type`), so the types match the code they call.

**Why the native interpreter.** `EnzymeInterpreter` folds `within_autodiff()` to `true` and leaves Enzyme intrinsics such as `ignore_derivatives` for the Enzyme pipeline to lower; Julia's JIT cannot resolve those, so a natively compiled rule inferred by it crashed at link time. It also re-inferred the rule's whole callee graph under Enzyme's cache owner. Native inference removes both. The two interpreters never mix: native code enters the Enzyme module only as an address-bound declaration, never as IR. Consequently `within_autodiff()` is `false` inside a natively called rule and `ignore_derivatives` is the identity there, as in any other Julia code; inside an emitted rule it stays `true`. The tutorial documents this.

A rule Julia compiles without a specialized entry point, or for a MethodInstance other than the one looked up, is an error rather than a fallback. Neither can happen for a valid rule: every rule takes an annotation, an immutable struct Julia passes unboxed, so Julia always gives it a specialized signature. Nested codegen remains the implementation where the call ABI does not exist: Julia < 1.12, device modules, precompilation.

A pointer-free tape is now passed to the reverse rule from a stack slot rather than a heap box; the box only ever went away because the rule got inlined.

### Nested differentiation

A derivative that is differentiated again (forward-over-reverse and friends) must differentiate through the rules it calls, and a natively called rule is an opaque declaration to the outer pass. The pointer restoration now leaves those declarations symbolic until `_thunk` compiles the module, so a deferred job linked into the outer primal module, and a derivative embedded by `enzyme_call`, still carry them. `materialize_native_rules!` then emits each such rule with `nested_codegen!` and defines the declaration as an always-inline wrapper before the outer differentiation runs. Covered for forward-over-forward, reverse-over-forward and forward-over-reverse in `test/rules/rule_inlining.jl`.

### `invoke(f, ci::CodeInstance, args...)` (third commit)

Julia 1.12 inference understands `invoke(f, ci, args...)` (`InvokeCICallInfo`) and 1.13's inliner lowers it to a direct `:invoke` of that CodeInstance. Neither goes through `abstract_call_gf_by_type`, so such a call never got Enzyme's rule marking, and the `:invoke` targets the caller's CodeInstance rather than the one Enzyme inferred for the signature — its code was emitted a second time without the `enzymejl_mi` attributes, and got differentiated through instead of having its rule applied. When the invoked signature is a primitive or has an inactive/forward/reverse rule, Enzyme now analyzes the call as the ordinary `f(args...)` through `abstract_call`, so it devirtualizes to Enzyme's own CodeInstance with the usual `NoInlineCallInfo` marking and the rule applies, for known callees as well as callable structs and closures. Gated to 1.13, where the inliner lowers it (on 1.12 such calls stay runtime `invoke` calls and fail as before). Covered by `test/rules/invoke_ci.jl`.

### Why no shim / no jl_invoke thunk

An earlier iteration compiled a `shim(f, args...) = @noinline f(args...)` entry-only and let `resolve_workqueue` turn the rule `:invoke` into its specsig→`jl_invoke` adapter. That adapter heap-boxes every non-ghost argument and the return: 2 allocations per forward-rule call and 6 per reverse-rule pair, ~5× slower than the inlined rule on scalar rules. Binding the specsig directly to the compiled `specptr` removes all of that: on a scalar forward rule and an array reverse rule with tape (5M calls each, rules forced through the call ABI) the native call is allocation-free at ~11 ns / ~7.5 ns per call, against ~4 ns with the rule body inlined — the residual is the out-of-line call itself, which is why small rules stay inlined by default.

### Effect on #3392

Measured on Julia 1.12.6 with the TensorKit factorization rules from that issue, one `autodiff_thunk` for `qr_compact` on a trivial-sector `TensorMap{ComplexF64}`: the first build goes from 65–71 s / 9.0 GiB on main to 45–46 s / 6.0 GiB, and every further thunk that reuses the rule from 12–16 s and +70–140 MiB RSS to 0.8 s and +1 MiB. The issue's whole "Trivial (ComplexF64)" testset goes from 158–168 s to about 115 s with a lower peak RSS. The remaining first-build time is the differentiation of the outer TensorKit function itself, which this PR does not touch.

### Docs and tests

The custom rules tutorial gains a "How rules are compiled" section describing the annotations, the semantics of `within_autodiff()` in each kind of rule, and nested differentiation. `test/rules/rule_inlining.jl` covers the three annotations (plus a rule too large for Julia's heuristics) in forward and reverse mode; `ignore_derivatives` and `within_autodiff()` in natively called rules; each specsig shape (boxed struct with uninitialized fields, small integers, ghost unions) with an `@inline` rule that `check_rule_specsig` compares against Julia's own codegen and a `@noinline` rule called natively; boxed tapes, `@nospecialize` arguments and mutable payloads; the argument classifier and the error for code without a specialized entry point; and nested differentiation. `test/rules/invoke_ci.jl` covers the 1.13 invoke case for functions, callable structs and closures.

Tested locally: `rules`, `basic`, `advanced` and `core` suites on 1.12 (14658 pass / 0 fail, 5 pre-existing broken) and `rules` on 1.13.0-rc1 (13956 pass / 0 fail).


gbaraldi:
We should query if swiftself is being used instead of blindingly enabling it. (I've thought about not using it since it pessimizes register allocation somewhat)

wsmoses:
@vchuravy this has conflicts with main, can you rebase?

src/rules/customrules.jl
```diff
@@ -1975,13 +1975,21 @@ function enzyme_custom_common_rev(
                     extract_roots_from_value!(B, tape, tape_al)
                 end
 
-                al0 = al = emit_allocobj!(B, TapeT, "tape.$TapeT")
-                al = bitcast!(B, al, LLVM.PointerType(llty, addrspace(value_type(al))))
-                store!(B, tape, al)
-                if tape_roots == 0 && any_jltypes(llty)
-                    emit_writebarrier!(B, get_julia_inner_types(B, al0, tape))
+                if tape_roots == 0 && !any_jltypes(llty)
+                    # The rule takes the tape by reference as readonly nocapture, so a
```
wsmoses: this is pointless, we should just always make it an alloca! [i appear to have missed this in an earlier audit]. But it makes sense it matters here since now not inlining will force an extra allocation precisely here

src/compiler.jl
```diff
@@ -1516,6 +1522,366 @@ function nested_codegen!(
...
+`invoke(F, args, nargs, ci)` instead — Enzyme's counterpart of Julia's
+`emit_specsig_to_fptr1`. Ghost and `Type{T}` arguments are materialized as
+literals, scalars and by-reference aggregates are copied into fresh boxes
+(recombined with their inline roots on 1.12+), and the boxed result is loaded
+back, or split into the `sret` buffer and its return roots.
+
+Returns `nothing` for a small `Union` return, which the adapter does not
+handle.
+"""
+function define_rule_jlcall!(mod::LLVM.Module, mi::Core.MethodInstance, @nospecialize(RT::Type), ci::Core.CodeInstance, invoke::Ptr{Cvoid}, name::String, world::UInt)::Union{Nothing, LLVM.Function}
```
wsmoses: we can use this [in follow up] for also fixing the rule calling convention for things that nested_codegen for a custom rule forces with the different abi. Like https://github.com/EnzymeAD/Enzyme.jl/issues/3429 and https://github.com/EnzymeAD/Enzyme.jl/issues/2513 and https://github.com/EnzymeAD/Enzyme.jl/issues/3322

I'll cc @gdalle @kshyatt @ChrisRackauckas and @sotashimozono for visibility of an fyi on

src/compiler.jl
```diff
@@ -1516,6 +1522,366 @@ function nested_codegen!(
...
+end
+
+# The function `customrules.jl` calls a rule through, shaped like the codegen'd rule
+# function used to be: the specsig, swift calling convention, the `enzymejl_mi` /
+# `enzymejl_rt` attributes the rule handlers read, and the per-parameter attributes
+# `prepare_llvm` gives every compiled function (`enzymejl_parmtype*`, and the
+# `enzymejl_rooted_typ` / `enzymejl_returnRoots` markers by which `fix_decayaddr!`
+# recognizes root-carrying arguments).
+function rule_function!(mod::LLVM.Module, mi::Core.MethodInstance, @nospecialize(RT::Type), name::String, world::UInt)::LLVM.Function
+    retty, params, param_attrs = rule_specsig(mi, RT)
```
wsmoses: after nested_codegen! can we also call rule_specsig and assert that the retty, params, and param_attrs match? Just as a way to double check we get this right in more places [including the future]. We could even do so for the original llvm codegen [not just rules] too

src/compiler/interpreter.jl
```diff
@@ -1506,4 +1557,72 @@ function abstract_call_known(
...
+            callee = pop!(workqueue)
+            CC.isinspected(workqueue, callee) && continue
+            CC.markinspected!(workqueue, callee)
+            CC.ci_has_invoke(callee) && continue
+            CC.use_const_api(callee) && continue
+            mi = CC.get_ci_mi(callee)
+            src = callee === ci && root_src !== nothing ? root_src : CC.typeinf_code(interp, mi, true)
+            src isa Core.CodeInfo || continue
+            CC.collectinvokes!(workqueue, src, CC.sptypes_from_meth_instance(mi))
+            if iszero(ccall(:jl_mi_cache_has_ci, Cint, (Any, Any), mi, callee))
```
vchuravy: Use `@ccall`

src/compiler/interpreter.jl
```diff
@@ -1506,4 +1557,72 @@ function abstract_call_known(
...
+                # jl_add_codeinst_to_jit requires the CodeInstance to be rooted and cached
+                CC.code_cache(interp)[mi] = callee
+            end
+            ccall(:jl_add_codeinst_to_jit, Cvoid, (Any, Any), callee, src)
```
vchuravy: Use `@ccall`

src/compiler.jl
```diff
@@ -1516,6 +1523,472 @@ function nested_codegen!(
...
+            throw(CallingConventionMismatchError{String}("Enzyme: failed to infer the custom rule $(funcspec)", funcspec, world))
+        end
+        if rule_call_convention(src) === :inline
+            llvmf = nested_codegen!(enzyme_context, mode, mod, funcspec, world, true)
+            # Keep the two signature derivations in agreement.
+            check_rule_specsig(llvmf, funcspec, ci.rettype)
+            return llvmf
+        end
+        Interpreter.add_codeinsts_to_jit!(interp, ci; root_src = src)
+        specptr, invoke = Interpreter.codeinst_entry(ci)
```
vchuravy: This means we are still using EnzymeInterpreter to infer the code, instead of avoiding the extra overhead and using the native interpreter.

src/compiler/interpreter.jl
```diff
@@ -450,6 +450,35 @@ function Core.Compiler.abstract_call_gf_by_type(
...
+        else
+            CC.get_ci_mi(ci).specTypes
+        end
+        specTypes = simplify_kw(abi)
+        is_primitive_func(specTypes) && return true
+        method_table = CC.method_table(interp)
+        return (interp.inactive_rules && EnzymeRules.is_inactive_from_sig(specTypes; world = interp.world, method_table)) ||
+            (interp.forward_rules && EnzymeRules.has_frule_from_sig(specTypes; world = interp.world, method_table)) ||
+            (interp.reverse_rules && EnzymeRules.has_rrule_from_sig(specTypes; world = interp.world, method_table))
+    end
```
vchuravy: This query should be factored out into a common method.

src/compiler/interpreter.jl
```diff
@@ -1280,6 +1309,30 @@ function abstract_call_known(
...
+                # `abstract_call_known` requires a known callee. A callee with
+                # no singleton type, for example a closure instance, keeps the
+                # `abstract_invoke` path below.
+                if f′ !== nothing
+                    fargs′ = fargs === nothing ? nothing : Core.Compiler.invoke_rewrite(fargs)
+                    return abstract_call_known(interp, f′, ArgInfo(fargs′, argtypes′), si, sv, get_max_methods(interp, f′, sv))
+                end
+            end
+        end
+    end
```
vchuravy, line 1350: This is a general bug fix for 1.13

src/compiler/interpreter.jl
```diff
@@ -1506,4 +1559,84 @@ function abstract_call_known(
...
+
+@static if HAS_INVOKE_RULES
+
+    """
+        add_codeinsts_to_jit!(interp, ci::CodeInstance; root_src = nothing)
+
+    Hand `ci` and its transitive `:invoke` callees, as inferred by `interp`, to Julia's
+    JIT (`jl_add_codeinst_to_jit`). Reading `ci`'s entry point then compiles it as
+    native code. This mirrors `Core.Compiler.add_codeinsts_to_jit!`, but takes each
+    source from `typeinf_code`, because `EnzymeInterpreter` keeps no codegen cache.
```
vchuravy: Enzyme Interpreter should keep a codegen cache.

vchuravy (commented):
Notes:

- Should we switch the abstract interpreter to the native one when we encounter a rule? Otherwise, we infer the subtree anyways. Potential issue is that `has_frule` is broader than the final query might be. 
- Bigger question if we can/should use the native interpreter for the invoke call. 
- If we don't use the native codegen, should we use the Primal version of the Enzyme Interpreter

src/compiler/interpreter.jl
```diff
@@ -1506,4 +1556,63 @@ function abstract_call_known(
...
+    isdefined(Core.Compiler, :SOURCE_MODE_ABI) &&
+    isdefined(Core.Compiler, :MaybeCompressed) &&
+    isdefined(Core.Compiler, :use_const_api) &&
+    isdefined(Core.Compiler, :is_inlineable) &&
+    isdefined(Core.Compiler, :is_declared_inline) &&
+    hasmethod(Core.Compiler.is_declared_inline, Tuple{Method}) &&
+    isdefined(Core.Compiler, :is_declared_noinline) &&
+    hasmethod(Core.Compiler.is_declared_noinline, Tuple{Method}) &&
+    isdefined(Core.Compiler, :typeinf_ext_toplevel) &&
+    hasmethod(Core.Compiler.typeinf_ext_toplevel, Tuple{Core.Compiler.AbstractInterpreter, Core.MethodInstance, UInt8})
```
vchuravy: We don't do this fine-grained feature testing. The failure should be loud.

vchuravy:
This is pretty much ready from my side. It now uses the native code if possible.

src/compiler.jl
```diff
@@ -1516,6 +1519,645 @@ function nested_codegen!(
...
+
+    Say if Julia's codegen passes a value of type `T` boxed, as a tracked
+    `jl_value_t*`, rather than unboxed on the stack (`deserves_argbox` in
+    `codegen.cpp`). Only a concrete immutable type that is a singleton, or
+    that Julia can allocate inline (`jl_datatype_isinlinealloc`), goes
+    unboxed. A struct with uninitialized fields, or one whose field layout
+    the GC cannot describe inline, is boxed even though `jl_type_to_llvm`
+    gives it an LLVM struct type.
+    """
+    function deserves_argbox(@nospecialize(T::Type))::Bool
```
wsmoses: i swear we have this elsewhere, gpucompiler for example?

src/compiler.jl
```diff
@@ -1516,6 +1519,645 @@ function nested_codegen!(
     return decl
 end
 
+"""
+    copy_extension_attrs!(call::LLVM.CallInst, fn::LLVM.Function)
+
```
wsmoses: can we move this to a separate file, like callconv.jl or something?

src/compiler.jl
```diff
@@ -1516,6 +1519,645 @@ function nested_codegen!(
...
+    With `gcstack_arg` set, `pgcstack` follows the return parameters as a
+    `swiftself` parameter. The native call path requires
+    [`jit_uses_swiftcc`](@ref), so the attribute holds there. Pass `false` to
+    model a function compiled without it, as `nested_codegen!` emits.
+
+    `classify_arguments`, `get_return_info` and `enzyme_custom_setup_args`
+    follow the same rules, so the derived signature matches the arguments
+    `customrules.jl` passes.
+    """
+    function rule_specsig(mi::Core.MethodInstance, @nospecialize(RT::Type); gcstack_arg::Bool = jit_gcstack_arg())
```
vchuravy: This is not specific to rules.

src/compiler.jl
```diff
@@ -1516,6 +1519,645 @@ function nested_codegen!(
...
+
+    """
+        rule_arg_kind(T) -> Symbol
+
+    Classify a rule argument of type `T` as `get_specsig_function` does:
+    `:ghost` (omitted), `:boxed` (a tracked pointer), `:byref` (an aggregate
+    passed through a derived pointer, plus an inline-roots pointer when it
+    holds tracked pointers), or `:byval` (a scalar in a register).
+    """
+    function rule_arg_kind(@nospecialize(T::Type))::Symbol
```
vchuravy: Rename to `arg_kind` and check if it doesn't exist in GPUCompiler already

src/compiler.jl
```diff
@@ -1516,6 +1519,645 @@ function nested_codegen!(
...
+
+    Infer and compile the rule `mi` at `world` with Julia's native interpreter
+    and JIT, exactly as an ordinary call of the rule would, and return its
+    `CodeInstance`. Return `nothing` when inference fails. The result is the
+    same `CodeInstance` ordinary callers of the rule use, so the rule is
+    compiled at most once per process. See [`invoke_codegen!`](@ref) for why
+    the native interpreter, and not `EnzymeInterpreter`, infers the rules
+    that are called natively.
+    """
+    function rule_codeinst(mi::Core.MethodInstance, world::UInt)::Union{Nothing, Core.CodeInstance}
```
vchuravy: Not rule specific, rename to just `codeinst`

src/compiler.jl
```diff
@@ -1516,6 +1519,645 @@ function nested_codegen!(
...
+    a MethodInstance other than `mi`, or without a specialized entry point
+    (the boxed `jl_fptr_args` ABI, which Julia picks when every argument is
+    boxed and so is the return). The declaration is derived from
+    `mi.specTypes`, so the first would bind it to code with another
+    signature. Every rule takes an annotation, an immutable struct passed
+    unboxed, so Julia always gives a rule a specialized entry point. Neither
+    case is known to occur; an error reports the broken assumption instead
+    of hiding it.
+    """
+    function native_rule_codeinst(mod::LLVM.Module, mi::Core.MethodInstance, world::UInt)::Union{Nothing, Tuple{Core.CodeInstance, Ptr{Cvoid}}}
```
vchuravy:
```suggestion
    function native_codeinst(mod::LLVM.Module, mi::Core.MethodInstance, world::UInt)::Union{Nothing, Tuple{Core.CodeInstance, Ptr{Cvoid}}}
```

src/compiler.jl
```diff
@@ -1516,6 +1519,645 @@ function nested_codegen!(
...
+        native_rule_return_type(mod::LLVM.Module, mi::MethodInstance, world::UInt) -> Union{Nothing, Type}
+
+    Return the return type of the rule `mi` as its natively compiled code has
+    it, when [`invoke_codegen!`](@ref) binds the rule, called from `mod`, to
+    that code. Return `nothing` when it emits the rule with `nested_codegen!`
+    instead. The rule handlers derive the tape type and check the returned
+    derivatives against this type, so it must come from the inference that
+    produced the code they call.
+    """
+    function native_rule_return_type(mod::LLVM.Module, mi::Core.MethodInstance, world::UInt)::Union{Nothing, Type}
```
vchuravy:
```suggestion
    function native_return_type(mod::LLVM.Module, mi::Core.MethodInstance, world::UInt)::Union{Nothing, Type}
```

src/compiler.jl
```diff
@@ -1516,6 +1519,645 @@ function nested_codegen!(
...
+            Nested differentiation is the exception to "never differentiates". When a
+            derivative is differentiated again, its rule calls must be
+            differentiated through, so they need a body. [`materialize_native_rules!`](@ref)
+            gives the declarations one before the outer differentiation runs.
+
+    Every rule is emitted by `nested_codegen!` where the call ABI does not
+    exist (see [`native_rules_available`](@ref)) and on Julia without the 1.12
+            compiler API (`Interpreter.HAS_INVOKE_RULES`).
+    """
+    function invoke_codegen!(
```
vchuravy: Check if there are other callers of nested_codegen that should become invoke_codegen

src/compiler.jl
```diff
@@ -1516,6 +1519,645 @@ function nested_codegen!(
...
+            enzyme_context::EnzymeContext,
+            mode::API.CDerivativeMode,
+            mod::LLVM.Module,
+            funcspec::Core.MethodInstance,
+            world::UInt,
+            alwaysinline::Bool = false,
+        )
+        if !(funcspec.specTypes isa DataType) || !native_rules_available(mod)
```
vchuravy: native_rule_available -> native_invoke_available

src/compiler.jl
```diff
@@ -1516,6 +1519,645 @@ function nested_codegen!(
...
+            # Emitted rules do not use the derived signature. Check it against
+            # them anyway, so `rule_specsig` stays correct for every rule shape.
+            # `EnzymeInterpreter` inferred the emitted rule, so take the return
+            # type from the emitted function.
+            check_rule_specsig(llvmf, funcspec, enzyme_custom_extract_mi(llvmf)[2])
+            return llvmf
+        end
+        ci, specptr = native
+
+        name = "ejl_rule_" * GPUCompiler.safe_name(string(funcspec.def.name)) * "_" * string(convert(UInt, pointer_from_objref(funcspec)))
```
vchuravy:
```suggestion
        name = "ejl_enzyme_" * GPUCompiler.safe_name(string(funcspec.def.name)) * "_" * string(convert(UInt, pointer_from_objref(funcspec)))
```

src/compiler.jl
vchuravy: Done in 3b04c2e6: the whole native-call machinery now lives in `src/compiler/callconv.jl`, with the rule-specific names made generic.

src/compiler.jl
vchuravy: GPUCompiler has `deserves_argbox`/`deserves_stack` for the boxing decision, but its `is_inlinealloc` approximates `jl_datatype_isinlinealloc` (it checks the `mayinlinealloc` flag and `n_uninitialized`, not the field descriptor width). Enzyme's version goes through `Base.allocatedinline`, which is exactly `jl_datatype_isinlinealloc`, so it stays; the docstring says why.

src/compiler.jl
vchuravy: Renamed to `specsig` and moved to callconv.jl (3b04c2e6).

src/compiler.jl
vchuravy: Renamed to `arg_kind`. GPUCompiler only has the type-level `deserves_argbox`; its `classify_arguments` starts from an LLVM function type, which is what `specsig` derives, so there is no classifier to reuse.

src/compiler.jl
vchuravy: Renamed to `codeinst`.

src/compiler.jl
vchuravy: Applied.

src/compiler.jl
vchuravy: Applied.

src/compiler.jl
vchuravy: Checked. The other callers keep `nested_codegen!`: the `BatchDuplicatedFunc` shadow constructor is inlined into the ABI wrapper, the injected math definitions are there to be differentiated, and the `Threads.@threads` runtime bodies call module-local derivative functions. All three need their IR in the module.

src/compiler.jl
vchuravy: Renamed to `native_invoke_available`.

src/compiler.jl
vchuravy: Applied.

src/compiler/interpreter.jl
vchuravy: Done in d57b90d8: `HAS_INVOKE_RULES` is `VERSION >= v"1.12-"` and `HAS_INVOKE_CI_LOWERING` is `VERSION >= v"1.13-"`; a renamed internal fails at the first use.

vchuravy:
Pushed 3b04c2e6 and 9c6cdc99 addressing the review:

- The native-call machinery moved to `src/compiler/callconv.jl` and the names are generic (`specsig`, `arg_kind`, `codeinst`, `call_convention`, `native_codeinst`, `native_return_type`, `native_invoke_available`, `materialize_native_invokes!`, `ejl_enzyme_*` declarations).
- Per Billy's request, the derivation is now checked against every function Julia's codegen emits, not only rules: with `Enzyme.Compiler.CheckSpecsig[] = true`, `prepare_llvm` compares each emitted function's signature with `specsig`'s. The test suite sets the flag; the full suite passes on 1.12 with it on (218079 tests). Off by default so a gap only ever breaks CI, not users. It found one thing along the way, which is not a derivation gap: `prepare_llvm` looks emitted functions up by name, and a derivative embedded for nested AD can reuse a fresh function's name because Julia's name counters restart per compilation; only functions without a prior `enzymejl_mi` are checked now.

#3501 is rebased on top.

test/runtests.jl
```diff
@@ -36,7 +36,12 @@ function test_worker(name)
     end
 end
 
-const init_code = quote end
+# Compare the signature of every function Julia emits against the one Enzyme
+# derives for natively called code, on every shape the tests contain.
+const init_code = quote
+    import Enzyme
+    Enzyme.Compiler.CheckSpecsig[] = true
```
wsmoses: is there a reason for gating this vs doing it unconditionally?

vchuravy: No good reason, and the gate is gone as of 9d4b9884: the check now runs for every emitted function, unconditionally.

I measured it first, on 1.12, by timing the check and counting the functions it inspects:

| workload | compile | functions checked | time in check |
|---|---|---|---|
| five autodiff calls (scalar, array, loop, matmul) | 15.2 s | 48 | 6.3 ms (0.04%) |
| TensorKit `qr_compact` thunk from #3392 | 41-44 s | 27 | 6-17 ms (0.02-0.04%) |

About 130 µs per emitted function, and end-to-end wall clock with it off vs on is inside run-to-run noise (15.26/15.47/14.99 s vs 15.25/16.06/15.10 s).

The only argument for gating was blast radius, since a derivation gap on a shape the tests don't contain would now throw for a user who never calls a rule. But a mismatch is a real ABI bug that otherwise surfaces as corruption at the first native call, so failing loudly where it arises is better. The full suite (218k tests) passes with it on.

wsmoses:
@vchuravy I'm going to let you do the final read of this [especially comments and functions], then merge when happy

src/compiler/callconv.jl
```diff
@@ -0,0 +1,673 @@
...
+    Say if functions called from `mod` may be bound to natively compiled code (see
+    [`invoke_codegen!`](@ref)). That binds a process address, so not during
+    precompilation, and only for modules that target the host. The rule
+    handlers pass `pgcstack` only through a `swiftself` parameter, so the JIT
+    must use the swift calling convention when it passes one (see
+    [`jit_uses_swiftcc`](@ref)).
+    """
+    function native_invoke_available(mod::LLVM.Module)::Bool
+        return !Base.generating_output() && module_targets_host(mod) &&
+            !(jit_gcstack_arg() && !jit_uses_swiftcc())
```
vchuravy: This limitation seems unnecessary; we just need to match expectations during emission.

vchuravy: Lifted in c177d96b. Julia only drops the `swiftself` attribute and the swift calling convention on a target without it (`jl_codegen_output_t::use_swiftcc`, RISC-V); the `pgcstack` parameter itself stays in the same place with the C convention, so there was a signature to derive rather than a case to refuse.

`specsig` now adds `swiftself` only where `jit_uses_swiftcc()` holds, and marks the parameter `gcstack` in every case, as Julia 1.13's codegen does (1.12 marks it not at all, hence our own). The places that read the presence of `pgcstack` back off a function went through `has_swiftself`, which is only the mark on a target with the convention: `check_specsig`, `materialize_native_invokes!` and the two rule handlers ask the new `gcstack_arg_index` / `has_gcstack_arg` for either mark instead. The rule handlers take the call site's `swiftself` from the callee's parameters, in `copy_abi_attrs!` (was `copy_extension_attrs!`), rather than adding it at a computed index, so a target with no attribute to copy is right by construction.

@gbaraldi this also answers your point: `jit_uses_swiftcc()` shapes the emitted signature now instead of switching the feature off. Should Julia stop using the convention on a target that has it, the derivation follows.

`rules` passes on 1.12.7 (13979 pass / 0 fail / 2 broken) and on 1.13.0-rc4 (13994 / 0 / 2), `basic` on 1.12.7. Only the swiftcc branch runs on CI hardware; the other is derived against `get_specsig_function`.

src/compiler/interpreter.jl
vchuravy: Factored out in 3cf48f40. `FutureCallinfoByType` and `invoked_ci_needs_rule` share `enzyme_call_kind(interp, specTypes)`, which returns `:primitive`, `:alwaysinline`, `:inactive`, `:frule`, `:rrule` or `nothing`; the invoke path is `kind !== nothing && kind !== :alwaysinline`, since an always-inlined signature gets a body from the inliner like any other call.

One thing the duplication had hidden: this query called `EnzymeRules.is_inactive_from_sig` and friends directly, where the ordinary path goes through the memoized `cached_is_inactive` / `cached_has_frule` / `cached_has_rrule`. The `invoke(f, ci, args...)` path gets that cache now.

PR #3481, opened 2026-08-25
Give a Const output buffer a shadow in the GPUArrays linalg rules

vchuravy:
`EnzymeRules.augmented_primal` for the GPUArrays `generic_matmatmul!` /
`generic_matvecmul!` rules read `out.dval` whenever `needs_shadow(config)`,
which errors with `type Const has no field dval` when the output buffer is
`Const`:

```
type Const has no field dval
```

That happens when the rule is reached through a dynamic dispatch, which some
GPUArrays versions emit out of `mul!`. `runtime_generic_augfwd` guesses the
call's return activity from its return type alone, so the rule is asked for a
shadow of the output whatever C's own activity is. The error surfaces at the
`llvmcall` in `enzyme_call` with none of the rule's own frames, since the
augmented-forward function is separately JIT'd.

A `Const` output has no shadow to hand back, so give it a freshly zeroed
buffer, one per batch element. Returning the primal instead would alias the
two: nothing downstream re-checks `dval === val` unless runtime activity is
on, so once the returned value is consumed the consumer's reverse pass
accumulates into the shadow and silently leaves that cotangent in the user's
`Const` buffer — and at width > 1 every batch element would share one array.
A constant's derivative is zero, and the rule's own reverse returns early on a
`Const` C, so the buffer is only ever a sink.

All four matmatmul/matvecmul rules share `_augreturn`, so all four are covered.

### Tests

`test/ext/jlarrays.jl` gets a "Const output buffer, dynamic dispatch" testset:
the erroring call at width 1 and 2, plus a chained `R = dyn_gemm!(C, A, B);
mul!(D, R, E)` that consumes the returned value and asserts the `Const` buffer
still holds `A·B` afterwards. 36/36 pass with the fix; the chained assertions
fail if the shadow aliases the primal.

### DIFF
--- a/ext/EnzymeGPUArraysCoreExt.jl
+++ b/ext/EnzymeGPUArraysCoreExt.jl
@@ -59,20 +59,30 @@ end
 @inline _isconst(config, x::Annotation) =
     EnzymeRules.runtime_activity(config) && x.dval === x.val
 
-# One value per batch element.
-@inline _bsplat(::Val{1}, v) = v
-@inline _bsplat(::Val{N}, v) where {N} = ntuple(Returns(v), Val(N))
+# A fresh zeroed buffer per batch element. A struct instead of a closure so
+# `ntuple` gets something concrete.
+struct _ZeroLike{V}
+    val::V
+end
+@inline (f::_ZeroLike)(::Int) = zero(f.val)
+@inline _bzeros(::Val{1}, v) = zero(v)
+@inline _bzeros(::Val{N}, v) where {N} = ntuple(_ZeroLike(v), Val(N))
 
 #=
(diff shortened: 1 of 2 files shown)


wsmoses:
this isn't the right resolution, was resolved by https://github.com/JuliaGPU/GPUArrays.jl/pull/764

vchuravy:
How is changing the code to inline something the right fix for a potential rule issue?

wsmoses:
the rule must return C as the result. Having the activities mismatch isn't supported by the current rule [and I'm a bit skeptical that adding an extra zero shadow here is the right resolution, but I can look more closely if you think so]

The PR to gpuarrays fixed dataflow analysis to ensure the invoke edge still kept the C / returned property necessary for activity analysis to not be busted

PR #3485, opened 2026-08-27
GPU gradient fixes

jgreener64:
I am getting Molly's gradients working on GPU and had to change the following to get things to work:

1. `autodiff_deferred` mishandles an explicit `Active{T}` return type (`Enzyme/src/Enzyme.jl`). It has `rt = A` where the non-deferred `autodiff` has `eltype(A0)`. With `A = Active{Float64}`, `rt` becomes `Active{Float64}`, so `default_adjoint(rt)` falls into its "not a floating-like value" branch and the seed is never produced - which on a GPU surfaces as an `InvalidIRError` about a `string(...)` call rather than an error message. Fixed here with `rt = eltype(A)`.

2. Enzyme's CUDA copy rules are bounded to `T <: AbstractFloat` (`Enzyme/ext/EnzymeCUDAExt.jl`, the `reverse` halves and the `_zero!` helpers). The `augmented_primal` halves have no such bound, so a host↔device copy of anything else - a `Bool` array, or a `CuArray{SVector{3,Float64}}` of coordinates - gets through the forward pass and then has no `reverse` method. Relaxed here to any eltype; the bodies are a memset and `dst .+= src`, both already correct for those types, and they are skipped entirely when the argument is `Const`.

3. Checking `isa Const` is not enough. With `set_runtime_activity` Enzyme hands a rule a `Duplicated` whose `dval` is the `val`, for a value it could not prove inactive statically. `_accumulate!` then adds the cotangent into the primal, and `_zero!` zeroes a primal. Fixed with an `_active_shadow` helper that returns `nothing` when `dval === x.val`, and skipping both operations in that case.

Co-authored with Claude.

### DIFF
--- a/ext/EnzymeCUDAExt.jl
+++ b/ext/EnzymeCUDAExt.jl
@@ -7,11 +7,11 @@ using Enzyme: EnzymeRules
 # Complex is handled here because the shadow operations are element-wise: zeroing is a
 # `memset` sized by `sizeof(T)`, and accumulation is a broadcast, both of which are
 # agnostic to the real/complex split.
-function _zero!(ptr::Ptr{T}, off::Integer, n::Integer) where {T <: Union{AbstractFloat, Complex{<:AbstractFloat}}}
+function _zero!(ptr::Ptr{T}, off::Integer, n::Integer) where {T}
     Base.Libc.memset(ptr + off * sizeof(T), 0, n * sizeof(T))
     return nothing
 end
-function _zero!(ptr::CuPtr{T}, off::Integer, n::Integer) where {T <: Union{AbstractFloat, Complex{<:AbstractFloat}}}
+function _zero!(ptr::CuPtr{T}, off::Integer, n::Integer) where {T}
     bytes = reinterpret(CuPtr{UInt8}, ptr + off * sizeof(T))
     CUDA.memset(bytes, UInt8(0), n * sizeof(T))
     return nothing
@@ -67,6 +67,15 @@ end
 
 @inline function _shadow(x, config, batch)
     return EnzymeRules.width(config) == 1 ? x.dval : x.dval[batch]
+end
+
+# With `set_runtime_activity` Enzyme hands a rule a `Duplicated` whose shadow is the primal
+# itself, for a value it could not prove inactive statically. Accumulating a cotangent into
+# such a shadow, or zeroing it, writes to the primal data.
+@inline function _active_shadow(x, config, batch)
+    x isa Const && return nothing
+    dval = _shadow(x, config, batch)
+    return dval === x.val ? nothing : dval
 end
 
 # This rule should _NOT_ be needed. Without it there appears to be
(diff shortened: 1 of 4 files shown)


ext/EnzymeCUDAExt.jl
```diff
@@ -7,11 +7,11 @@ using Enzyme: EnzymeRules
 # Complex is handled here because the shadow operations are element-wise: zeroing is a
 # `memset` sized by `sizeof(T)`, and accumulation is a broadcast, both of which are
 # agnostic to the real/complex split.
-function _zero!(ptr::Ptr{T}, off::Integer, n::Integer) where {T <: Union{AbstractFloat, Complex{<:AbstractFloat}}}
```
wsmoses, line 10: this is suss we should be conservative here, what do you need here?

src/Enzyme.jl
```diff
@@ -732,7 +732,7 @@ code, as well as high-order differentiation.
         end
     else
         @assert A isa DataType
-        rt = A
+        rt = eltype(A)
```
wsmoses, line 735: add a test with this?

ext/EnzymeCUDAExt.jl
jgreener64: We need it for SVectors of floats, and also structs of integers and floats. Added a check for `isbits` which I think is the required check.

src/Enzyme.jl
jgreener64: Added a test, which failed before.

ext/EnzymeCUDAExt.jl
wsmoses: technically speaking it's incorrect, an integer shuold not be zero'd, for example. In most cases this is fine, but an fyi

jgreener64: Changed to not zero cases like integers.

ext/EnzymeCUDAExt.jl
```diff
@@ -201,15 +257,17 @@ for (DstArr, SrcArr) in ARRAY_COPY_DIRECTIONS
...
-			_accumulate!(@view(dsrc[soffs.val:soffs.val+n.val-1]), @view(ddest[doffs.val:doffs.val+n.val-1]))
+                    ddest = _active_shadow(dest, config, batch)
+                    isnothing(ddest) && continue
+                    dsrc = _active_shadow(src, config, batch)
+                    if !isnothing(dsrc)
+                        _accumulate!(@view(dsrc[soffs.val:soffs.val+n.val-1]),
+                                     @view(ddest[doffs.val:doffs.val+n.val-1]))
                     end
-		    _zero!(pointer(ddest, doffs.val), 0, n.val)
+                    _zero!(pointer(ddest, doffs.val), 0, n.val)
```
wsmoses: we need to gc preserve ddest here

jgreener64: Added `GC.@preserve ddest`.

jgreener64:
Does this need further changes?

PR #3486, opened 2026-08-27
Fix undefined batched ABI wrapper index

ChrisRackauckas-Claude:
> Ignore this PR until it has been reviewed by @ChrisRackauckas.

## What changed and why

`create_abi_wrapper` used the undefined variable `t` while extracting a batched shadow return from a concrete annotation. This changes the extraction loop to use the already-computed `twidth`.

The regression uses an uninferred tuple return with a concrete `BatchDuplicated` return annotation in forward mode, which exercises the generic ABI wrapper without involving the separate `ForwardWithPrimal` sret layout.

## Failing before

I ran the new forward-only regression against unmodified `a865b354` on Julia 1.12.7. It failed before reaching an assertion:

```text
ERROR: UndefVarError: `t` not defined in `Enzyme.Compiler`
Stacktrace:
 [1] create_abi_wrapper(...)
   @ Enzyme.Compiler src/compiler.jl:3897
 [2] autodiff(... ::Type{BatchDuplicated{Tuple{Vector{Float64}}, 2}}, ...)
```

## Passing after

The same regression is included in `test/typeunstable.jl`. The direct MWE returns both expected batch derivatives:

```text
result = (var"1" = (var"1" = ([2.0, 2.0],), var"2" = ([6.0, 6.0],)),)
```

I ran:

```bash
TMPDIR="$PWD/../.tmp" JULIA_PKG_PRECOMPILE_AUTO=0 julia +1.12 --startup-file=no --project=test test/runtests.jl typeunstable
```

```text
Test Summary: | Pass  Total     Time
  Overall     |   36     36  1m11.3s
    SUCCESS
```

Additional local checks:

```text
../git-runic --julia ~/.juliaup/bin/julia --project ../.runic-env -f upstream/main
runic did not modify any files

git diff --unified=0 upstream/main -- | typos -
# no findings

git diff --check upstream/main
# no findings
```

## Not verified

I did not run the full Enzyme test suite, GPU tests, integration tests, or documentation build. This changes no public API or documentation.

## Reviewer attention

The loop now starts at one and stops at `twidth`, which is the shadow width already used to construct `ival`.

https://chatgpt.com/codex/tasks/01a03f84-af17-7a90-955c-10b5d980e4da

### DIFF
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -3866,7 +3866,7 @@ function create_abi_wrapper(
                 twidth = if width == 1
                     1
                 else
-                    if returnPrimal && returnNum == count_Sret - 1
+                    if (rettype <: Const) && returnNum == 0
                         1
                     else
                         width
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -3866,7 +3866,7 @@ function create_abi_wrapper(
                 twidth = if width == 1
                     1
                 else
-                    if (rettype <: Const) && returnNum == 0
+                    if returnPrimal && returnNum == count_Sret - 1
                         1
                     else
                         width
@@ -3894,7 +3894,7 @@ function create_abi_wrapper(
                     end
                 else
                     ival = UndefValue(LLVM.LLVMType(API.EnzymeGetShadowType(twidth, SPT0)))
-                    for idx = t:width
+                    for idx in 1:twidth
                         pv = extract_value!(builder, eval, idx - 1)
                         pv0 = pv
                         pv = bitcast!(builder, pv, LLVM.PointerType(SPT0, addrspace(value_type(pv))))
--- a/test/typeunstable.jl
+++ b/test/typeunstable.jl
@@ -330,3 +330,20 @@ unstable_batch_3425(p) = (y = unstable_tuple_3425(p); y[1]^2 + y[2]^2)
         end
     end
 end
+
+batched_uninferred_target(x) = (2 .* x,)
+batched_uninferred_return(x) = Base.invokelatest(batched_uninferred_target, x)
+
+@testset "Batched forward concrete annotation with uninferred return" begin
+    x = [1.0, 2.0]
+    dx1 = ones(2)
+    dx2 = fill(3.0, 2)
+    return_activity = BatchDuplicated{Tuple{Vector{Float64}}, 2}
+    result = Enzyme.autodiff(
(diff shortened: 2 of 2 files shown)


src/compiler.jl
```diff
@@ -3894,7 +3894,7 @@ function create_abi_wrapper(
                     end
                 else
                     ival = UndefValue(LLVM.LLVMType(API.EnzymeGetShadowType(twidth, SPT0)))
-                    for idx = t:width
+                    for idx in 1:twidth
```
wsmoses, line 3897: this one i agree with for sure, can you explain the need for the earlier change [and maybe separate the two into diff prs w different mwe tests]

ChrisRackauckas-Claude: Split now. https://github.com/EnzymeAD/Enzyme.jl/pull/3486 only changes `t:width` to `1:twidth`, and its MWE now uses `Forward` rather than `ForwardWithPrimal`, so it exercises only the undefined-index bug. I reran it against unmodified main and observed `UndefVarError: t`; the narrowed branch returns both expected batch derivatives and the targeted suite passes 36/36.

The earlier change was for a separate failure that becomes visible after fixing `t`. For batched `ForwardWithPrimal`, this ABI path has two sret entries: the derivative entry (`returnNum == 0`) is a width-wide aggregate, while the final primal entry (`returnNum == count_Sret - 1`) is a single pointer. With width 2, I observed the former as `[2 x ptr]` and the latter as `ptr`. The existing `twidth` condition assigns width 2 to that scalar primal entry because the requested return annotation is `BatchDuplicated`, not `Const`; `extract_value!` then treats the pointer as an aggregate and the MWE segfaults at `create_abi_wrapper:3898` (exit 139).

That follow-up should preserve the existing `Const` special case and add the primal case, i.e. use an OR rather than the replacement I originally proposed. I will submit it separately after this PR, with a distinct `ForwardWithPrimal` MWE that currently reproduces the exit-139 failure on the narrowed branch.

ChrisRackauckas-Claude: The separate `ForwardWithPrimal` slot-width fix and its distinct SciML-free reproducer are now in draft PR https://github.com/EnzymeAD/Enzyme.jl/pull/3488. It is stacked on https://github.com/EnzymeAD/Enzyme.jl/pull/3486 so the undefined-index prerequisite remains separate.

PR #3511, opened 2026-09-03
Make deferred codegen GPU-only with ids derived from the job specification

vchuravy:
Stacked on #3510. Part of the cache redesign towards session-local state and relocatable, cross-session cacheable compilation.

`deferred_id_codegen` embedded a host JIT trampoline address as the id of a
deferred job: GPUCompiler's host stub returns the id unchanged, so on the host
the id doubled as the function pointer. That address is only valid in the
session that created it, and a `deferred_id_codegen` CodeInstance loaded from
a package image returns a dead pointer.

Deferred codegen is only needed inside GPU kernels: on the host, and for
higher-order differentiation, `autodiff` is the entry point (Enzyme recognizes
nested `autodiff` calls itself). So:

- The id is now derived from the job's specification (`DeferredSpec`, entirely
  type-level), so it is the same in every session, and `deferred_codegen_jobs`
  can be repopulated from the specializations of `deferred_id_codegen`
  (`rebuild_deferred_registry!`). A high bit keeps the ids disjoint from
  GPUCompiler's counters.
- `deferred_codegen` emits the marker twice, under the id and a twin id that
  is registered to the same job. GPUCompiler's pass resolves both to one
  function, so the comparison folds away in device code before any later pass
  sees the function's address (GPUCompiler rewrites the signatures of
  functions using the kernel state and only accepts call and store uses). On
  the host both markers reach GPUCompiler's identity stub and differ, and the
  call throws `DeferredOnHostError`.
- The host trampoline machinery is gone: `get_trampoline`, `add_trampoline!`,
  and the lazy call-through and indirect-stubs managers in `JIT`.

Known gap: a kernel CodeInstance loaded from a package image whose device
object is not cached still misses in `deferred_codegen_jobs` inside
GPUCompiler's pass, which has no lookup fallback; that needs an upstream hook.

Tests that called `autodiff_deferred` natively on the host now use `autodiff`
(`core/abi`, the nested helpers in `basic`, `advanced` and
`rules/rule_inlining`, the CPU reference in `metal`), except the two `abi`
checks of errors that deferred codegen raises before the host check; the
`autodiff_deferred_thunk` docstring example is no longer a host doctest.
`core/deferred` covers the host error, id stability within and across
sessions, the twin ids and the registry rebuild. CUDA's test suite passes with
GPUCompiler 1.23/CUDA 6.2.2 and, apart from a pre-existing reverse-mode
limitation on `trylock` that CUDA 6.3's per-allocation locks expose, with
GPUCompiler 2.5/CUDA 6.3.1.

https://claude.ai/code/session_01LdVAXqghGYwHGrnzaFchPT

### DIFF
--- a/src/Enzyme.jl
+++ b/src/Enzyme.jl
@@ -694,8 +694,10 @@ end
 """
     autodiff_deferred(::ReverseMode, f, Activity, args::Annotation...)
 
-Same as [`autodiff`](@ref) but uses deferred compilation to support usage in GPU
-code, as well as high-order differentiation.
+Same as [`autodiff`](@ref) but uses deferred compilation, for use inside GPU kernels.
+Calling it on the host throws `Enzyme.Compiler.DeferredOnHostError`; use `autodiff` there,
+including for higher-order differentiation, where Enzyme differentiates through nested
+`autodiff` calls directly.
 """
 @inline function autodiff_deferred(
     mode::ReverseMode{ReturnPrimal,RuntimeActivity,StrongZero,RABI,Holomorphic,ErrIfFuncWritten},
@@ -830,8 +832,9 @@ end
 """
     autodiff_deferred(::ForwardMode, f, Activity, args::Annotation...)
 
-Same as `autodiff(::ForwardMode, f, Activity, args...)` but uses deferred compilation to support usage in GPU
-code, as well as high-order differentiation.
+Same as `autodiff(::ForwardMode, f, Activity, args...)` but uses deferred compilation, for
+use inside GPU kernels. Calling it on the host throws `Enzyme.Compiler.DeferredOnHostError`;
+use `autodiff` there, including for higher-order differentiation.
 """
 @inline function autodiff_deferred(
     mode::ForwardMode{ReturnPrimal,RABI,ErrIfFuncWritten,RuntimeActivity,StrongZero},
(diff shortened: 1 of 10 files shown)


src/Enzyme.jl
```diff
@@ -1382,13 +1385,12 @@ The reverse function will return the derivative of `Active` arguments, updating
 arguments in place. The same arguments to the forward pass should be provided, followed by
 the adjoint of the return (if the return is active), and finally the tape from the forward pass.
 
-Example:
-
-```jldoctest
+Like [`autodiff_deferred`](@ref), this is for use inside GPU kernels; on the host it throws
+`Enzyme.Compiler.DeferredOnHostError`, and [`autodiff_thunk`](@ref) is the entry point.
```
vchuravy, line 1389: @wsmoses is this behavior change fine? I think it is technically a breaking change, but it let's us delete a lot of code.

test/core/abi.jl
```diff
@@ -29,13 +29,13 @@ end
     
     @test () === autodiff(Forward, f, Const(nothing))
 
-    res = autodiff_deferred(Reverse, Const(f), Const, Const(nothing))
+    res = autodiff(Reverse, Const(f), Const, Const(nothing))
```
wsmoses, line 32: So a lot of these tests are duplicated for both deferred and not, so just changing back to non deferred may not actually be helpful

src/Enzyme.jl
wsmoses: I'm a bit torn here. People do use autodiff_deferred outside of gpu kernels to make sure they're safe, even after we started auto-upgrading to deferred within gpu calls. Also the deferred codepath is different so we'll need a different suite of abi/etc tests to confirm we didn't mess up

wsmoses: But I do agree it lets us delete a lot of code and simplify.

wsmoses: there are a nontrivial number of users of: https://github.com/search?q=%22autodiff_deferred%22&type=code

PR #3512, opened 2026-09-03
Support GPUCompiler 2.x alongside 1.6

### LINKED ISSUE #3492: GPUCompiler v2?

vchuravy:
Stacked on #3511. Part of the cache redesign towards session-local state and relocatable, cross-session cacheable compilation.

GPUCompiler 2.x moved its code-instance cache onto CompilerCaching.jl and
renamed the hooks Enzyme implements: `ci_cache_token` became `cache_owner`,
`ci_cache` (Julia 1.10) became `get_code_cache`, `runtime_slug` and `codegen`
are gone, and `compile_method_instance` drives inference through
`drive_inference!`, which GPUCompiler only provides for its own interpreter.
It also keeps Julia-value references symbolic (`Relocations`) until a job is
linked: a toplevel host job resolves them before Enzyme sees the module, but a
job compiled on behalf of a kernel (`toplevel = false`) does not.

Enzyme now supports both majors (`HAS_GPUCOMPILER_2`):

- `enzyme_cache_owner(job)` builds the `EnzymeCacheToken`; it is installed as
  `GPUCompiler.cache_owner` or `GPUCompiler.ci_cache_token` depending on the
  version, likewise `get_code_cache` / `ci_cache` on 1.10.
- `GPUCompiler.drive_inference!(::EnzymeInterpreter, mi)` runs
  `CompilerCaching.typeinf!` on 1.11+ and the `CodeCache` path on 1.10.
- `resolve_relocations!` bakes the primal module's Julia-value references to
  host addresses right after `emit_llvm`, whatever the job's `toplevel`. That
  is the IR shape GPUCompiler 1.x always produced, so nothing downstream
  (`absint`, shadow globals, `try_replace_constant_load!`) changes.
  `compile_unhooked` still returns the (emptied) records, because GPUCompiler's
  deferred-codegen path reads them off the result when linking.
- `check_ir!` walks PLT stubs after their users. Rewriting a load of a stub's
  got reads the library and symbol out of the stub's `ijl_load_and_lookup`
  call, and walking the stub folds that call away. GPUCompiler 1.x happened to
  list the stubs last; 2.x lists them first.
- `reflect` uses `compile_unhooked` where `codegen` no longer exists.

Compat is widened to `GPUCompiler = "1.6.2, 2.5.4"`; CI gains jobs that pin
GPUCompiler to 1.x (the default resolution is now 2.x), on the GPU runners too.

Deliberately not in this PR:

- Relocatable derivative modules. Selecting the `:patch` strategy for the
  primal job leaves the `julia.constgv` slots symbolic, and every function
  touching a global `Ref`/`Dict`, and every `autodiff_deferred`, then fails
  with `EnzymeNoShadowError`: shadow-global creation and `absint` key off the
  baked addresses. Supporting it means reading the relocation records in
  `absint`/`abs_typeof`/shadow creation, patching in Enzyme's ORC JIT, and
  emitting records for the shadow objects Enzyme allocates at compile time.
  That is later work in the stack.
- Decoding the module-resident replicas of isbits constants that a device job
  gets from GPUCompiler 2.x (`materialize_box!`). Nothing reaches that on the
  host, and a CUDA derivative cannot be built under GPUCompiler 2.x until
  CUDACore 6.3 ports its launch rules (#3540); `test/cuda.jl`'s Reverse Kernel
  test is marked broken there. It follows as its own PR.

Tested locally: `advanced`, `relocation`, `cache_token`, `method_table`,
`core/deferred`, `core/abi`, `rules/ruleinvalidation` and `passes` under
GPUCompiler 2.6.0 and 1.23.0 on Julia 1.12. CI: CUDA and Metal pass on
Julia 1.10–1.12 under both the 1.x pin and the default resolution; the two
assertions-build failures are the pre-existing ones (Julia 1.12 AllocOpt
assertion while precompiling test deps, Julia 1.13 `absint.jl:88` and the
jlarrays `atomicmodify` abort).

https://claude.ai/code/session_01DZ7DdJSyNZ3QRtKLnS2sdj


src/compiler/validation.jl
```diff
@@ -586,6 +621,11 @@ function check_ir!(interp, @nospecialize(job::CompilerJob), errors::Vector{IRErr
                                 throw(AssertionError(msg))
                             end
 
+                            if resolved !== nothing
+                                newf = LLVM.const_inttoptr(LLVM.ConstantInt(resolved), value_type(inst))
+                                @goto plt_resolved
```
vchuravy: We should avoid the goto

vchuravy:
Needs #3513 to ensure that the buildkite jobs pass.

Project.toml
```diff
@@ -49,7 +49,7 @@ CUDA = "5, 6"
 EnzymeCore = "0.8.21"
 Enzyme_jll = "0.0.291"
 GPUArraysCore = "0.1.6, 0.2"
-GPUCompiler = "1.6.2"
+GPUCompiler = "1.6.2, 2.2"
```
vchuravy:
```suggestion
GPUCompiler = "1.6.2, 2.5.4"
```

vchuravy:
@wsmoses I have an ordering issue here. I would like to merge & release this soon to unblock the work on KernelAbstractions. Right now I have excluded the tests that are broken due to #3540, is that okay with you?

src/compiler/validation.jl
```diff
@@ -396,13 +396,19 @@ function try_replace_constant_load!(@nospecialize(inst::LLVM.Instruction); check
     if isa(addr, LLVM.GlobalVariable) && (haskey(metadata(addr), "julia.constgv") || !check_mutability)
         paddr = addr
         addr = LLVM.initializer(paddr)
+        # Folding needs the object's address. A GPUCompiler 2.x job compiled on behalf of a
```
wsmoses: I don't think this is the right fix. In a [separate from gpucompiler2, but obviously prereq for] pr, can we rewrite this to take the new gv_to_objs map we added to gpucompiler earlier, and then we can check that map rather than here doing the inttoptr interpreting

wsmoses (commented):
also high level but can we separate this into pieces? [even if we cannot run the tests for, it'll make it easier to review]

.buildkite/pipeline.yml
```diff
@@ -39,7 +39,15 @@ steps:
...
-          Pkg.status()' || exit 3
+          Pkg.status()
+          # `Pkg.develop` re-resolves and nothing forces it to keep the pin; fail loudly
+          # rather than let this job silently duplicate the default-resolution one.
+          # (No Julia string interpolation here: buildkite would eat the dollar sign.)
+          if gpucompiler != "default"
+              v = Pkg.dependencies()[Base.UUID("61eb1bfa-7361-4325-ad38-22787b887f55")].version
+              v < VersionNumber(parse(Int, gpucompiler) + 1) ||
+                  error(string("GPUCompiler pin did not hold, resolved ", v))
+          end' || exit 3
```
vchuravy: I don't think this is needed

.buildkite/pipeline.yml
```diff
@@ -39,7 +39,15 @@ steps:
           Pkg.add(specs)
```
vchuravy, line 37: Instead of using Pkg.add we should use `Pkg.pin` for GPUCompiler 1

.buildkite/pipeline.yml
vchuravy, line 32: Does GPUCompiler 1 and CUDA v6 work together?

test/cuda.jl
```diff
@@ -324,7 +324,8 @@ launch bookkeeping itself and reports a spurious activity mismatch. Both `autodi
 below throw, so the `@test`s wrap them rather than sitting after them.
 https://github.com/EnzymeAD/Enzyme.jl/issues/3540
 =#
-const KERNEL_LAUNCH_RULES_MISSING = isdefined(CUDA.CUDACore, :compile_and_launch)
+const KERNEL_LAUNCH_RULES_MISSING =
+    isdefined(CUDA, :CUDACore) && isdefined(CUDA.CUDACore, :compile_and_launch)
```
vchuravy: Use an EnzymeRules query?

.buildkite/pipeline.yml
vchuravy: Dropped — with a real `Pkg.pin` the version is fixed for the resolver, so there is nothing left for the check to catch.

.buildkite/pipeline.yml
vchuravy: Done. `Pkg.pin` only accepts direct dependencies, so the `add` stays (it is what makes GPUCompiler one) and the pin follows it.

.buildkite/pipeline.yml
vchuravy: Yes. CUDACore requires GPUCompiler `2.1.0 - 2` from 6.3 on, and `1.4.0 - 1` before that, so pinning GPUCompiler 1 resolves CUDA 6.2 — the `cuda-*-gpucompiler-1` jobs went green on that. Noted it in the matrix comment.

test/cuda.jl
vchuravy: Switched to `EnzymeRules.has_rrule_from_sig(Tuple{typeof(CUDA.CUDACore.compile_and_launch), Vararg{Any}})`. The `isdefined` guards stay only to establish that this is the 6.3-style launch path at all (CUDA 5.x has no `CUDACore`, 6.2 has no `compile_and_launch`); the actual broken-ness now comes from the rule query, so the tests unbreak on their own once the rules are ported.

timesselens:
The Metal jobs (Julia 1.12 and 1.13) and the CUDA jobs (Julia 1.11 and 1.12) of this PR's Buildkite build 8115, on
6a00e519, fail in `try_replace_constant_load!`:

```
Metal autodiff: Error During Test at test/metal.jl:28
  MethodError: no method matching get_base_and_offset(::Nothing; offsetAllowed::Bool, inttoptr::Bool)
```

`test/cuda.jl` stops the same way at `mul_kernel`, its first reverse-mode kernel (on a GTX 1080 with CUDA.jl 6.4).

This seems to be what happens: since 6a00e519, a `:patch` or `:table` back-end leaves the slots of Julia
values as declarations for the requesting job to resolve (CUDACore uses `:patch`, Metal kernels use `:table`). So
`LLVM.initializer(paddr)` is `nothing` in both branches of `try_replace_constant_load!`. It looks like ec3d3b30 removed
the two guards for that case as unreachable.

A possible fix is to restore the two guards as they were:

```diff
--- a/src/compiler/validation.jl
+++ b/src/compiler/validation.jl
@@ -411,6 +411,10 @@ function try_replace_constant_load!(@nospecialize(inst::LLVM.Instruction); check
     if isa(addr, LLVM.GlobalVariable) && (haskey(metadata(addr), "julia.constgv") || !check_mutability)
         paddr = addr
         addr = LLVM.initializer(paddr)
+        # Folding needs the object's address. A GPUCompiler 2.x job compiled on behalf of a
+        # kernel (`toplevel = false`) keeps the slot symbolic until the kernel is linked, so
+        # there is none yet; the load stays a load.
+        addr === nothing && return inst
         gname = LLVM.name(paddr) * "\$false"
         addr, _ = get_base_and_offset(addr; offsetAllowed = false, inttoptr = true)
         originally_tracked = true
@@ -418,6 +422,8 @@ function try_replace_constant_load!(@nospecialize(inst::LLVM.Instruction); check
         paddr = operands(addr)[1]
         if isa(paddr, LLVM.GlobalVariable) && (haskey(metadata(paddr), "julia.constgv") || !check_mutability)
             addr = LLVM.initializer(paddr)
+            # As above: an unresolved slot has no address to fold to.
+            addr === nothing && return inst
             gname = LLVM.name(paddr) * "\$true"
             base_addr, _ = get_base_and_offset(addr; offsetAllowed = true, inttoptr = false)
             originally_tracked = true
```

This might fit with your plan. #3611 notes that on a non-`:bake` back-end `try_replace_constant_load!` still needs
the address, and #3521 looks the slot up first (`relocation_slot_value`) and keeps these same guards as its fallback.
Restoring them now might let this PR get past these failures on its own, and the lookup could come later with #3521
and #3611. Since #3521 has the guards word for word, its rebase should stay simple. But you know the plan for these
drafts far better than I do, so please say if this doesn't fit.

Also wrote a CPU test for `test/passes.jl`, placed right after "Literal-pointer symbol resolution" rather than at
the end of the file, where #3616, #3536 and #3547 add theirs. Its IR uses typed-pointer syntax, like the existing
`nodecayed_phis!` test, so that every Julia version in CI can parse it:

```julia
# GPUCompiler 2.x leaves the slot of a Julia value without an initializer until the kernel that
# requested the derivative is linked (CUDA's `:patch` and Metal's `:table` relocation lowering).
# IR validation must leave loads from such a slot alone: there is no value to fold them to.
@testset "try_replace_constant_load! skips a slot without an initializer" begin
    LLVM.Context() do ctx
        mod = parse(
            LLVM.Module, """
            @slot = external global {} addrspace(10)*, !julia.constgv !0
            @pslot = external global {} addrspace(10)**, !julia.constgv !0

            define {} addrspace(10)* @direct() {
            top:
              %v = load {} addrspace(10)*, {} addrspace(10)** @slot, align 8
              ret {} addrspace(10)* %v
            }

            define {} addrspace(10)* @indirect() {
            top:
              %p = load {} addrspace(10)**, {} addrspace(10)*** @pslot, align 8
              %v = load {} addrspace(10)*, {} addrspace(10)** %p, align 8
              ret {} addrspace(10)* %v
            }

            !0 = !{}
            """
        )
        for fname in ("direct", "indirect")
            ld = operands(last(collect(instructions(first(blocks(functions(mod)[fname]))))))[1]
            # as `check_ir!` calls it, and as the call lowering does
            @test Enzyme.Compiler.try_replace_constant_load!(ld) == ld
            @test Enzyme.Compiler.try_replace_constant_load!(ld; check_mutability = false, do_replace = false) == ld
        end
        # nothing was folded into the slots either
        @test LLVM.initializer(globals(mod)["slot"]) === nothing
        @test LLVM.initializer(globals(mod)["pslot"]) === nothing
    end
end
```

On 6a00e519, all four calls throw the `MethodError`. With the guards, both loads are left alone and the slots stay
empty. In these runs, `test/metal.jl` then goes from the `MethodError` to 2/2 (M1 Max, Metal.jl 1.11.1), and on the GTX
1080 `test/cuda.jl` gets past `mul_kernel`.

If a PR into `vc/stack-09-gpucompiler-2` would be easier for you, the commit is ready.

<details><summary>Verification</summary>

- The new test, `passes` on 6a00e519 plus the guards (GPUCompiler 2.8.2, Enzyme_jll 0.0.293, macOS aarch64): 51/51 on
  Julia 1.10.12 and 1.11.9, 53/53 on 1.12.5 and 1.13.0 (two existing tests only run from Julia 1.12 on). On 6a00e519
  alone, the new test's four calls throw the `MethodError`.
- `test/metal.jl` (M1 Max, Julia 1.12.5, Metal.jl 1.11.1, GPUCompiler 2.8.1): the `MethodError` at `test/metal.jl:28`
  on 6a00e519, 2/2 with the guards.
- `test/cuda.jl` (GTX 1080, Linux x86_64, CUDA.jl 6.4): it stops at `mul_kernel` on 6a00e519 and gets past it with the
  guards.
- Buildkite build 8115 on 6a00e519: the Metal jobs (1.12, 1.13) and the CUDA jobs (1.11, 1.12) fail with the
  `MethodError`.
</details>

PR #3513, opened 2026-09-03
Add lock/unlock/trylock rules that mirror the lock in the reverse pass

vchuravy:
Reverse-mode differentiation through a `ReentrantLock` that lives inside a `Duplicated` object failed inside `Base._trylock` with "cannot handle unknown instruction: cmpxchg" on the `havelock` field. The same code worked with a global lock, and it surfaced in practice because CUDA.jl 6.3 takes per-allocation locks in `unsafe_copyto!`, so a reverse-mode copy between `Duplicated` CuArrays now hit it.

Enzyme.jl never declared any rule for the lock primitives. Activity analysis saw the lock pointer loaded from a shadowed object, treated the call as a potential active store through it, and enzyme-core recursed into `_trylock` and `_unlock`, where the cmpxchg has no derivative.

Marking the primitives inactive (the first version of this PR) would silently drop the lock from the reverse sweep: the region a lock protects is replayed backwards, so the reverse pass must hold the same lock over the same region. This adds forward and reverse rules for the single-argument `lock`, `unlock` and `trylock` methods on `AbstractLock` instead:

- Forward mode runs the primal only.
- Reverse mode: the augmented primal runs the primal and stores the lock on the tape. The adjoint of `lock` is `unlock`, the adjoint of `unlock` is `lock`, and the adjoint of `trylock` is `unlock` exactly when the forward `trylock` acquired.
- Shadow locks are never touched. `islocked` is a pure query and stays inactive.
- The closure-taking `lock(f, l)` / `trylock(f, l)` methods are deliberately not matched because `f` may be active; they reach the rules through their inner single-argument calls.

Tests: `test/locks.jl` gains a `RecordingLock` that logs every acquire and release into a shared vector, and checks the exact event order for lock/unlock and trylock on locks inside `Duplicated` structs in both modes, e.g. for a nested `lock(d); lock(s); ...; unlock(s); unlock(d)` region the log is

```
(:d, :lock), (:s, :lock), (:s, :unlock), (:d, :unlock),   # forward
(:d, :lock), (:s, :lock), (:s, :unlock), (:d, :unlock),   # reverse
```

It also checks that a failed `trylock` is not unlocked in the reverse pass, and that all locks are released afterwards. Ran `test/locks.jl` on Julia 1.12 and 1.10.

https://claude.ai/code/session_01LMa2ZdykwFNF7pAYjwFNc1

### DIFF
--- a/src/internal_rules/core.jl
+++ b/src/internal_rules/core.jl
@@ -662,3 +662,112 @@ function EnzymeRules.reverse(
 ) where {RT,T}
     return (nothing,)
 end
+
+# Locks
+#
+# Taking or releasing a lock carries no derivative information, but the
+# region it protects is replayed in the reverse sweep, so the reverse pass must
+# hold the same lock over the same region.  The adjoint of `lock` is therefore
+# `unlock` and the adjoint of `unlock` is `lock`; `trylock` releases in the
+# reverse pass exactly when it acquired in the forward pass.  The shadow lock
+# is never touched.
+#
+# The closure-taking `lock(f, l)` / `trylock(f, l)` methods are deliberately
+# not matched, since `f` may be active; they reach these rules through the
+# inner single-argument calls.
+
+function EnzymeRules.forward(
+        config::EnzymeRules.FwdConfig,
+        func::Const{typeof(Base.lock)},
+        ::Type{<:Const},
+        l::Annotation{<:Base.AbstractLock},
+    )
+    func.val(l.val)
+    return nothing
+end
+
+function EnzymeRules.forward(
+        config::EnzymeRules.FwdConfig,
+        func::Const{typeof(Base.unlock)},
+        ::Type{<:Const},
+        l::Annotation{<:Base.AbstractLock},
+    )
+    func.val(l.val)
+    return nothing
+end
+
+function EnzymeRules.forward(
+        config::EnzymeRules.FwdConfig,
+        func::Const{typeof(Base.trylock)},
+        ::Type{<:Const},
+        l::Annotation{<:Base.AbstractLock},
+    )
+    acquired = func.val(l.val)
+    return EnzymeRules.needs_primal(config) ? acquired : nothing
+end
+
+function EnzymeRules.augmented_primal(
+        config::EnzymeRules.RevConfig,
+        func::Const{typeof(Base.lock)},
+        ::Type{<:Const},
+        l::Annotation{<:Base.AbstractLock},
+    )
+    func.val(l.val)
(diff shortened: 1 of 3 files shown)


wsmoses (changes requested):
I don't think this is right. Presumably the derivative of lock is unlock and vice versa?

vchuravy:
:robot: 

Good point, reworked: lock/unlock/trylock now have forward and reverse rules instead of being inactive. The adjoint of `lock` is `unlock`, of `unlock` is `lock`, and of `trylock` is `unlock` iff the forward call acquired. The new tests log every acquire/release through a `RecordingLock` and check the reverse sweep re-takes the locks over the mirrored region. `islocked` stays inactive since it is a pure query.

src/internal_rules/core.jl
```diff
@@ -662,3 +662,113 @@ function EnzymeRules.reverse(
...
+end
+
+function EnzymeRules.augmented_primal(
+        config::EnzymeRules.RevConfig,
+        func::Const{typeof(Base.lock)},
+        ::Type{<:Const},
+        l::Annotation{<:Base.AbstractLock},
+    )
+    func.val(l.val)
+    return EnzymeRules.AugmentedReturn(nothing, nothing, l.val)
```
wsmoses: l.val doesn't need to be taped, its available by definition

src/internal_rules/inactive.jl
```diff
@@ -145,6 +145,14 @@ function EnzymeRules.inactive_noinl(::typeof(Base.fieldnames), args...)
     return nothing
 end
 
+# Querying a lock never mutates it and never carries derivative information.
+# `lock`, `unlock` and `trylock` are not inactive: the reverse sweep of a
+# locked region must itself hold the lock, so they get proper rules in
+# core.jl that run `unlock` as the adjoint of `lock` and vice versa.
+function EnzymeRules.inactive(::typeof(Base.islocked), ::Base.AbstractLock)
```
wsmoses: maybe we do inactive_noinl?

src/internal_rules/core.jl
vchuravy: 🤖 

Done — the tape is gone. `lock`/`unlock` now tape `nothing` and `trylock` tapes only the `Bool` saying whether it acquired; the reverse rules call `unlock(l.val)` / `lock(l.val)` on the annotation argument.

src/internal_rules/inactive.jl
vchuravy: 🤖 

Switched to `inactive_noinl` — `islocked` is a plain load, no reason to keep it from inlining.

PR #3518, opened 2026-09-04
Attach thunk artifacts to CodeInstances so package images skip enzyme-core

vchuravy:
Stacked on #3517. Part of the cache redesign towards session-local state and relocatable, cross-session cacheable compilation.

Draft: the artifact tier works end to end (a package that differentiates during precompilation reloads with zero enzyme-core runs), but native custom rules bake a specptr into the bitcode, so thunks that call them are marked non-persistable and stay session-local for now; making those symbolic is follow-up work.

Attach thunk artifacts to CodeInstances so package images skip enzyme-core

Compiling a thunk was all-or-nothing: `cached_compilation` ran enzyme-core and
JIT-linked the result, and nothing of that survived the process. A package that
differentiated during precompilation shipped an image whose thunk had to be
built again from scratch on first use.

Compilation is now split. `emit` produces a `ThunkArtifact`: the post-optimized
module as bitcode, the entry symbols, the tape type, the edges, the
nested-differentiation module string, and the manifest of Julia values the
bitcode names (with a `persistable` flag). `link_artifact!` turns an artifact
into a `CompileResult` with no compiler involved: it re-registers the manifest's
value references, parses the bitcode and links it into the session's JIT.

On Julia 1.11+ (where CompilerCaching provides the integrated cache; a direct
dependency now, an empty shell on 1.10) a persistable artifact is attached to
the job's primal-partition `CodeInstance`, keyed by the codegen configuration
(`config_key`), through `CompilerCaching.results`. It rides with the
`CodeInstance` into a package image. `cached_compilation` then, on a
`ThunkCache` miss, looks for such an artifact and links it instead of emitting,
so a fresh session (or one after `reset_session!`) starts a precompiled thunk
without running enzyme-core. `EMIT_COUNT` counts the runs that do happen.

Linking reuses an entry symbol already present in the JIT rather than adding it
twice (`JIT.lookup_or_null`), which a within-process relink after a session
reset would otherwise do. An artifact is attached only when its values are
reconstructible and no session address was baked into the bitcode
(`persistable`, and no `enzyme.nonrelocatable` module flag); otherwise it stays
session-local as before. On Julia 1.10 there is no CodeInstance tier and the
`ThunkCache` is the only cache, unchanged.

Test: `test/ci_results.jl` checks that an artifact is attached and reused
without a further emit within a session and across a `reset_session!`, and, in
two child processes sharing a fresh depot, that a package differentiating during
its own precompilation loads in a fresh process and computes the right gradient
with zero emits. Passes with `session_cache`, `thunk_handle`,
`symbolic_lookups`, `julia_value_refs`, `tests`, `basic`, `advanced`,
`core/deferred`, `core/abi`, `rules/rules`, `rules/rule_inlining`, `mixed`,
`ext/jlarrays` on Julia 1.12 with GPUCompiler 2.5 and 1.23, and on Julia 1.10
(where the artifact tier is inactive).

https://claude.ai/code/session_01LdVAXqghGYwHGrnzaFchPT


vchuravy:
Folded into #3517 together with #3520: attaching the artifact to a CodeInstance and holding the entry point in one turned out to be one change.

PR #3520, opened 2026-09-04
Keep FFIABI entry points in CodeInstances

vchuravy:
Stacked on #3519. Part of the cache redesign; this is the first step of moving Enzyme's derivative entry points onto Julia's own representation, so that rule calls and thunk calls share one shape (an entry read from a `CodeInstance`).

Keep FFIABI entry points in CodeInstances

A compiled thunk's entry point was a bare `Ptr{Cvoid}` on the handle, a form
Julia knows nothing about: a caller's codegen could not see a call to it, and it
had to be tracked with Enzyme's own session bookkeeping.

On Julia 1.12+ the entry point is now stored in the `specptr` of a
`CodeInstance` of the synthetic function `enzyme_thunk_entry`, one instance per
thunk, found by a content-derived key (what the thunk was compiled from, its
codegen configuration, and which of the two entry points it is), so the same
thunk maps to the same method instance and instance in any session. This is the
shape Julia gives a natively compiled function and the shape
`bind_native_invokes!` already reads for custom rules. `thunk_pointer` reads the
entry back from the instance. The instance is rooted in its method instance's
cache and is reused across a `reset_session!` with its entry refreshed.

The instance's `invoke` is a real boxed-ABI wrapper (`thunk_invoke_boxed`, a
`jl_callptr_t`): `Core.invoke(enzyme_thunk_entry, ci, Val(key), fn, args...)`
rebuilds the thunk object for the instance and calls it on the boxed arguments,
so the marshalling to the entry's C ABI stays the thunk's own (`enzyme_call`).
The carrier `enzyme_thunk_entry(::Val, args...)` takes the thunk's arguments for
that reason, and the instance is valid from the job's world on rather than from
the world counter, which reads as "no world" inside a generated function's
generator. `specptr` is not a specialized signature of the carrier (bit `0b01`
stays clear), so a caller Julia compiles reaches the entry through `invoke`
only, never by a direct call; Enzyme's own calls read `specptr` as before.

The entry is published the way Julia's JIT publishes one: `specptr` first, then
`invoke`, then the flag that says the two agree (`JL_CI_FLAGS_INVOKE_MATCHES_SPECPTR`, bit `0b10` of `specsigflags` on
1.12 and of `flags` on 1.13). That flag matters: `jl_read_codeinst_invoke`, which
every reader of an entry in the runtime goes through, waits for it whenever it
sees both pointers set, so an instance published without it hangs such a reader
forever (reproduced on 1.13.0-rc4; 1.12 has the same wait).

Every instance is Enzyme-owned (`ThunkEntryOwner`). Julia's runtime only
dispatches to, compiles, or reuses the code of instances under its own owner
(`jl_rettype_inferred_native` and `jl_get_ci_equiv` both compare the owner), so
an instance of Enzyme's is never reached through `enzyme_thunk_entry` itself,
only through the entry Enzyme stored. An earlier version of this change claimed
Julia's owner for `InlineABI` entries; that instance carried an entry with the
wrong ABI for its `invoke`, and on 1.13 `jl_get_ci_equiv` could have bound a
native compile of the carrier to it, so it is gone. The `Core.invoke` direction
for derivatives needs a native-owner instance with a `CodeInfo` body compiled by
Julia, a separate instance from these.

Julia 1.10 and 1.11 keep the pointer on the link as before (`HAS_CI_THUNKS`).
Nothing else changes: `enzyme_call` still receives the entry as a runtime
argument, and the session epoch still decides whether a link belongs to this
session, so a stale entry serialized into a package image is never read without
a fresh link.

Test: `test/thunk_handle.jl` checks that a compiled thunk's link carries a
`CodeInstance` whose `specptr` is the entry and whose `invoke` is set, that
`jl_read_codeinst_invoke` returns that entry with the agreement flag set (rather
than waiting on it), that `Core.invoke` of the instance gives what a call of the
thunk gives, that the content-derived key finds that same instance and
differs between the adjoint and primal entries, that the entry is Enzyme-owned,
and that a session reset reuses the instance and refreshes its entry. Passes
with `ci_results`, `session_cache`, `precompile` on Julia 1.12 and with
`ci_results`, `session_cache` on Julia 1.13.0-rc4.

https://claude.ai/code/session_01WZ9ygd17aLfCESakQx36ot


vchuravy:
Folded into #3517 together with #3518.

PR #3526, opened 2026-09-04
Handle read-only libcalls taking a decayed pointer in fix_decayaddr!

vchuravy:
`fix_decayaddr!` had two handlers for a call consuming an illegal
`addrspace(10) -> addrspace(0)` cast: rebuild `llvm.memcpy` / `memmove` /
`memset` over addrspace 10, or assume the argument is sret-like and copy the
object through a temporary alloca. An ordinary libcall is neither -- its
signature is fixed at `ptr`, so it cannot be re-created over addrspace 10, and
it has no sret -- so it fell through to the sret path, found no marker, and
threw "did not have sret when expected".

This is reachable from plain Julia code: `===` on an immutable over 512 bytes
with no padding lowers to `emit_bits_compare`, which decays both operands
through `julia.pointer_from_objref` and compares them with `memcmp`. LLVM's
`LibCallSimplifier::optimizeMemCmp` then rewrites that to `bcmp`, since the
result feeds nothing but an `icmp eq ..., 0`.

Add a third handler, tried after the memcpy branch so the existing paths keep
every call they used to own. When every argument position holding the decayed
pointer is only read -- the position carries `readonly`/`readnone`, or the
callee as a whole is read-only, which is how `bcmp` arrives
(`memory(argmem: read)`) -- rewrite it to the derivation Julia's late GC
lowering understands:

```llvm
%tok = call token @llvm.julia.gc_preserve_begin(ptr addrspace(10) %obj)
%d   = addrspacecast ptr addrspace(10) %obj to ptr addrspace(11)
%raw = call ptr @julia.pointer_from_objref(ptr addrspace(11) %d)
... call ...
call void @llvm.julia.gc_preserve_end(token %tok)
```

`LateLowerGCPass` lowers `julia.pointer_from_objref` to the addrspacecast the
original cast meant, so the address is unchanged. Any sret-like marker on the
position disqualifies it, and calls that write through the pointer still hit
the assertion.

Also snapshot the distinct users before walking `LLVM.uses(inst)`. The handlers
rewrite all argument positions of a user at once, so detaching a use mid
iteration was already unsound; the new branch makes that shape easier to reach.

### Tests

Three IR-level testsets in `test/passes.jl`: the read-only rewrite, a
non-read-only call still hitting the assertion, and one mirroring what Julia
actually emits (`bcmp`, `memory(argmem: read)`, and a `jl_roots` operand bundle
that must pass through untouched).

### Caveat

The reported failure has not been reproduced standalone -- I could produce the
right IR shape from Julia, but Enzyme differentiated it cleanly without forming
the illegal cast, so something further in the reporting workload collapses
`pointer_from_objref(addrspacecast p10->p11)` into the raw cast. The fix is
derived from the assertion path rather than from a captured reproducer, so it
is worth confirming against the original workload before merging.


wsmoses (approved):
add a test?

vchuravy:
The test is platform dependent and really annoying...

vchuravy:
Might also not sufficient since  fix_decayaddr! only works on intrinsics.

test/passes.jl
```diff
@@ -505,3 +505,31 @@ end # VERSION >= v"1.12"
...
+            """
+        )
+
+        Enzyme.Compiler.annotate!(mod)
+        for name in ("bcmp", "memcmp")
+            fn = LLVM.functions(mod)[name]
+            @test Enzyme.Compiler.has_fn_attr(fn, LLVM.EnumAttribute("nofree"))
+        end
+    end
+end
```
vchuravy: The test should start with a Julia code emission and subsequent optimization

src/llvm/transforms.jl
```diff
@@ -1420,6 +1420,121 @@ function nodecayed_phis!(mod::LLVM.Module)
     return nothing
 end
 
+function is_readonly_attr(attr::LLVM.Attribute)::Bool
```
wsmoses: this already exists in the codebase, we shouldn't copy it

vchuravy: Done in a160e767 — split the attribute loop out of `is_readonly(::LLVM.Function)` as `is_readonly(::LLVM.Attribute)` in `compiler/utils.jl`, and `fix_decayaddr!` now calls that instead of a private copy.

test/passes.jl
vchuravy: Reworked in a160e767. That `annotate!`/`nofreefns` testset is gone; the remaining tests now start from Julia code emission plus optimization, as you asked.

`decay_egal_module()` does:

```julia
@noinline function decay_egal(@nospecialize(a), @nospecialize(b))
    return (a::DecayBig) === (b::DecayBig)   # DecayBig = NTuple{100,Float64}, 800 bytes
end

code_llvm(io, decay_egal, Tuple{Any, Any}; raw=true, optimize=false, dump_module=true)
Enzyme.Compiler.optimize!(mod, Enzyme.Compiler.JIT.get_tm())
```

`@nospecialize` keeps the operands boxed, so `emit_bits_compare` decays two `addrspace(10)` values through `julia.pointer_from_objref` and calls `memcmp`; `optimize!` then rewrites that to `bcmp`. The libcall, its `memory(argmem: read)` / `readonly` attributes, the `jl_roots` bundle and the surrounding code all come out of that, so nothing about the shape is hand-asserted any more, and there is nothing platform- or LLVM-version-specific spelled out in the test (the check accepts either `bcmp` or `memcmp`).

The one thing the test still does itself is collapse `julia.pointer_from_objref(addrspacecast p10 -> p11)` into the direct `p10 -> p0` cast — neither Julia nor `optimize!` forms that here, it is what a later simplification of the two-step derivation leaves behind. That is isolated in `collapse_decay!` and documented as the reconstructed step; it is still the PR's open caveat that I could not get the original workload to produce it end to end. The negative test now drops the read-only markers off that same call instead of declaring its own callee.

Verified the tests fail on the assertion (`did not have sret when expected`) with the new `fix_decayaddr!` branch disabled, and pass with it.

PR #3536, opened 2026-09-07
Sink addrspacecasts above GEPs on tracked pointers before AD

### LINKED ISSUE #3532: GC segfault error with mutable structs and runtime activity in Julia 1.11
The following MWE gives a GC segfault error in Julia 1.11 (but not 1.10 or 1.12)

```Julia 
# Minimal reproducer of the Julia 1.11 GC segfault in Enzyme reverse-mode AD (see
# enzyme-julia-1-12-differentiability.md, section "the Julia 1.11 segfault"). Pure Enzyme.jl, no KernelAbstractions.
# Julia 1.11.9 + Enzyme 0.13.200: segfault in `gc_mark_outrefs`. Julia 1.10.11 and 1.12.6: exact gradient 2*sum(x).
#
# Ingredients, all needed: (1) a scalar `dt` read from a mutable struct reached through a pointer field of the
# (mutable) model, needed again in the reverse pass; (2) a non-inlined function, so Enzyme differentiates it in split
# mode with a tape; (3) `set_runtime_activity`, which makes Enzyme keep the primal field address `&m.ts.dt` in the
# tape instead of caching the value — on 1.11 that address is a GEP on a GC-tracked pointer, and Enzyme.jl's tape
# typing hands it to the GC as an object; (4) a GC while the tape is live. The GC "claims" the interior address,
# sets a mark bit in the preceding field `ms`, then faults dereferencing `ms` (10_800_000) as a type tag.
# Plain `Reverse` instead of `set_runtime_activity(Reverse)` passes; inlining `step!` passes (combined mode
# re-executes the load in the reverse pass). The same happens with a KernelAbstractions kernel or a custom rule
# in place of the loop, which is how SpeedyWeather hits it.
using Enzyme
mutable struct Stepper; ms::Int; dt::Float32; end
mutable struct Model; x::Vector{Float32}; ts::Stepper; end
@noinline function step!(y, m::Model)
    (; dt) = m.ts
    x = m.x
    for i in eachindex(y)
        y[i] += dt * 2f0 * x[i] + 1f0
    end
    return nothing
end
function run!(y, m::Model)
    step!(y, m)
    GC.gc()
    return nothing
end
n = 128; m = Model(rand(Float32, n), Stepper(10_800_000, 10800f0)); dm = make_zero(m)
y = zeros(Float32, n); dy = ones(Float32, n)
println("Julia $VERSION  Enzyme $(pkgversion(Enzyme)) — before: ms = ", m.ts.ms); flush(stdout)
autodiff(set_runtime_activity(Reverse), run!, Const, Duplicated(y, dy), Duplicated(m, dm))
println("after: ms = ", m.ts.ms, "  d(dt) = ", dm.ts.dt, " (expect 2*sum(x) = ", 2 * sum(m.x), ")")
println("RESULT: SUCCESS")
```

vchuravy:
Fixes #3532.

## Root cause

Julia forms a field address as an `addrspacecast` from Tracked (addrspace 10) to Derived (addrspace 11) followed by a GEP. With typed pointers (LLVM <= 16, so Julia <= 1.11) InstCombine rewrites `gep(addrspacecast(x))` into `addrspacecast(gep(x))`, which leaves the interior pointer `&m.ts.dt` in addrspace 10:

```llvm
%18 = load atomic i8 addrspace(10)*, ...            ; m.ts
%19 = getelementptr inbounds i8, i8 addrspace(10)* %18, i64 8
%20 = bitcast i8 addrspace(10)* %19 to float addrspace(10)*
%21 = addrspacecast float addrspace(10)* %20 to float addrspace(11)*
%22 = load float, float addrspace(11)* %21
```

Julia's own GC lowering tolerates this. But with runtime activity Enzyme's cache selection picks `%20` for the tape of the split-mode `step!`, and `tape_type` has to treat every addrspace(10) pointer as a Julia object. The GC is then handed a pointer into the middle of the `Stepper`, sets a mark bit in the preceding `ms` field and faults reading it as a type tag. On 1.12 opaque pointers keep the cast above the GEP, so Enzyme caches the `m.ts` object and recomputes the address in the reverse pass.

The same InstCombine fold also hoists the cast out of phis and selects, so a join of field addresses becomes a phi or select of addrspace(10) interior pointers. `nodecayed_phis!` only rewrites addrspace(11) and (13) phis, so that shape was not covered either.

## Fix

A new pre-AD pass, `rederive_tracked_geps!`, rebuilds every GEP on a Tracked pointer as addrspacecast-then-GEP in addrspace 11, follows chained bitcasts and GEPs, and redirects loads and stores through them. A phi or select fed by a tracked interior pointer is rebuilt in addrspace 11 as well, casting whole-object inputs at the incoming edge. Any other user keeps the Tracked value, so the rewrite is always sound and merely best effort. Enzyme core only runs InstCombine under auto-sparsity, so nothing re-sinks the cast before the cache decisions.

The pass runs right before `nodecayed_phis!`, which then roots the resulting addrspace(11) phis as a whole-object phi plus a byte offset, as it does for any other decayed phi. The two are exposed to the pipeline as one entry point, `canonicalize_gc_pointers!`: the GEP pass establishes that every addrspace(10) value is a whole object, the phi pass that every cached pointer is rooted. They are kept as separate walkers because one is structural and the other accumulates offsets (and owns the addrspace 13 / `gc_loaded` handling), so neither can absorb the other.

With the fix the 1.11 tape holds `i8 addrspace(10)*` (the `m.ts` object) where it held `float addrspace(10)*`, and no addrspace(10) GEPs remain in the module.

## Testing

- The reproducer from the issue segfaults on stock 1.11.9 and prints the correct gradient with this change on 1.10, 1.11 and 1.12.
- New FileCheck tests in `test/passes.jl`: the GEP chain from the issue, and a diamond phi, a pointer-increment loop phi and a select of interior pointers, checked both after `rederive_tracked_geps!` alone and after `nodecayed_phis!` has rooted them. Pass on 1.11 (typed pointers) and 1.12 (opaque pointers).
- New regression testset at the end of `test/basic.jl`.
- `mixed`, `typeunstable`, `rooted_args`, `applyiter`, `array` and `runtime_calls` pass locally on 1.11 with the pass enabled.


wsmoses:
this pass already exists, its the fix_decayaddr?

vchuravy:
`fix_decayaddr` is running after AD whereas this runs before AD. 

Enzyme is caching an interior pointer on it's tape due to instcombine doing `gep(addrspacecast(x))` into `addrspacecast(gep(x))`

wsmoses:
https://github.com/EnzymeAD/Enzyme.jl/blob/c5f5b55d3076b02bdb8661fd8c0d35d1e395e759/src/compiler/optimize.jl#L223 defs comes before?

vchuravy:
Are `fix_decayaddr` and `nodecayed_phis!` the same?

wsmoses:
Sorry I specifically meant the latter (they are two different things I forgot the right name)

vchuravy:
Ah I see they are similar but attack the problem from the opposite direction, I think we need both, but `rederive_tracked_geps` needs to run first to seed `nodecayed_phis`.

vchuravy:
@wsmoses can you have another look at this?

src/llvm/transforms.jl
```diff
@@ -1522,6 +1477,335 @@ function legalize_readonly_decay!(
     return nothing
 end
 
+function is_undef_or_poison(@nospecialize(v::LLVM.Value))::Bool
+    isa(v, LLVM.UndefValue) && return true
+    @static if LLVM.version() >= v"12"
+        isa(v, LLVM.PoisonValue) && return true
+    end
+    return false
+end
```
vchuravy, line 1486: Move to utils.jl

src/llvm/transforms.jl
```diff
@@ -1522,6 +1477,335 @@ function legalize_readonly_decay!(
...
+# one object share a cast and no cast is stranded inside a loop that `v` is
+# invariant of. Returns `nothing` when there is no such point, i.e. when `v` is
+# defined by a terminator, and the caller then casts at the use instead.
+function derived_base!(st::RederiveState, @nospecialize(v::LLVM.Value))
+    haskey(st.bases, v) && return st.bases[v]
+    at = if isa(v, LLVM.Instruction)
+        LLVM.API.LLVMGetNextInstruction(v)
+    else
+        LLVM.API.LLVMGetFirstInstruction(first(blocks(st.f)))
+    end
```
vchuravy, line 1546: is there an LLVM.jl API for this?

PR #3543, opened 2026-09-08
Keep parameter types in a side table keyed by an opaque id

vchuravy:
Stacked on #3156 (which is stacked on #2643). Draft: the first step of a change that will take a few rounds.

## Why

Every function Enzyme emits carries its Julia signature as *process addresses in string attributes*: `enzymejl_mi` and `enzymejl_rt` on the function, `enzymejl_parmtype` (plus `_ref` and `_rooted_typ`) on each parameter and on the return value. They are read back with `unsafe_pointer_to_objref` at compile time, by `enzyme_custom_extract_mi`, `enzyme_extract_parm_type`, the type-analysis setup in `set_module_types!`, and the error paths.

Inside one session that is fine. It stops being fine once modules outlive the session: the cache stack (#3517, #3521, #3522) keeps a thunk's compiled artifact on its `CodeInstance`, together with the pre-optimization module string that nested differentiation splices, and a package image carries both. `bakes_addresses` only inspects `inttoptr` constants, so a module whose only process addresses live in these attributes is judged session-portable, and nested differentiation through a thunk linked from an image would decode dead addresses. Nothing in the suite exercises that path yet; it is a latent crash, not a failing test.

## Plan

1. **Side table** (this PR): `EnzymeContext` already gained `mi_cache` (function name to method instance and return type) in #3156. It now also carries `param_types`, a table from an opaque id to the Julia type, calling convention and rooted type of one parameter (0 for the return value). Every writer records into it as it writes the address attributes: `prepare_llvm` for compiled functions, the ABI wrappers in `lower_convention`, and the native rule declarations in `callconv.jl` (which now take the context). The parameter is tagged with its id in an `enzymejl_parm_id` attribute, and that id is the key. A name-and-index key was tried first and broke as soon as a pass removed a dead argument (the kwsorter of `stdm` loses its `Bool`, and index 1 becomes the array); an attribute moves with the parameter through renames, argument removal, the attribute copies onto wrappers and module links. Ids come from a process-wide counter, so an entry can never be claimed by a parameter recorded in another context. `enzyme_extract_parm_type(enzyme_context, fn, idx)` prefers the table and falls back to the attributes; `absint` and `julia_activity_rule` use it. Behaviour is unchanged when the table misses.
2. Cover the remaining readers: the `set_module_types!` loop that pairs `enzymejl_mi`/`enzymejl_rt` back into `handle_compiled`, the `enzymejl_rooted_typ` reads, and the `enzymejl_gc_alloc_rt` metadata on allocations.
3. Stop writing addresses. Keep the `_str` attributes for debugging, and have `bakes_addresses` (on the cache stack) reject any surviving address-valued attribute so an artifact can never be marked portable with one.
4. On the cache stack: serialize the table with the artifact (method instances and types are ordinary Julia values, so it rides in `analysis_results` like the rest), and merge it into the context when an artifact is linked or spliced.

`mi_cache` from #3156 is still keyed by function name; it should move to the same id scheme (an `enzymejl_fn_id` attribute) in step 2, since a name key has exactly the fragility the parameter table just lost.

## Test

`typeunstable`, `errors`, `basic`, `rules/rules`, `absint`, `rules/rrules`, `rules/rule_inlining`, `mixed` on Julia 1.12 pass. The `stdm` case above is covered by `basic` (the Statistics tests), which is what exposed the index shift.

https://claude.ai/code/session_01WZ9ygd17aLfCESakQx36ot

---

**2026-09-08, later:** rebased onto #3156 after #2643 removed `enzymejl_world` and threaded `world::UInt` explicitly; no design change here. Same test set passes on Julia 1.12.

### DIFF
--- a/src/absint.jl
+++ b/src/absint.jl
@@ -1034,7 +1034,8 @@ function abs_typeof(
     if isa(arg, LLVM.Argument)
         f = LLVM.Function(LLVM.API.LLVMGetParamParent(arg))
         idx = only([i for (i, v) in enumerate(LLVM.parameters(f)) if v == arg])
-        typ, byref = enzyme_extract_parm_type(f, idx, false) #=error=#
+        typ, byref = enzyme_context === nothing ? enzyme_extract_parm_type(f, idx, false) :
+            enzyme_extract_parm_type(enzyme_context, f, idx, false) #=error=#
         if typ !== nothing
             return (true, typ, byref)
         end
--- a/src/compiler.jl
+++ b/src/compiler.jl
@@ -1304,6 +1304,7 @@ function set_module_types!(enzyme_context::EnzymeContext, interp, mod::LLVM.Modu
                     parameter_attributes(f, arg.codegen.i),
                     StringAttribute("enzymejl_parmtype_ref", string(UInt(arg.cc))),
                 )
+                record_param_type!(enzyme_context, f, arg.codegen.i, arg.typ, arg.cc, arg.rooted_typ)
 		if arg.rooted_typ !== nothing
 			push!(
 			    parameter_attributes(f, arg.codegen.i),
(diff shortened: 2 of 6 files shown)


src/compiler/callconv.jl
```diff
@@ -262,6 +262,7 @@ end
         fattrs = function_attributes(fn)
         push!(fattrs, StringAttribute("enzymejl_mi", string(convert(UInt, pointer_from_objref(mi)))))
         push!(fattrs, StringAttribute("enzymejl_rt", string(convert(UInt, unsafe_to_pointer(RT)))))
```
vchuravy, line 264: Remove the attributes now,

src/compiler/callconv.jl
```diff
@@ -351,15 +353,15 @@ end
     end
 
     """
-        declare_native!(mod, mi, RT, specptr, name, world)
+            declare_native!(enzyme_context, mod, mi, RT, specptr, name, world)
```
vchuravy: The formatting got worse here.

src/absint.jl
```diff
@@ -1034,7 +1034,8 @@ function abs_typeof(
     if isa(arg, LLVM.Argument)
         f = LLVM.Function(LLVM.API.LLVMGetParamParent(arg))
         idx = only([i for (i, v) in enumerate(LLVM.parameters(f)) if v == arg])
-        typ, byref = enzyme_extract_parm_type(f, idx, false) #=error=#
+        typ, byref = enzyme_context === nothing ? enzyme_extract_parm_type(f, idx, false) :
+            enzyme_extract_parm_type(enzyme_context, f, idx, false) #=error=#
```
vchuravy, line 1038: We must ensure that enzyme_context is always available.

src/compiler.jl
```diff
@@ -2679,6 +2680,44 @@ function enzyme_custom_extract_mi(orig::LLVM.Function, error::Bool = true)
...
+
+# Record the Julia type, calling convention and rooted type of parameter `idx` of `fn`
+# (0 for the return value) in the context's side table, and tag the parameter with the
+# table key.
+function record_param_type!(enzyme_context::EnzymeContext, fn::LLVM.Function, idx::Int, @nospecialize(typ), cc, @nospecialize(rooted))
+    id = Threads.atomic_add!(PARM_ID, UInt64(1)) + 1
+    enzyme_context.param_types[id] = (typ, UInt(cc), rooted)
+    push!(parm_attributes(fn, idx), StringAttribute("enzymejl_parm_id", string(id)))
+    return nothing
+end
```
vchuravy, line 2699: This seems the wrong direction. We should have a mi/ci based side table.

PR #3547, opened 2026-09-09
Skip addrspace(11) argument phis reached through variable GEPs in nodecayed_phis!

vchuravy:
Fixes the Julia 1.13 SciML and Molly integration failures seen on #3545 (same bug in both).

`nodecayed_phis!` rewrites every addrspace(11) pointer phi into an addrspace(10) base plus an offset so Enzyme can root the cached value. A phi whose incoming values all derive from addrspace(11) arguments has no such base and was already skipped, but only when `get_base_and_offset` could see through the GEPs, i.e. constant indices with identical offsets.

LLVM 20 (Julia 1.13) strength-reduces loop indices into pointer induction variables:

```llvm
%invariant.gep = getelementptr i8, ptr addrspace(11) %arg, i64 -8
%p = phi ptr addrspace(11) [ %arg, %top ], [ %gep, %latch ]
%gep = getelementptr double, ptr addrspace(11) %invariant.gep, i64 %iv
```

The variable-index GEP stopped the base walk, the phi was rewritten, and `nodecayed_getparent` failed with "Could not analyze garbage collection behavior" on the argument. This breaks any loop over an SVector or a `zip` of tuples passed by reference, e.g. `OrdinaryDiffEqCore.initialize_saveat` (SciML integration test) and Molly's periodic torsion force (`_foldl_impl` over an `Iterators.Zip` of NTuples).

**Fix:** the two detection loops are replaced by `nodecayed_all_args`, which walks bases through GEPs of any index shape, phis and selects, and skips the phi whenever every root is an addrspace(11) argument, undef or poison. The offset no longer matters for that decision, so a phi of `%arg` and `%arg + 8` is now skipped as well instead of erroring.

**Tests:** an IR-level test in `test/passes.jl` (variable GEP and select cases, runs on every Julia version) and the two reduced Julia cases in `test/rooted_args.jl`, which fail on 1.13 without the fix.

**Verified locally on 1.13.0-rc4:** the reduced cases and the real `initialize_saveat` fail before and pass after; the full SciML (3 testsets) and Molly (4 testsets, 36 tests) integration suites pass with the fix. `test/passes.jl` + `test/rooted_args.jl` pass on 1.10.12, 1.12.7 and 1.13.0-rc4.

https://claude.ai/code/session_014ggwnn7AEg7YdTaeHdDo71


jgreener64:
Thanks for this.

PR #3552, opened 2026-09-09
Make EnzymeContext a scoped value

vchuravy:
The context of a compilation was passed explicitly through `nested_codegen!`, `invoke_codegen!`, `materialize_native_invokes!`, `enzyme!` and `create_abi_wrapper`. The rule handlers, which Enzyme calls back into from the middle of the pipeline where no such argument reaches, recovered it from a raw pointer stashed in the C++ `EnzymeLogic` (`EnzymeLogicSetExternalContext`, #2624).

Bind it to a `ScopedValue` in `compile_unhooked` instead:

```julia
const ENZYME_CONTEXT = ScopedValue{EnzymeContext}()

function GPUCompiler.compile_unhooked(output::Symbol, job::CompilerJob{<:EnzymeTarget})
    return @with ENZYME_CONTEXT => EnzymeContext() begin
        compile_unhooked_impl(output, job)
    end
end
```

Everything running under that compilation reads it with `enzyme_context()`, so both the parameter and the pointer round trip go away, and a rule handler is down to a plain `invoke_codegen!(mode, mod, mi, world, true)`. A nested compilation binds its own context and hands the outer one back when it returns, exactly as the explicit threading did.

Consequences:

* `Logic` no longer carries a context, so it needs neither `EnzymeLogicSetExternalContext`/`EnzymeLogicGetExternalContext` nor a `GC.@preserve` to keep the object alive across the pass.
* `compile_unhooked` is split into a wrapper that binds the context and `compile_unhooked_impl` holding the body, so the body is not re-indented by the `@with` block.
* Reaching the context outside a compilation is now an error with a message rather than an assert on a NULL pointer. A rule reached from a `Logic` that Enzyme.jl did not create used to trip that assert; it now works as long as it runs inside the compilation's dynamic scope.
* `ScopedValues` is a new dependency, because `Base.ScopedValues` only exists on 1.11+. On 1.11 and later the package re-exports it.

The lower bound is 1.6.2, not 1.0: the backport that Julia 1.10 uses got a scoped value with *no default* wrong twice. Before 1.6.1 `Base.isassigned` reports `false` inside the very `@with` block that bound the value; 1.6.1 fixed that but left `get` calling `isassigned` where it means "has a default", so it evaluates the unassigned `default` field on the path where the value is bound — harmless for an isbits value, an `UndefRefError` for a reference type such as `EnzymeContext`. Both showed up only in the Julia 1.10 Downgrade job, which pins the lowest allowed version.

Compared to the LLVM.jl style of implicit context (a `task_local_storage(:LLVMContext)` stack with `activate`/`deactivate`): an `LLVM.Context` has object lifetime and may be created in one function and disposed in another, so activation has to hang off construction and disposal there. An `EnzymeContext` has call lifetime -- one `compile_unhooked`, plain GC'd Julia state with nothing to dispose -- so a lexical scope expresses the same invariant without the risk of an unbalanced `deactivate` pinning it in task-local storage, it infers as `EnzymeContext` rather than `Any`, and it propagates into spawned tasks. Lookup cost measures the same for both (~40ns, once per `nested_codegen!`).

Tested with `rules`, `core`, `callconv`, `threads` and `runtime_calls`: 14264 pass, 3 broken, 0 fail. Those cover the custom rules (`invoke_codegen!`), deferred and nested compilation, the native invoke path (`materialize_native_invokes!`) and the pfor rules (`nested_codegen!`).

https://claude.ai/code/session_013MPpgeyjKsFTLvMpBM38vA


vchuravy:
The only thing we would need to be careful with is, threads created from C++ like https://github.com/EnzymeAD/Enzyme/pull/3062

wsmoses:
why?

vchuravy:
> why?

Why this change, or why would we need to be careful?

ScopedValues are attached to Julia tasks and their children. A new C++ thread will not inherit them.

vchuravy:
:robot: 
CI note, so a red CUDA job on this PR isn't mistaken for a regression: `buildkite/enzyme-dot-jl/cuda-julia-v1-dot-10` failed once on `8e377a1e` and **passed on retry of that same job on that same commit** ([build 8054](https://buildkite.com/julialang/enzyme-dot-jl/builds/8054)).

It was not a test failure but a GC abort:

```
[435277] signal (6.-6): Aborted
in expression starting at test/cuda.jl:318
gc_dump_queue_and_abort at src/gc.c:1815
gc_mark_outrefs -> gc_mark_loop -> _jl_gc_collect -> maybe_collect -> ijl_gc_pool_alloc
LLVMType at LLVM/src/core/type.jl:49
compile_unhooked_impl at src/compiler.jl:6102
tape_type -> CUDACore EnzymeCoreExt augmented_primal:321
diffejulia_square__101321wrap at test/cuda.jl:0
```

A CUDA `augmented_primal` rule runs at run time from Enzyme-generated code, calls `tape_type`, and a GC triggered by an allocation inside that nested compilation walks a bad object. That is the `get_pgcstack` / `get_pgcstack_or_new` frame problem #3541 is about, not anything in this PR: same failure shape, nondeterministic, and it clears on retry.

Worth stating the one way this PR touches that area, since it is not zero: entering the `@with` block writes task state (the ScopedValues backport swaps `task.logstate` on 1.10; 1.11+ sets the task's scope), where before this PR a nested compilation entered from such a frame only passed an argument. If the task pointer in that frame were wrong, though, the first nested compile would corrupt memory every time rather than once in twelve builds, and the 1.11/1.12 CUDA jobs use the same mechanism and pass. So: exposure to allocation timing, not causation.

PR #3553, opened 2026-09-09
Take the world from the EnzymeContext, not from an LLVM attribute

vchuravy:
_Requested by **Valentin** · [project thread](https://claude.ai/code/project/chan_014xRteBwvsoFDvuaGmJ7Uws?thread=cmsg_014xRteBwvsoFDvuaGmJ7UwsHUD71ZJxvPi9NtPfzLJusn)_

Alternative to #2643. Same destination — the world age lives in `EnzymeContext` and `enzymejl_world` is gone — reached a different way. Builds on #3552, which has merged; this now sits directly on `main`.

## The difference

Both PRs have to answer the same question: a rule handler, an activity rule or a type-analysis error path runs as a callback from the middle of Enzyme's C++ pipeline; how does it find the context?

**#2643** reaches it through the C++ objects the callbacks are handed. `Logic` keeps a pointer to the context (`EnzymeLogicSetExternalContext`), rule handlers recover it with `EnzymeGradientUtilsGetExternalContext(gutils)`, type-analysis handlers with `EnzymeTypeAnalyzerGetLogic(data)`, and, per its description, a callback that is handed neither publishes the world in task-local storage around the C++ call. Every reader needs a handle to look up from, so helpers like `byref_from_val_if_mixed` grow a `gutils` argument to carry one down.

**This PR** takes the `EnzymeContext` that #3552 bound to a `ScopedValue` in `compile_unhooked`, and puts the world in it. A callback reads `enzyme_world()`: no handle, no threaded argument, no C API, and no task-local seam to remove later. `EnzymeLogicSetExternalContext` and the pointer round trip went away in #3552, and the rule handlers here lose arguments rather than gaining them.

## What this commit does

* `EnzymeContext` gains `world`, bound from `job.world` in `compile_unhooked`.
* `enzyme_world()` replaces `enzyme_extract_world(fn)` at all 40 call sites (rules, activity rules, `set_module_types!`, `shadow_alloc_rewrite`, the `julia_error` branches, `emit_methodinstance!`).
* `enzymejl_world` is written nowhere: the `prepare_llvm` loop, the `declare_native!` push, the two attribute copies onto Enzyme.jl's wrappers, and the `wrapper_f` push are all deleted, along with `enzyme_extract_world` and the attribute scan in `unsafe_to_llvm`. enzyme-core's copy of the attribute in `Utils.cpp` becomes dead and can be removed there.
* `nested_codegen!`, `invoke_codegen!`, `materialize_native_invokes!` and `create_abi_wrapper` no longer take a `world` — every caller was passing the context's.
* `enzyme_world_if_active()` returns `nothing` outside a compilation, for the three places that may run after a compilation has already returned its module or that must not fail while building an error message: the `nodecayed_phis!` error paths (`test/passes.jl` calls that pass directly on parsed IR), `sret_ty`, and `unsafe_to_llvm`, which already treated the world as optional.

One behaviour change worth calling out: `unsafe_to_llvm` used to find no world whenever the enclosing function carried no attribute — every function enzyme-core created without the copy — and then skipped the `guaranteed_nonactive` marking in `setup_global`. It now has the compilation's world in those cases, so that marking runs where it silently did not before. That is the point of the change, but it is the one place where the result differs rather than the plumbing.

The `"enzymejl_world"` strings left in `test/passes.jl` are IR fixtures; the passes under test do not read them.

## Testing

Full local suite on Julia 1.12: 218180 pass, 60 broken, 0 fail.

CI on this head is green everywhere except two integration jobs: the whole test matrix (1.10–1.13 across Linux, macOS and Windows, both libEnzyme flavours, assertions on and off), EnzymeCore, EnzymeTestUtils, Downgrade, Documentation and runic all pass.

The two red jobs, Molly 1.13 and SciML 1.13, fail identically on `main` at this PR's base commit and are the `nodecayed_phis!` addrspace(11) bug that #3547 fixes; they are not this PR's. Lux 1.13 also segfaulted once on this head, in `jl_as_global_root` during an arg-tuple instantiation, and passed on a re-run of the same commit.


src/Enzyme.jl
wsmoses, line 174: @vchuravy I'm sketched out by this. When did this become a scoped value -- it shouldn't be, I thought we were threading things through

Issue #2464, opened 2025-07-13
Enzyme crashes when differentiating nested Turing model inference calls


PTWaade:
Dear Enzyme,

As always, thanks for your amazing work!

My Julia instance crashes when I use Enzyme as an AD backend when trying to do "meta-Bayesian" modelling with Turing (i.e. to put a Turing sample() call within another Turing model).
Getting this to work could be a great contribution to the world of cognitive modelling, where this type of model is common, but must usually be hand-specified.
Below I give a very minimal example; see [https://github.com/TuringLang/Turing.jl/issues/2615](https://github.com/TuringLang/Turing.jl/issues/2615) for a more involved example and some background.

When I use Enzyme's Forward mode, I get an error (call_type is not defined). 
When I use Enzyme's Reverse mode, my Julia instance crashes completely. 

Let me know if there is anything further I can do!

## MWE
```
using Turing
using LogExpFunctions
using ADTypes: AutoEnzyme
import Enzyme: set_runtime_activity, Forward, Reverse

#### CREATING INNER AND OUTER MODELS ####

# Inner model function
@model function inner_model(observation, prior_μ = 0, prior_σ = 1)

    # The innter model's prior
    prob ~ LogitNormal(prior_μ, prior_σ)

    #The inner model's likelihood
    observation ~ Bernoulli(prob)
end

#Checking that inner model works
sample(inner_model(1), NUTS(), 1000)


# Outer model function
@model function outer_model(observation, action, inner_sampler = NUTS(), inner_n_samples = 1000)
    
    ### Sample parameters for the inner inference and response ###

    #Parameters of the inner model's prior
    subj_prior_μ ~ Normal(0, 1)
    subj_prior_σ ~ Exponential(1)
    
    #Inverse temperature for actions
    β ~ Exponential(1) 


    ### "Perceptual inference": running the inner model ###

    #Condition the inner model
    inner_m = inner_model(observation, subj_prior_μ, subj_prior_σ)

    #Run the inner Bayesian inference
    chns = sample(inner_m, inner_sampler, inner_n_samples, progress = false)

    #Extract the subject's point estimate of the reward probability
    subj_mean_expectationₜ = mean(chns[:prob])


    ### "Response model": picking an action ###

    #Transform the point estimate with the inverse temperature parameter
    action_probability = logistic(subj_mean_expectationₜ * β)

    #Sample the action
    action ~ Bernoulli(action_probability)

end


[...]

PR #3573, opened 2026-09-11
Tell the use analysis that the getfield rules need their object shadow

vchuravy:
Part of #2464.

`common_jl_getfield_rev` calls `invert_pointer` on the object it reads a field from, then `lookup_value` on the result. That materializes the object's shadow and builds a scope cache for it, so the shadow is genuinely consumed by the reverse pass.

Enzyme's differential use analysis cannot see inside an LLVM rule. We register exactly one DiffUse handler, `enzyme_custom`; every other rule has augmented/reverse/forward handlers and no DiffUse handler, so enzyme-core falls back to its default reasoning. For this shape it concludes the object's shadow is dead, and `visitLoadLike` erases the placeholder the rule has already consumed:

```
Erased value with a use
```

This registers a DiffUse handler for `jl_f_getfield` / `ijl_f_getfield`, and for `julia.call` / `julia.call2` wrapping them, reporting the object operand's shadow as needed whenever the rule would invert it. Everything else returns `useDefault`, so the default logic is unchanged.

One detail worth knowing for future handlers: enzyme-core looks DiffUse handlers up via `getFuncNameFromCall`, which returns the *called operand* — `julia.call` — and not the inner callee. A handler registered only under `jl_f_getfield` never fires for the wrapped form, which is the form Julia actually emits here.

### Effect

This is what blocked differentiating a Turing model that calls `sample` internally, through the chain indexing in `chn[:prob]`. Reverse mode on the reproducer from #2464 now completes.

### Testing

Verified against released Enzyme_jll 0.0.293, no enzyme-core change needed:

- the reduced reproducer returns the exact analytic gradient in both forward and reverse
- the full #2464 reproducer completes in reverse and matches central finite differences to 0.2% (the Metropolis-Hastings accept/reject makes it slightly non-smooth, so the RNG has to be seeded per evaluation)
- `test/typeunstable.jl`, `test/basic.jl`, `test/sugar.jl`, `test/mixed.jl`, `test/applyiter.jl`, `test/runtime_calls.jl` and `test/usermixed.jl` all pass

No regression test, and I would rather say so than bury it. Twelve synthetic variants all pass on an unpatched build: plain vectors, `OrderedDict` keysets, loop-carried phis, `Any`-typed getfield results, `Const` and `Duplicated` containers. Whether the object's shadow is materialized at all depends on the surrounding shadow ordering, which is what makes it hard to provoke in a small function. I also considered `test/integration/Turing`, but it pins `Turing = "=0.42.1"`, which predates the chain type whose indexing triggers this, so a test added there probably would not exercise the path. Happy to keep trying, or to bump that pin separately, if you would rather this not land untested.

Forward mode on the full #2464 reproducer additionally needs EnzymeAD/Enzyme#3213; reverse mode needs only this change.

https://claude.ai/code/session_01QcTbLNBXz7VEHFHMXHcFdp

### DIFF
--- a/src/rules/llvmrules.jl
+++ b/src/rules/llvmrules.jl
@@ -244,6 +244,16 @@ include("parallelrules.jl")
     return false
 end
 
+@register_diffuse function jlcall_diffuse(orig::LLVM.CallInst, gutils::GradientUtils, @nospecialize(val::LLVM.Value), isshadow::Bool, mode::API.CDerivativeMode)
+    F = operands(orig)[1]
+    if isa(F, LLVM.Function)
+        if in(LLVM.name(F), ("ijl_f_getfield", "jl_f_getfield"))
+            return common_jl_getfield_diffuse(2, orig, gutils, val, isshadow, mode)
+        end
+    end
+    return (false, true)
+end
+
 @register_aug function jlcall_augfwd(B, orig, gutils, normalR, shadowR, tapeR)
     F = operands(orig)[1]
     if isa(F, LLVM.Function)
@@ -2286,6 +2296,12 @@ end
         "enzyme_custom",
         @diffusefunc(enzyme_custom_diffuse)
     )
+    for nm in ("julia.call", "julia.call2")
+        API.EnzymeRegisterDiffUseCallHandler(nm, @diffusefunc(jlcall_diffuse))
+    end
+    for nm in ("jl_f_getfield", "ijl_f_getfield")
+        API.EnzymeRegisterDiffUseCallHandler(nm, @diffusefunc(jl_getfield_diffuse))
+    end
     register_handler!(
         ("julia.call",),
         @augfunc(jlcall_augfwd),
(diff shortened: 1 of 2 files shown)


src/rules/llvmrules.jl
```diff
@@ -244,6 +244,16 @@ include("parallelrules.jl")
     return false
 end
 
+@register_diffuse function jlcall_diffuse(orig::LLVM.CallInst, gutils::GradientUtils, @nospecialize(val::LLVM.Value), isshadow::Bool, mode::API.CDerivativeMode)
+    F = operands(orig)[1]
+    if isa(F, LLVM.Function)
+        if in(LLVM.name(F), ("ijl_f_getfield", "jl_f_getfield"))
+            return common_jl_getfield_diffuse(2, orig, gutils, val, isshadow, mode)
+        end
+    end
```
vchuravy, line 253: Can we generalize this so that when we see other functions that might require diffuse we call the right one?

PR #3577, opened 2026-09-13
Mark integer atomics inactive so 1.13 does not abort on atomicmodify

vchuravy:
First of two stacked PRs fixing `ext/jlarrays` on Julia 1.13. This one is the fatal half: it aborts the test worker, so the whole CI job fails with no test summary. #3578 is stacked on top of it.

## The abort

`Threads.atomic_add!` and friends stopped being an `llvmcall` in Julia 1.13 and became `(@atomic :acquire_release x.value + v).first`, which emits the **variadic** `julia.atomicmodify` intrinsic:

```llvm
%22 = call { i64, i64 } (ptr addrspace(11), ptr, i8, i8, ...)
      @julia.atomicmodify.i64.p11(ptr addrspace(11) align 8 %21, ptr nonnull @0, i8 6, i8 1, i64 1)
```

Enzyme's C++ side treats that declaration as an ordinary subfunction to recurse into, and a variadic callee's `arg_size()` never matches the call's operand count:

```
number of arg operands != function parameters
julia: EnzymeLogic.cpp:2011: const AugmentedReturn& EnzymeLogic::CreateAugmentedPrimal(...):
Assertion `nowrite_shadows.size() == todiff->arg_size()' failed.

[3877] signal 6 (-6): Aborted
CreateAugmentedPrimal at EnzymeLogic.cpp:2011
recursivelyHandleSubfunction at AdjointGenerator.h:5793
```

`test/ext/jlarrays.jl` reaches it through GPUArrays `DataRef` refcounting: `copy(::DataRef)` calls `retain`, which bumps a `Threads.Atomic{Int}`.

## The workaround

An atomic over an integer carries no derivative, so the read-modify-write has nothing to differentiate in the first place. An `EnzymeRules.inactive` rule says so, and as a side effect blocks the inlining that would otherwise bury the intrinsic in a caller — so Enzyme never sees it.

Atomics over floating-point values are deliberately not covered: those do accumulate derivatives and need real handling, not an inactive marker.

**This is a workaround, not the fix.** Enzyme should not recurse into variadic Julia intrinsics at all. Filing that separately; this unblocks CI in the meantime.

## Testing

New `integer atomic is inactive` testset in `test/threads.jl` — it aborts on 1.13 without the rule and passes with it. `test/threads.jl` is green on 1.10, 1.11, 1.12 and 1.13; `test/basic.jl` is clean on 1.13.

Nothing in the test suite used `Threads.Atomic` before this, so the rule had no existing coverage either way.

https://claude.ai/code/session_019SB3Y5mNfYxDTmBkZdun2W


wsmoses (commented):
@vchuravy this is the wrong fix can you revert, just mark the intrinsic as inactive or type analysis info on enzyme side

vchuravy:
You don't have to wait for me to open a revert, if I am to aggressive in merging. 

The full fix would be https://github.com/EnzymeAD/Enzyme/pull/2880 which is waiting on a review. The problem is that the Julia intrinsic is not inactive for float data?

Issue #3551, opened 2026-09-09
Reverse-mode AD fails on an opaque generator closure


BJMCox:
This fails in a fresh process with Julia 1.12.7 and Enzyme 0.13.202 on macOS arm64:

```julia
using Enzyme

function f(x)
    g = Base.Experimental.@opaque i -> x * i
    return sum(Base.Generator(g, 1:2))
end

Enzyme.autodiff(Reverse, f, Active, Active(3.0))
```

Removing `Base.Experimental.@opaque` gives the expected derivative, `3.0`, also verified in a fresh process.

<details>
<summary>Full error output (script paths shortened)</summary>

```text
ERROR: LoadError: EnzymeRuntimeException: Enzyme execution failed within

f(::Float64)
     @ Main mre.jl:3

Hint: catch this exception as `err` and call `code_typed(err)` to inspect the surrounding code.

Enzyme: jl_call calling convention not implemented in aug_forward for   %3 = call nonnull "enzyme_type"="{[-1]:Pointer}" ptr addrspace(10) (ptr, ptr addrspace(10), ...) @julia.call(ptr nonnull @jl_new_opaque_closure_jlcall, ptr addrspace(10) null, ptr addrspace(10) @"ejl_inserted$_Core_Tuple_156$false$4579069968", ptr addrspace(10) @"ejl_inserted$jl_global_157$false$4944444256", ptr addrspace(10) @"ejl_inserted$_Core_Any_158$false$4873570800", ptr addrspace(10) %1, ptr addrspace(10) addrspacecast (ptr inttoptr (i64 4606807568 to ptr) to ptr addrspace(10)), ptr addrspace(10) nonnull %"box::Float64") #8, !dbg !15

Stacktrace:
  [1] f
    @ mre.jl:4 [inlined]
  [2] augmented_julia_f_154wrap
    @ mre.jl:0
  [3] macro expansion
    @ ~/.julia/packages/Enzyme/3wKgd/src/compiler.jl:7149 [inlined]
  [4] enzyme_call
    @ ~/.julia/packages/Enzyme/3wKgd/src/compiler.jl:6615 [inlined]
  [5] AugmentedForwardThunk
    @ ~/.julia/packages/Enzyme/3wKgd/src/compiler.jl:6561 [inlined]
  [6] autodiff
    @ ~/.julia/packages/Enzyme/3wKgd/src/Enzyme.jl:430 [inlined]
[...]

PR #3588, opened 2026-09-13
Differentiate calls through opaque closure specptrs

ChrisRackauckas-Claude:
## Summary

Fixes the silent wrong gradient reported in https://github.com/EnzymeAD/Enzyme.jl/issues/3551#issuecomment-5603846841:

```julia
const f = Base.Experimental.@opaque (x::Float64) -> 3x
g(x) = f(x)
autodiff(Reverse, g, Active, Active(2.0))   # was ((0.0,),), now ((3.0,),)
```

### Root cause

A call to a `const` `OpaqueClosure` lowers to an **indirect call through the closure's `specptr` field** (`call (inttoptr (load (gep const-addr)))`), not to a `julia.call`. `check_ir!` only resolves `jlplt_*_got` loads, `jl_lazy_load_and_lookup`, `ConstantExpr inttoptr` C symbols, and `julia.call` builtins, so the callee stayed unresolved. Enzyme core's `up-constfn` activity rule then treats any call through a provably-constant callee as inactive — argument activity is never consulted — producing a silently zero gradient.

### Fix

In `check_ir!`, for calls whose callee is not a `Function`/`InlineAsm`, `resolve_indirect_callee!`:

1. Chases the callee operand through `inttoptr`/`bitcast`/constant offsets (via `get_base_and_offset`), dereferencing a `load` only when its address is itself constant (a `julia.constgv` global or a literal pointer), then asks `jl_lookup_code_address` for the `MethodInstance`.
2. If the pointer can't be resolved, checks whether the first call argument is a constant `Core.OpaqueClosure` (opaque closure calls pass the closure object as arg0) and recovers the `mi` from `specptr` or `source.specializations`.
3. Emits the `mi` into the module with `emit_mi_into_module!` — a mirror of `GPUCompiler.compile_method_instance` on the `jl_emit_native` path, needed because `compile_method_instance` rejects methods only valid from `typemax(UInt)`, which is exactly how Julia marks opaque closure methods. Inference still runs under the Enzyme interpreter so calls inside the emitted body keep Enzyme's markers; the emitted `mi` is added to the compilation `edges`.
4. Rewrites the call only when the emitted `LLVM.Function`'s type exactly equals `called_type(inst)`; otherwise the original indirect call is kept.

`check_ir!` iterates to a fixpoint since newly linked functions need checking too, and emitted callees are cached per-module. Anything unresolvable falls back to the previous behavior unchanged.

Note this does **not** address the other failure mode in #3551 — an opaque closure *constructed inside* the differentiated function (`jl_new_opaque_closure_jlcall`, an unimplemented `julia.call` builtin) — which is a separate path.

## Test plan

New testset `opaque closure calls` in `test/runtime_calls.jl` covers const non-capturing and capturing opaque closures.

Before (on `main`, `19de408`):
```
primal: 6.0
grad on main: ((0.0,),)
```
After (this branch): derivative `3.0` (non-capturing) / `4.0` (capturing); the same new tests fail with `0.0` on the unfixed code.

Test groups run locally (Julia 1.12.4, Enzyme_jll 0.0.293): `runtime_calls` (8 pass), `basic` (539 pass / 2 pre-existing broken), `advanced`, `passes`, `errors`, `mixed`, `mixedrrule`, `absint`, `mixedapplyiter`, `callconv` (143 pass) — all green. Julia versions other than 1.12 and GPU configs were not exercised locally (the new code is gated `VERSION >= v"1.12.0-DEV.1823"`, falling back to previous behavior elsewhere).

🤖 Generated with [Devin](https://devin.ai) (harness: Devin CLI 3000.10.21, model: SWE-2 Max, session: `/home/crackauc/.local/share/devin/cli/summaries/history_e89ce673d4064823.md` — local transcript, no shareable URL)

Co-authored with Chris Rackauckas.


ChrisRackauckas:
I'm not the right one to follow up here, but the point was more to show that is the case.

PR #3590, opened 2026-09-14
Take the GPUArrays-generic rules over from CUDA.jl

vchuravy:
CUDA.jl's `CUDACore/ext/EnzymeCoreExt.jl` carries rules that have nothing to do with CUDA. `GPUArrays.mapreducedim!` and `GPUArrays._mapreduce` are functions GPUArrays owns, and the `make_zero`/`make_zero!` and `Base.fill!` definitions sitting next to them only need `fill!`, broadcast and a host round-trip. All of them were keyed on `DenseCuArray`/`AnyCuArray`, so every other backend went without: AMDGPU's `AMDGPUEnzymeCoreExt` has none of them, and `sum` over a `JLArray` aborted — something `test/ext/jlarrays.jl` already documented in a comment.

This PR takes them over, keyed on `AbstractGPUArray`/`AnyGPUArray`:

- **new `EnzymeGPUArraysExt`** (weakdep GPUArrays) for the two reduction entry points, which are GPUArrays functions and so need the full package rather than GPUArraysCore;
- **`make_zero`/`make_zero!` and the `Base.fill!` rules into `EnzymeGPUArraysCoreExt`**, which `AbstractGPUArray` alone is enough for.

`DenseCuArray` is `CuArray{T,N}` with no wrappers, so `AbstractGPUArray` is its exact analogue and no coverage is lost.

### The CUDA.jl side

JuliaGPU/CUDA.jl#3272 deletes the originals (−313 lines). The two halves can land independently: `AnyCuArray <: AnyGPUArray` holds including the wrapped-array arm, so while both exist CUDA's copies stay strictly more specific and keep winning. No ambiguity, no behavior change during the transition.

### Fixed while porting

A faithful copy would have shipped these to every backend:

- the forward `mapreducedim!` rule keyed its first branch on `A` and its batched branch on `R`, so a `Const` input with a `Duplicated` output left the shadow stale — and the inner `A isa Const` test was unreachable;
- both its shadow and augmented-primal calls dropped the `init` kwarg;
- the `_mapreduce` reverse rule typed `dres::Active{RT}` and threw a `MethodError` at width > 1; it now takes the cotangent untyped and unwraps per batch;
- the `fill!` rules branched on `!(T <: AbstractFloat)` to pick the inactive path, which sends complex the wrong way. Unreachable while the bound was `MemsetCompatTypes`, live now that it isn't.

### `fill!` element types

The bound widens from `MemsetCompatTypes` (`UInt8`, `Int8`, `UInt16`, `Int16`, `Float16`, `UInt32`, `Int32`, `Float32`) to `Union{AbstractFloat, Complex{<:AbstractFloat}, Integer}`. That union describes CUDA's memset implementation, not what the derivative needs — the rule only needs `zero(T)` and `sum` — so `Float64` and complex arrays no longer fall through to the fill kernel. Element types outside the new union still have no rule, so nothing that works today can start throwing on a missing `zero`.

Shadow writes use the file's existing `_isconst`, not `A isa Const`: under runtime activity a shadow can alias its primal, and zeroing it would destroy the primal.

### Tests

`test/ext/jlarrays.jl` gains 26 assertions — reductions (reverse, batched reverse, forward, batched forward, complex), `make_zero` (bulk path, host path for a mixed-field struct, aliasing), and `fill!` (reverse, scaled, batched, forward with `Duplicated` and with a `Const` fill value, `Float32` cotangent type, complex). `JLArray` is a CPU-backed `AbstractGPUArray`, so this runs in CI with no device.

Each rule is load-bearing on `JLArray`: without the reduction rules `sum` core-dumps, and without the `fill!` rule you get `MethodError: no method matching mkcontext(::Kernel{JLBackend, ...})` out of KernelAbstractions' Enzyme extension. 54/54 pass with them; `test/make_zero.jl` (157k assertions) is unaffected.

### Known gap, not introduced here

Reverse-mode `sum(x; dims=)` throws `custom_rule_method_error` because the `_mapreduce` augmented rule only accepts an `Active` return. Same on CUDA today; ported as is.

https://claude.ai/code/session_018ofARZr5bm2nYMhzHTrHxK


ext/EnzymeGPUArraysCoreExt.jl
```diff
@@ -46,6 +46,72 @@ end
     end
 end
 
+#=
+`make_zero`/`make_zero!` are defined element-wise, which would require scalar
+indexing on a GPU array, so zero the array in bulk instead. That is only
+equivalent to the element-wise definition when every part of the element type is
+a float, which covers `Float32`, `ComplexF64`, `SVector{3,Float64}` etc. For
```
kshyatt, line 53: wait, wouldn't this also be true if the elements were `Integer` types too?

michel2323:
Tested this on AMDGPU (MI300X, ROCm 7.0.2, Julia 1.13.0, AMDGPU.jl 2.8.0 with GPUCompiler 2.7), merged on top of #3512.

Without it, `Enzyme.make_zero` on a `ROCArray` falls into the generic path, recurses through `GPUArrays.DataRef -> AMDGPU.Managed -> ReentrantLock` and throws:

```
ConcurrencyViolationError("setfield!: atomic field cannot be written non-atomically")
  make_zero(::Type{ReentrantLock}, seen::IdDict{Any, Any}, prev::ReentrantLock, ::Val{false})
  make_zero(::Type{AMDGPU.Managed{AMDGPU.Runtime.Mem.HIPBuffer}}, ...)
  make_zero(::Type{GPUArrays.DataRef{AMDGPU.Managed{AMDGPU.Runtime.Mem.HIPBuffer}}}, ...)
  make_zero(::Type{ROCArray{Float64, 1, AMDGPU.Runtime.Mem.HIPBuffer}}, ...)
```

With it, `EnzymeGPUArraysCoreExt` picks up `ROCArray` through `AbstractGPUArray` and `make_zero` returns a zeroed `ROCArray`. Together with JuliaGPU/AMDGPU.jl#1095 (Enzyme rules for the `ROCArray` constructors), the checkpointed whole-model reverse-mode test of SciDAC-ImPACTS/UnstructuredOceans.jl#11 runs on `ROCBackend` and matches both finite differences and the CPU gradient to the last digit.

ext/EnzymeGPUArraysCoreExt.jl
michel2323: I think `Integer` is treated as not active and left untouched by `make_zero`.

Issue #3602, opened 2026-09-16
Enzyme still messing up shadows for immutable types


kshyatt:
Follow up to https://github.com/EnzymeAD/Enzyme.jl/issues/2304

I tried for a long long time to get 🤖 to remove the dependencies and was not able to, sorry.

Here's the problem:

Enzyme loses the inline (`bits_ref`) half of a shadow for mixed pointer/inline immutables, so `TensorKit.TensorMap.space` comes back zeroed (`ℂ^0`) and a downstream contraction throws `DimensionMismatch`.

Setup:
```bash
git clone -b ksh/enzyme https://github.com/QuantumKitHub/PEPSKit.jl && cd PEPSKit.jl
julia +1.12 --project=. -e 'using Pkg; Pkg.instantiate(); Pkg.add("Enzyme")'
julia +1.12 --project=. enzyme_shadow_bits_repro.jl
```

The cause, measured with a probe inside the custom rule: shadow tensors reaching
the rule have their GC-pointer field intact -- `data` is the *same object* as the
correctly spaced shadow produced upstream, with the correct length -- while the
inline field is zeroed, so `space` reads `ℂ^0 ← (ℂ^0 ⊗ ℂ^0 ⊗ ℂ^0 ⊗ ℂ^0)` with no
duals where the primal has `(ℂ^2)'`.  Element [1] of each affected array is
spared; [2..n] are hit.

Rebuilding each affected shadow around its surviving `data` buffer (using the
primal's space) makes this go away, which is what PEPSKit does on top of this
branch; that the buffer is shared is why the workaround is sound.

Here's the reproducer itself:

```julia
using Random, LinearAlgebra, PEPSKit, TensorKit, Enzyme, VectorInterface
Enzyme.LLVM.clopts("--non-global-value-max-name-size=1048576")

Random.seed!(81812781144)
const psi0 = InfinitePEPS(randn, ComplexF64, ℂ^2, ℂ^2; unitcell = (1, 1))
const ctm  = SimultaneousCTMRG(; maxiter = 150, tol = 1.0e-8, verbosity = 0)
const env0 = CTMRGEnv(InfiniteSquareNetwork(psi0), ℂ^12)

function E(psi)
    n = InfiniteSquareNetwork(psi)
    env, = leading_boundary(env0, n, ctm)
    return real(network_value(n, env))
end

dpsi = zerovector(psi0)
try
    _, val = Enzyme.autodiff(
[...]

Issue #3477, opened 2026-08-21
Enzyme behaving badly with shadows of immutables


kshyatt:
# Enzyme reverse mode: per-iteration allocations collapse to one when a custom rule consumes an immutable struct with mixed pointer/inline layout

Developed with 🤖 

## Summary

When a loop allocates an **immutable struct that mixes a GC-tracked pointer field with an
inline (non-pointer) field**, and a function with a **custom `EnzymeRules` reverse rule**
consumes that struct, Enzyme collapses all iterations' allocations onto a **single
object**. Both the cached **primal** and the **shadow** are shared. The result is a
silently wrong gradient — no error, no warning.

In the barest form, a 4-iteration loop returns the gradient of **only its final
iteration**.

## Reproducer

`enzyme_loop_shadow_repro.jl` (Enzyme-only, no other dependencies).

```julia
struct Mixed; data::Vector{Float64}; k::Int; end   # pointer + inline  <- trigger

red(x) = sum(x.data)                               # carries a custom rule (below)

const CO = [1.0, 2.0, 3.0, 4.0]
@noinline mk(a, i) = Mixed(CO[i] .* copy(a.data), a.k)

f(a) = (s = 0.0; for i in 1:4; s += red(mk(a, i)); end; s)

a = Mixed([1.0, 2.0, 3.0], 3); da = Enzyme.make_zero(a)
Enzyme.autodiff(Reverse, f, Active, Duplicated(a, da))

da.data[1]   # got 4.0, expected sum(CO) == 10.0
```

The rule is an ordinary accumulating reverse rule:

```julia
function EnzymeRules.reverse(config::EnzymeRules.RevConfigWidth{1},
        func::Const{typeof(red)}, dret::Active, cache, A::Annotation)
    isa(A, Const) || (A.dval.data .+= dret.val)
    return (nothing,)
end
```

`df/d a.data[1]` is exactly `sum(CO)`: iteration `i` contributes `CO[i] * 1`.

## Direct evidence: primal and shadow are both shared

Printing object identities from inside the reverse rule, for the loop above:

```
[rev] primal.data=0xee8652f407420a50 val=4.0 | shadow.data=0x9af5aaa06020b93a dval[1](before)=0.0
[rev] primal.data=0xee8652f407420a50 val=4.0 | shadow.data=0x9af5aaa06020b93a dval[1](before)=0.0
[...]

PR #3606, opened 2026-09-18
Fix empty enzyme_truetype on memcpy with a negative folded offset

vchuravy:
Partial fix for #3602 (the zeroed `TensorMap.space` shadows).

## Problem

When a `memcpy` is annotated with `enzyme_truetype`, the destination offset comes from `get_base_and_offset`, which folds the constant part of the GEP chain. For loop bodies LLVM commonly rewrites the element address as `base + stride*i - k`. In the PEPSKit reproducer from #3602 the destination is `base + 72*i - 64` (a `TensorMap` with the `space` field at offset 8), so the folded constant offset is `-64`.

The offset was reduced with `%`, which keeps it negative. `to_fullmd` then treats every field as lying past the end of the copied range and returns an empty `MDNode`. Enzyme interprets an empty `enzyme_truetype` as "nothing to copy", so the shadow destination never receives the inline (integer) bytes and keeps its zero initialisation. The peeled first iteration has a proper non-negative offset and is copied correctly, which produces the "element 1 correct, elements 2..n zeroed" pattern from the issue: the `space` of shadow `TensorMap`s inside arrays came back as `ℂ^0` and a later contraction threw `DimensionMismatch`.

Pre-Enzyme IR from the reproducer, peeled iteration vs. loop body (note the `!enzyme_truetype !0` on the second):

```llvm
; peeled first iteration
%29 = getelementptr i8, ptr addrspace(13) %28, i64 8
call void @llvm.memcpy.p13.p0.i64(ptr addrspace(13) %29, ptr %.sroa.11765, i64 64, i1 false), ..., !enzyme_truetype !2127

; loop body
%40 = getelementptr i8, ptr addrspace(13) %28, i64 %39   ; %39 = 72*i
%41 = getelementptr i8, ptr addrspace(13) %40, i64 -64
call void @llvm.memcpy.p13.p0.i64(ptr addrspace(13) %41, ptr %.sroa.11765, i64 64, i1 false), ..., !enzyme_truetype !0
```

and the corresponding augmented forward: the peeled iteration gets a copy into the shadow destination, the loop body gets only the primal copy.

## Fix

Reduce the offset with `mod` so it lands in `[0, sz_et)`, and skip the annotation if the offset is still negative rather than emitting an empty one.

## Verification

- With the fix, the `gauge_fixed_iteration` thunk from the PEPSKit reproducer completes and probes inside the custom rules no longer see any `ℂ^0` shadow spaces. The full reproducer from #3602 runs to completion with PEPSKit's `_repair_shadow_spaces!` workaround disabled, and the gradient matches the workaround-enabled run.
- Locally ran `test/rules/*.jl`, `mixedrrule.jl`, `rooted_args.jl`, `mixed.jl`, `usermixed.jl`, `mixedapplyiter.jl`, `applyiter.jl`, `array.jl`, `typeunstable.jl`, `sugar.jl`, `make_zero.jl`, `recursive_accumulate.jl` on Julia 1.12: all pass.

I was not able to get plain Julia code outside PEPSKit to produce a `memcpy` whose destination both goes through an annotated `julia.gc_loaded` and has the negative folded offset, so this PR carries no standalone regression test. Suggestions for a shape that triggers it are welcome.

Note this does not cover the second problem behind #3602 / #3477 (custom-rule arguments whose inline-roots stack slot is overwritten in a loop); that is a separate change in `customrules.jl`.


src/compiler.jl
```diff
@@ -6129,7 +6133,7 @@ end
                     )
 
 			 size = Compiler.datatype_layoutsize(jTy)
-                        if offset < size && isa(sz, LLVM.ConstantInt) && size - offset >= convert(Int, sz)
+                        if 0 <= offset < size && isa(sz, LLVM.ConstantInt) && size - offset >= convert(Int, sz)
```
vchuravy, line 6136: Should we turn this into `@assert offset > 0`

src/compiler.jl
```diff
@@ -6096,7 +6096,7 @@ end
...
```
vchuravy:
```suggestion
                        @assert offset >= 0 
                        if offset < size && isa(sz, LLVM.ConstantInt) && size - offset >= convert(Int, sz)
```

PR #3607, opened 2026-09-18
Rebuild overwritten roots of custom-rule arguments from the tape

### LINKED ISSUE #3477: Enzyme behaving badly with shadows of immutables

vchuravy:
Fixes #3477 and the custom-rule half of #3602 (the "[1, 0, …, n-1]" gradient pattern). The other half of #3602, the zeroed inline shadow bytes, is #3606.

## Problem

On Julia 1.12 a by-reference immutable with mixed pointer/inline layout (e.g. `struct Mixed; data::Vector{Float64}; k::Int; end`) is passed to a custom rule as an inline-data stack slot plus a separate inline-roots array. When such a call sits in a loop, both slots are reused by every iteration and Enzyme correctly flags them as overwritten (`uncacheable = [1, 1]`).

`enzyme_custom_setup_args` already cached the recombined primal value (inline data plus roots) on the byval tape in the forward pass. But in the reverse pass it still built the roots array handed to the rule from the *live* roots memory, which by then holds whatever the last iteration stored there. The shadow side had the same problem: the shadow roots were read from the live shadow-roots slot, and the `is_constant_value(op)` path taken when the inline half is inactive (a `Vector` plus an `Int`) assembles the shadow from the primal value plus those stale shadow roots.

So every reverse rule invocation in the loop saw the last iteration's pointers for both primal and shadow, collapsing all iterations onto one object. Reverse IR from the reproducer in #3477 before the fix, showing the shadow root loaded from the live slot `%"'ipa"` for every call:

```llvm
%95 = getelementptr inbounds [1 x ptr addrspace(10)], ptr %"'ipa", i64 0, i32 0
%rules_shadow_root_load_1_26 = load ptr addrspace(10), ptr %95
store ptr addrspace(10) %rules_shadow_root_load_1_26, ptr %94   ; into roots.arg.Duplicated{Mixed}
```

## Fix

- **primal:** when the roots were overwritten, rebuild the roots array in the reverse pass from the tape-cached recombined value instead of the live slot, and mark the recombining loads must-cache so they are not rematerialised from the stale slot.
- **shadow:** when the roots were overwritten, cache the shadow roots on the byval tape in the forward pass and rebuild the shadow roots array from the tape in the reverse pass (batch widths handled).
- **MixedDuplicated:** in the reverse pass the shadow's by-ref memory only holds meaningful float fields (the live accumulators); take the remaining inline data from the cached primal value, which the shadow must mirror.

## Verification

- The reproducer from #3477 (`red(mk(a, i))` in a loop of 4) now gives `da.data[1] == 10.0` in both combined and split mode, and the reverse rule sees each iteration's own primal and shadow; before it gave `4.0` with the last iteration's pointers repeated four times.
- New `test/rules/overwritten_roots.jl` covers `Duplicated`, `BatchDuplicated` (width 2) and `MixedDuplicated` arguments, in combined and split mode, checking both the gradient and the per-iteration values seen by the rule.
- Locally ran `test/rules/*.jl`, `mixedrrule.jl`, `rooted_args.jl`, `mixed.jl`, `usermixed.jl`, `mixedapplyiter.jl`, `applyiter.jl`, `array.jl`, `typeunstable.jl`, `sugar.jl`, `make_zero.jl`, `recursive_accumulate.jl` on Julia 1.12: all pass.
- Together with #3606, the PEPSKit reproducer from #3602 runs to completion with PEPSKit's shadow-repair workaround disabled.


src/rules/customrules.jl
```diff
@@ -541,7 +541,7 @@ Custom rule for method argument $arg_idx of type $(arg.typ) has mismatch between
                         if roots_op != nothing
                             if uncacheable[arg.codegen.i + 1] != 0
                                 # Roots are overwritten, recombine with root
-                                val = recombine_value!(B, val, roots_val)
+                                val = recombine_value!(B, val, roots_val; must_cache = true)
```
wsmoses: undo this change

wsmoses: actually i take it back leave the must cache

src/rules/customrules.jl
```diff
@@ -559,7 +559,17 @@ Custom rule for method argument $arg_idx of type $(arg.typ) has mismatch between
...
+                                # The roots were overwritten, so the forward pass recombined
+                                # them into the cached value. Rebuild the roots array from
+                                # that cached value rather than reading the live roots
+                                # memory, which by now holds whatever was last stored there
+                                # (e.g. the final iteration of a loop).
+                                roots_val = create_rooted_array(alloctx, roots, "roots_op_from_tape_")
+                                extract_roots_from_value!(B, val, roots_val)
+                            else
+                                roots_val = lookup_value(gutils, roots_val, B)
+                            end
```
wsmoses: okay this change and above I agree with, pondering second

src/rules/customrules.jl
```diff
@@ -756,7 +766,47 @@ Custom rule for method argument $arg_idx of type $(arg.typ) has mismatch between
...
+                                    b_ival = UndefValue(value_type(roots_ival0))
+                                    for idx in 1:width
+                                        al = create_rooted_array(alloctx, roots, "shadow_roots_from_tape_")
+                                        store!(B, extract_value!(B, sroot_cache, idx - 1), al)
+                                        b_ival = insert_value!(B, b_ival, al, idx - 1)
+                                    end
+                                    roots_ival0 = b_ival
+                                end
+                            end
+                        elseif reverse
```
wsmoses: this above confirmed

src/rules/customrules.jl
```diff
@@ -821,7 +871,7 @@ Custom rule for method argument $arg_idx of type $(arg.typ) has mismatch between
                                     metadata(ld0)["enzyme_mustcache"] = MDNode(LLVM.Metadata[])
                                     if roots_op != nothing
                                         if uncacheable[arg.codegen.i + 1] != 0
-                                            ld0 = recombine_value!(B, ld0, local_shadow_root)
+                                            ld0 = recombine_value!(B, ld0, local_shadow_root; must_cache = true)
```
wsmoses: fine

src/rules/customrules.jl
```diff
@@ -919,11 +969,34 @@ Custom rule for method argument $arg_idx of type $(arg.typ) has mismatch between
...
+                            @assert tape isa LLVM.Value
+                            cached = extract_value!(B, tape, length(byval_tapes), "rules_mixed_shadow_extract_")
+                            @assert value_type(cached) == llrty
+                            push!(byval_tapes, cached)
+                            tmp = alloca!(alloctx, llrty, "rules_mixed_shadow_tmp")
+                            store!(B, cached, tmp)
+                            copy_floats_into!(B, llrty, tmp, ev)
+                            load!(B, llrty, tmp, "rules_mixed_shadow_merged")
+                        else
+                            ld0 = load!(B, llrty, ev, "rules_mixed_shadow_load")
```
wsmoses: this logic is wrong. Can you fork this part into its own pr [and we can merge the rest and work on this subsequently]

vchuravy: This entire changeset? Or only the else part?

wsmoses: This section, the sections above are fine

wsmoses: Not just the else but this whole logical diff that is separated from the other two parts above

vchuravy: Done!

wsmoses (approved):

